PR #1, opened 2019-09-23
Fix the link to the issue page

aNiemack:
It was still pointing to the DART-2.0 testing repository, it should now go to the DART issues page.


nancycollins:
change looks good.  will try first 'confirm merge' and see what happens.

PR #5, opened 2020-01-08
Clean master

timhoar:


nancycollins:
i've reviewed these - they've been through code reviews to be put on DART_development and they're fine to go on the public repo.

PR #8, opened 2020-03-13
More information on adding support for new observations

timhoar:
Michael Ying issued a PR #7 that required a bit more than I wanted to foist on him.
Since the obs_def_mod.html is not operational from the web side of things, that information
can be spelled out a bit more completely in the online docs.


docs/pages/Getting_Started.md
```diff
@@ -2243,13 +2263,14 @@ and a capability to do a standard input driven interactive definition of the
 observation definition.
 
 DART makes a further distinction between specific *observation types* and
-generic *quantities*. The role of the various obs_def input files
+generic *quantities*. Essentially, an *observation type* can be considered to be instrument-specific and the *quantity* is something you learned about in physics.
+As an example, 'RADIOSONDE_TEMPERATURE' is an *observation type* of the fundamental *quantity* 'TEMPERATURE'. The role of the various obs_def input files
 is to define the mapping between the types and quantities, and optionally to
-provide type-specific processing routines.
+provide type-specific processing routines. Specifying many observation types allows DART to be able to evaluate some observations and assimilate others - even if both instrument measure the same *quantity*.
```
jlsucar: "both instrument" should be changed here. Also the beginning of the sentence says "many" but the end says "both." How about 

Specifying many observation types allows DART to be able to evaluate some observations and assimilate others - even if many instruments measure the same *quantity*.

docs/pages/Getting_Started.md
```diff
@@ -2214,25 +2216,43 @@ the tests available.
...
-2. [Simple observation definitions](#simpleObs)
-3. [Obs definitions for observations needing special handling](#complexObs)
-4. [Example observation definition](#exampleObsDef)
-5. [Observation sequence file](#obsSeqFile)
+1. [Adding new observation *quantities*](#newQUANTITIES)
+1. [Adding new observation *types*](#newTYPES)
+1. ['Simple' observation support](#simpleObs)
+1. [More complex observations](#complexObs)
+1. [Example observation definition](#exampleObsDef)
+1. [Observation sequence file](#obsSeqFile)
```
jlsucar: Are these intended to all be "1"?  It looks okay in markup, but it's a bit jarring to see.

docs/pages/Getting_Started.md
```diff
@@ -2258,8 +2279,26 @@ obs_def modules (for example, `obs_def_1d_state_mod.f90` or
 files are included in the preprocessor namelist, a minimal `obs_def_mod.f90` is
 created which can only support identity forward observation operators.
 
-To add a new observation type which does not fit into any of the
-already-defined obs_def files, a new file should be created in the
+<span id="newQUANTITIES" class="anchor"></span> [](#newQUANTITIES)
+#### To add a new *quantity*,
```
jlsucar: The style of #### Statement,

lowercase continuation of the statement

was confusing as I looked through this. Can you make this consistent with the rest of the document?

docs/pages/Getting_Started.md
```diff
@@ -2258,8 +2279,26 @@ obs_def modules (for example, `obs_def_1d_state_mod.f90` or
...
+! DART PREPROCESS OBS_QTY_INFO INSERTED HERE
+~~~
+
+If you add an observation quantity with an integer value greater than **max_defined_quantities**,
+make sure you increase *max_defined_quantities*.  
+
+<span id="newTYPES" class="anchor"></span> [](#newTYPES)
+#### To add a new observation *type* which does not fit into any of the already-defined obs_def files
+
+a new file should be created in the
```
jlsucar: Same comment as above - the style of subheaders should stay the same as the rest of the document.

docs/pages/Getting_Started.md
timhoar: Thanks ... I opted for.  '... - even if the instruments measure the same _quantity_.'

docs/pages/Getting_Started.md
timhoar: I KNOW! Cool, n'est pas?   It beats renumbering when you insert a new item, so yes, it is deliberate. Too wierd? By having them all be the same, you can rearrange them at will without worrying about it.

jlsucar: Yeah it's fine. I guess it is one of those Markdown things one just has to learn.

docs/pages/Getting_Started.md
timhoar: Done.

docs/pages/Getting_Started.md
timhoar: Made me think a little ... I added a paragraph to explain what should happen if the new observation type should be added to an existing obs_def module. to wit:

The observation support modules are in the `DARTHOME/observations/forward_operators` directory.
If the observation type logically belongs in an existing observation module, it should be 
fairly obvious how to add support for a new observation type. Feel free to use ideas from other modules - look around!

nancycollins: do we say anywhere here that for observation *types* that this is where they are defined.  for observation *quantities*, they have to already exist in the obs_kind_mod.

timhoar: Yes ... the section immediately before the observation types is all about quantities, 

check out:

https://github.com/timhoar/DART/blob/getting_started_page/docs/pages/Getting_Started.md#newQUANTITIES

timhoar: And in the paragraph of adding observation types (which appears AFTER the observation quantity section):   _If the new observation type also requires a new observation quantity, there is more work to be done._

PR #9, opened 2020-03-24
Merging the fun_with_pop scripts into Manhattan

johnsonb-ucar:
The ${DARTROOT}/models/POP/shell_scripts directory has been partitioned into two subdirectories, cesm1_x and cesm2_1. The latter subdirectory contains scripts to create CESM2 single member and multi-instance cases which can run perfect_model_obs and filter in the G configuration with 1° ocean/sea ice grids with active POP2 and CICE5 components.


johnsonb-ucar:
Unless instructed otherwise, I'm planning to close this pull request and reopen another one on the development master branch after I've had a chance to test the scripts with the f09 atmospheric grid.

models/POP/shell_scripts/cesm1_x/CESM1_1_1_setup_hybrid
```diff
@@ -747,9 +747,9 @@ cat << EndOfText >! stage_cesm_files
 set restart_time = $init_time
 
 # get the settings for this case from the CESM environment
-cd ${caseroot}
+cd \${caseroot}
```
timhoar, line 750: I am not sure this one needs 'escaping' ... it will expand into the full value when the script is being created, which is what I think should happen.   The variables that it gets from the ccsm_getenv call must be 'escaped', agreed.

models/POP/shell_scripts/cesm1_x/CESM1_2_1_setup_hybrid
```diff
@@ -754,9 +754,9 @@ cat << EndOfText >! stage_cesm_files
 set restart_time = $init_time
 
 # get the settings for this case from the CESM environment
-cd ${caseroot}
+cd \${caseroot}
```
timhoar, line 757: same deal ... I don't think this one should be escaped.

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_hybrid
```diff
@@ -0,0 +1,766 @@
...
+./xmlchange RESUBMIT=$resubmit
+
+./xmlchange PIO_TYPENAME=pnetcdf
+
+# These are archiving options that may be used.
+# You can turn the short/long term archivers on or off,
+# but these settings should be made in either event.
+
+./xmlchange DOUT_S_ROOT=${ARCHDIR}
+./xmlchange DOUT_S_SAVE_INTERIM_RESTART_FILES=TRUE
```
timhoar, line 320: I am not sure we generate any interim restart files. I think the only time we would generate them is in a single free run. Did you need this to be TRUE?

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_hybrid
```diff
@@ -0,0 +1,766 @@
...
+#     CONTINUE_RUN=TRUE  => you are starting from a previous step but not
+#                           the very first one.
+#  2) be sure 'restart_time' is set to the day and time that you want to
+#     restart from if not the initial time.
+
+set restart_time = $init_time
+
+# get the settings for this case from the CESM environment
+
+cd \${CASEROOT}
```
timhoar, line 482: No need to escape this ... unless I'm mistaken.

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_hybrid
```diff
@@ -0,0 +1,766 @@
...
+   # After running this script resubmit the job to rerun.
+   #----------------------------------------------------------------------
+
+   echo "Staging restart files for run date/time: " \${restart_time}
+
+   #  The short term archiver is on, so the files we want should be in one
+   #  of the short term archive 'rest' restart directories.  This assumes
+   #  the long term archiver has NOT copied these files to the HPSS yet.
+
+   if (  \${DOUT_S} == TRUE ) then
```
timhoar, line 514: DOUT_S should be set from a .xmlquery call ... similar to RUNDIR and CONTINUE_RUN.  You can turn short-term archiving on or off during the course of a run, and you might need to run stage_cesm_files more than just at the beginning.

models/POP/shell_scripts/cesm2_1/CESM_DART_config
```diff
@@ -0,0 +1,317 @@
...
+# ==============================================================================
+
+${COPY} ${DARTROOT}/models/POP/shell_scripts/${cesmtagmajor}/assimilate.csh          .
+${COPY} ${DARTROOT}/models/POP/shell_scripts/${cesmtagmajor}/perfect_model.csh       .
+
+# ==============================================================================
+# Stage the DART executables in the CESM execution root directory: EXEROOT
+# If you recompile the DART code (maybe to support more observation types)
+# we're making a script to make it easy to install new DART executables.
+# ==============================================================================
```
timhoar, line 155: So there is a question of whether or not we want to automatically create inflation files when setting up an experiment or not. It's always a nuisance trying to automate the creation of the inflation files if they do not exist and the user decides to use inflation  ... the first cycle has to be from namelist and then all remaining cycles are different ...

models/POP/shell_scripts/cesm2_1/DART_params.csh
```diff
@@ -0,0 +1,216 @@
...
+# job settings and machine-specific commands:
+#
+# mach            Computer name
+# ACCOUNT         Project that hours will be charged to
+# queue           can be changed during a series by changing the ${CASE}.run
+# timewall        can be changed during a series by changing the ${CASE}.run
+# ==============================================================================
+
+setenv mach         cheyenne
+setenv ACCOUNT      P93300612
```
timhoar, line 179: Frequently we obfuscate the ACCOUNT so as to force them to use their own project number.

models/POP/shell_scripts/cesm2_1/HPSS_get_ensemble
```diff
@@ -0,0 +1,67 @@
...
+while ($iyear <= $end_year)
+    
+    # Zero pad the year string so that there are 4 digits
+    if ($iyear < 10) then
+        setenv zero_filled_year 000${iyear}
+    else if ($iyear < 100) then
+        setenv zero_filled_year 00${iyear}
+    else if ($iyear < 1000) then
+        setenv zero_filled_year 0${iyear}
+    endif
```
timhoar, line 47: replace all with one line
setenv zero_filled_year = `printf %04d $iyear`

models/POP/shell_scripts/cesm2_1/HPSS_get_ensemble
```diff
@@ -0,0 +1,67 @@
...
+#
+# ==============================================================================
+# 
+# ---------------------
+# Purpose
+# ---------------------
+#
+# This script gets 80 POP restart files from HPSS.
+#
+# ==============================================================================
```
timhoar, line 17: No it doesn't ... 270-209 + 1 = 62 files.
Since you have  a 'year_span' ... are you trying to support getting separated years to use as an initial condition?
If that is the case, you could mention that in the Purpose, and I'd suggest using 'year_stride' rather than 'year_span'

models/POP/shell_scripts/cesm2_1/assimilate.csh
```diff
@@ -0,0 +1,358 @@
...
+
+# IFF we want PRIOR inflation:
+
+if ( $PRIOR_INF > 0 ) then
+
+   if ($PRIOR_TF == false) then
+      # we are not using an existing inflation file.
+      echo "inf_flavor(1) = $PRIOR_INF, using namelist values."
+
+   else if ( -e pop_inflation_cookie ) then
```
timhoar, line 158: I admit I have not followed the creation or destruction of the pop_inflation_cookie.
Usually, the DART_config creates it and the assimilate.csh destroys it. I don't see it
getting destroyed (eaten) in this script.

models/POP/shell_scripts/cesm2_1/assimilate.csh
```diff
@@ -0,0 +1,358 @@
...
+
+      # If one exists, use it as input for this assimilation
+      if ( $nfiles > 0 ) then
+         set latest = `cat latestfile`
+         ${LINK} $latest input_priorinf_mean.nc
+      else
+         echo "ERROR: Requested PRIOR inflation but specified no incoming inflation file."
+         echo "ERROR: expected something like ${CASE}.pop.output_priorinf_mean.YYYY-MM-DD-SSSSS.nc"
+         exit -4
+      endif
```
timhoar, line 189: FYI: I created this logic and I hate it.  I should have employed a pointer file strategy. My bad, and I apologize. As the directory gets filled, the 'ls' command has taken longer than the assimilation in certain circumstances!

models/POP/shell_scripts/cesm2_1/assimilate.csh
```diff
@@ -0,0 +1,358 @@
...
+${REMOVE} restarts_in.txt restarts_out.txt
+
+foreach FILE ( rpointer.ocn_????.restart )
+   head -n 1 ${FILE} >> restarts_in.txt
+end
+
+# WARNING: this is the part where the files just get overwritten
+${COPY} restarts_in.txt restarts_out.txt
+
+# POP always needs a pop_in and a pop.r.nc to start.
```
timhoar, line 281: It's actually filter than needs a pop_in and a pop.r.nc to start.

models/POP/shell_scripts/cesm2_1/assimilate.csh
```diff
@@ -0,0 +1,358 @@
...
+#
+#=========================================================================
+
+echo "`date` -- BEGIN FILTER"
+${LAUNCHCMD} ${EXEROOT}/filter || exit -6
+echo "`date` -- END FILTER"
+
+if ( $?LSB_PJL_TASK_GEOMETRY ) then
+   setenv LSB_PJL_TASK_GEOMETRY "${ORIGINAL_LAYOUT}"
+endif
```
timhoar, line 323: I believe this is detritus. At one point filter was running better on a subset of tasks (LSB_PJL_TASK_GEOMETRY)
than the whole processor set (there just wasn't enought obs to warrant it). This block reset the tasks to the original task list ... but I don't see the original list being created anywhere. Am I missing that?

models/POP/shell_scripts/cesm2_1/assimilate.csh
```diff
@@ -0,0 +1,358 @@
...
+# The observation file names have a time that matches the stopping time of POP.
+#=========================================================================
+
+set YYYYMMDD = `printf %04d%02d%02d ${OCN_YEAR} ${OCN_MONTH} ${OCN_DAY}`
+set YYYYMM   = `printf %04d%02d     ${OCN_YEAR} ${OCN_MONTH}`
+set OBSFNAME = obs_seq.0Z.${YYYYMMDD}
+set OBS_FILE = ${BASEOBSDIR}/${YYYYMM}/${OBSFNAME}
+
+if (  -e   ${OBS_FILE} ) then
+   ${LINK} ${OBS_FILE} obs_seq.out
```
timhoar, line 75: I don't understand how this isn't failing on the second cycle.  The obs_seq.out already exists, and the LINK command does not (and should not) have the 'force' option specified.

models/POP/work/mkmf_fill_inflation_restart
```diff
@@ -0,0 +1,18 @@
...
+../../../build_templates/mkmf -p fill_inflation_restart -t ../../../build_templates/mkmf.template   \
+ -a "../../.." path_names_fill_inflation_restart
+
+exit $status
+
+# <next few lines under version control, do not edit>
+# $URL$
+# $Revision$
+# $Date$
+
```
timhoar, line 18: Please remove this block.

models/POP/work/mkmf_fill_inflation_restart
```diff
@@ -0,0 +1,18 @@
+#!/bin/csh
+#
+# DART software - Copyright UCAR. This open source software is provided
+# by UCAR, "as is", without charge, subject to all terms of use at
+# http://www.image.ucar.edu/DAReS/DART/DART_download
+#
+# DART $Id$
```
timhoar, line 7: Please remove.

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
```diff
@@ -0,0 +1,763 @@
...
+echo '--run-unsupported'
+echo '--pecount ${use_tasks_per_node}x${nthreads}'
+
+${cesmroot}/cime/scripts/create_newcase \
+--case ${CASEROOT} \
+--mach ${mach} \
+--res ${resolution} \
+--compset ${compset} \
+--run-unsupported \
+--pecount ${use_tasks_per_node}x${nthreads}
```
timhoar, line 185: For a single-instance case does compiling on 36 tasks speed things up? I don't use this option - maybe I should?

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
```diff
@@ -0,0 +1,763 @@
...
+./xmlchange RUNDIR=${RUNDIR}
+
+./xmlchange DATM_MODE=CPLHIST
+./xmlchange DATM_CPLHIST_CASE=${CASE}
+./xmlchange DATM_CPLHIST_YR_ALIGN=$stream_year_align
+./xmlchange DATM_CPLHIST_YR_START=$stream_year_first
+./xmlchange DATM_CPLHIST_YR_END=$stream_year_last
+
+# The streams files were generated with a NO_LEAP calendar in mind.
+# We need to test these with a GREGORIAN calendar.
```
timhoar, line 299: I bet we can pitch this comment now ...

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
```diff
@@ -0,0 +1,763 @@
...
+./case.setup
+
+if ( $status != 0 ) then
+   echo "ERROR: Case could not be set up."
+   exit -2
+endif
+
+# ==============================================================================
+# Modifying namelist templates for each instance.
+# ==============================================================================
```
timhoar, line 387: Since there is only 1 instance, do you want to mention that the reason there is a loop is to minimize the differences between this script and the hybrid setup script. So if you get familiar with this one, the assimilation setup will be pretty familiar.

models/POP/shell_scripts/cesm2_1/perfect_model.csh
```diff
@@ -0,0 +1,147 @@
...
+# :t is a csh modifier that returns the "tail" of the filename
+# essentially stripping away the filepath that prepends the file
+set FILE = $FILE:t
+# :r is a csh modifier that returns the "root" of the filename 
+# essentially stripping away the file suffix
+set FILE = $FILE:r
+# :e is a csh modifier that returns the "extension" of the filename
+# since we stripped away the .nc in the previous command, the current
+# extension is the datetime in the form YYYY-MM-DD-SSSSS
+set OCN_DATE_EXT = `echo $FILE:e`
```
timhoar, line 59: Kevin showed me this ... you can just stack these:
set OCN_DATE_EXT = $FILE:t:r:e
even if you don't want to stack them ... the syntax for OCN_DATE_EXT can be simplified:
set OCN_DATE_EXT = $FILE:e        no need for the extra stuff.

models/POP/shell_scripts/cesm2_1/perfect_model.csh
```diff
@@ -0,0 +1,147 @@
...
+   echo "ERROR ... DART required file ${CASEROOT}/input.nml not found ... ERROR"
+   echo "ERROR ... DART required file ${CASEROOT}/input.nml not found ... ERROR"
+   exit -3
+endif
+
+echo "`date` -- END COPY BLOCK"
+
+set OCN_RESTART_FILENAME = ${CASE}.pop.r.${OCN_DATE_EXT}.nc
+# BKJ I think setting this is unnecessary
+# set     OCN_NML_FILENAME = pop_in
```
timhoar, line 109: then lets get rid of them

models/POP/shell_scripts/cesm2_1/perfect_model.csh
```diff
@@ -0,0 +1,147 @@
...
+
+${LINK} ./$OCN_RESTART_FILENAME pop.r.nc
+# BKJ I think setting this is unnecessary
+# ${LINK} ../$OCN_NML_FILENAME     pop_in
```
timhoar, line 113: At one point, we supported LANL's POP as well as CESM's POP ... I don't remember if/why these were ever used by one or the other or both ...

models/POP/shell_scripts/cesm2_1/perfect_model.csh
```diff
@@ -0,0 +1,147 @@
...
+
+#=========================================================================
+# Block 4: Advance the model and harvest the synthetic observations.
+# output files are:
+# True_state.nc   ...... the DART state
```
timhoar, line 118: Is the creation of this still 'automatic'?  Does it depend on a namelist setting?
Is the file 'True_state.nc' or 'true_state.nc' or ... does it have a stage name ?

models/POP/shell_scripts/cesm2_1/perfect_model.csh
```diff
@@ -0,0 +1,147 @@
...
+${LAUNCHCMD} ${EXEROOT}/perfect_model_obs
+
+if ($status != 0) then
+   echo "ERROR ... DART died in 'perfect_model_obs' ... ERROR"
+   echo "ERROR ... DART died in 'perfect_model_obs' ... ERROR"
+   exit -4
+endif
+
+${MOVE} obs_seq.perfect  ../pop_obs_seq.${OCN_DATE_EXT}.perfect
+${MOVE} dart_log.out     ../pop_dart_log.${OCN_DATE_EXT}.out
```
timhoar, line 134: If you don't move the true state file ... you only have the last timestep at the end ... traditionally, the True_State.nc had the entire timeseries that generated all the observations.

models/POP/shell_scripts/cesm2_1/st_archive.sh
```diff
@@ -0,0 +1,400 @@
+#!/bin/sh
+#
+# This code is part of the CESM distribution,
+# So it is not protected by the DART copyright agreement.
+#
+# DART $Id$
```
timhoar, line 6: Please delete these lines.

models/POP/shell_scripts/cesm2_1/st_archive.sh
```diff
@@ -0,0 +1,400 @@
+#!/bin/sh
+#
+# This code is part of the CESM distribution,
+# So it is not protected by the DART copyright agreement.
```
timhoar, line 4: Is this different than the one under CESM? What is different? If this file exists in the CESM distrubution, I try to provide the full filename as a comment so people can see what is different.

models/POP/shell_scripts/cesm2_1/stitch_post.sh
```diff
@@ -0,0 +1,73 @@
...
+#
+# Copyright 2020 University Corporation for Atmospheric Research
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.
+# Please view the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+# ==============================================================================
+
+#Aim: stitch together filter output to make diagnostic files
```
timhoar, line 11: This needs a lot more commenting ... what directory do I run it in, the first real executable statement is 'remove all the netcdf files' ... that could be BAD ... how about some bulletproofing?

models/POP/shell_scripts/cesm2_1/stitch_post.sh
```diff
@@ -0,0 +1,73 @@
...
+wait
+
+## rename variables
+for var in 'SALT' 'TEMP' 'UVEL' 'VVEL' 'PSURF'
+do
+
+   ncrename -v $var'_CUR',$var $mean &
+   ncrename -v $var'_CUR',$var $sd &
+   ncrename -v $var'_CUR',$var $inf_mean &
+   ncrename -v $var'_CUR',$var $inf_sd &
```
timhoar, line 54: Why are we renaming the variables?

models/POP/shell_scripts/cesm2_1/stitch_post.sh
```diff
@@ -0,0 +1,73 @@
...
+
+   # concatenate the 4 files into analysis
+   # append
+   ncrcat -A -v $var $mean $sd $inf_mean $inf_sd analysis.nc
+   echo -n "done variable " $var "time "
+   date +"%T"
+done
+
+# switch time and copy dimensions back
+ncpdq -O -a time,copy analysis.nc analysis.nc
```
timhoar, line 66: Switch them back to what, and why?

models/POP/shell_scripts/cesm2_1/stitch_prior.sh
```diff
@@ -0,0 +1,73 @@
...
+#
+# Copyright 2020 University Corporation for Atmospheric Research
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.
+# Please view the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+# ==============================================================================
+
+#Aim: stitch together filter output to make diagnostic files
```
timhoar, line 11: This file is 99% the same as stich_post.sh ... can you make one script that takes an argument that would do the job of both files?

models/POP/shell_scripts/cesm2_1/user_datm.streams.txt.CPLHISTForcing.Solar_template
```diff
@@ -0,0 +1,47 @@
+<dataSource>
+    Raeder et al. 80 member CAM/DART reanalysis
+</dataSource>
```
timhoar, line 3: Ben ... please look at 
/glade/work/thoar/git/DART_CLM/models/clm/shell_scripts/cesm2_0/datm.streams.txt.CPLHISTForcing.Solar_complete
for an example of what I'd like to see in these.   When we were able to curate the files in the RDA, it became a lot more sensible to reference them there - and I added some curation information. I know the POP stream templates are very similar to what was in the repository, but now seems like a good time to update them.

models/POP/shell_scripts/cesm2_1/DART_params.csh
```diff
@@ -0,0 +1,216 @@
...
+# settings below this point - e.g. component namelists, history file options,
+# the processor layout, xml file options, etc - that you will almost certainly
+# want to change before doing a real science run.
+# ==============================================================================
+
+# The CESM compile step takes enough resource that cheyenne requires a wrapper
+# If your platform does not have this restriction, set BUILD_WRAPPER to '' 
+# setenv BUILD_WRAPPER ''
+setenv BUILD_WRAPPER "qcmd -q share -l select=1 -A ${ACCOUNT} --"
+setenv nodes_per_instance 2
```
timhoar, line 195: nodes_per_instance is set here and in the setup scripts.

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
```diff
@@ -0,0 +1,763 @@
...
+echo ''
+echo 'Invoking create_newcase.'
+echo ''
+echo '${cesmroot}/cime/scripts/create_newcase'
+echo '--case ${CASEROOT}'
+echo '--mach ${mach}'
+echo '--res ${resolution}'
+echo '--compset ${compset}'
+echo '--run-unsupported'
+echo '--pecount ${use_tasks_per_node}x${nthreads}'
```
timhoar, line 177: The single quotes prevent the expansion of the variable names. You must use double quotes.
This happens in multiple scripts,.

models/POP/shell_scripts/cesm2_1/DART_params.csh
```diff
@@ -0,0 +1,216 @@
...
+# ==============================================================================
+# Set the variables needed for the DART configuration.
+# dartroot     Location of the root of _your_ DART installation
+# BASEOBSDIR   Part of the directory name containing the observation sequence 
+#              files to be used in the assimilation. The observations are presumed
+#              to be stored in sub-directories with names built from the year and
+#              month. 'BASEOBSDIR' will be inserted into the appropriate scripts.
+# ==============================================================================
+
+setenv dartroot               /glade/work/${USER}/git/DART
```
timhoar, line 88: In all the setup scripts, etc., the variable is DARTROOT ... yet here it is lowercase.  my pmo setup bombed straightaway ... I think DARTROOT should be uppercase. I have no strong opinion on cesmdata or cesmroot or ...

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
```diff
@@ -0,0 +1,763 @@
...
+#
+# extra_string  An optional extra_string to differentiate a case from others
+#               that have the same compset and resolution.
+# CASE          The value of "case" will be used many ways; directory and file
+#               names both locally and on HPSS, and script names; so consider
+#               its length and information content.
+# ==============================================================================
+
+setenv extra_string ''
+setenv CASE         ${compset}.${resolution}${extra_string}
```
timhoar, line 104: I am not sure about this strategy ... it only works because you've chosen to use the shortest possible definition of a very specific compset ... 'G' ... this is not necessarily how it will be used in general.   In CLM and CAM, we choose to (have to?) specify the compset longname.  Your 'G' resolves to '2000_DATM%NYF_SLND_CICE_POP2_DROF%NYF_SGLC_WW3'
which is kinda awful to be part of a casename.


Compset longname is 2000_DATM%NYF_SLND_CICE_POP2_DROF%NYF_SGLC_WW3
Compset specification file is /glade/u/home/cmip6/cesm_tags/release-cesm2.1.1/cime/../components/pop//cime_config/config_compsets.xml
Compset forcing is 1972-2004
ATM component is  Data driven ATM COREv2 normal year forcing
LND component is Stub land component
ICE component is Sea ICE (cice) model version 5
OCN component is POP2 
ROF component is Data runoff modelCOREv2 normal year forcing:
GLC component is Stub glacier (land ice) component
WAV component is Wave Watch

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
```diff
@@ -0,0 +1,763 @@
...
+   echo "WARNING: the script to configure for data assimilation is not available."
+   echo "         CESM_DART_config should be present locally or in"
+   echo "         ${DARTROOT}/models/POP/shell_scripts/${cesmtagmajor}/"
+   echo "         You can stage this script later, but you must manually edit it"
+   echo "         to reflect the location of the DART code tree."
+endif
+
+# ==============================================================================
+# Configuring the case.
+# ==============================================================================
```
timhoar, line 222: Before configuring the case (which changes the env*xml files), we have made copies of the env*xml files with their original contents - for comparison, mostly.  Is this step not wanted? Is there some other way to compare ... some other copy to compare against?

# Save a copy for debug purposes
foreach FILE ( *xml )
   if ( ! -e        ${FILE}.original ) then
      ${COPY} $FILE ${FILE}.original
   endif
end

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
```diff
@@ -0,0 +1,763 @@
...
+echo "   ./xmlchange STOP_N=$STOP_N"
+echo ""
+echo "   and"
+echo "   ./case.submit"
+echo ""
+echo "Check the streams listed in the streams text files.  If more or different"
+echo 'dates need to be added, then do this in the $CASEROOT/user_*files*'
+echo "then invoke './preview_namelists' so you can check the information in the"
+echo "${CASEROOT}/CaseDocs or ${RUNDIR} directories."
+echo ""
```
timhoar, line 761: We ultimately decided that creating a 'here' document with these instructions and then cat'ting them out was a good idea. The document with the instructions then persists in the CASEROOT directory.
See cam-fv/shell_scripts/cesm2_1/setup_advanced
If you don't want to do that - its fine as it stands. Just a suggestion.

timhoar: Actually ... you do it in the CESM_DART_config ...

models/POP/shell_scripts/cesm2_1/CESM_DART_config
```diff
@@ -0,0 +1,317 @@
...
+# -- Submit the job using ${CASEROOT}/case.submit
+#
+# ==============================================================================
+# Get the environment of the case - defines number of instances/ensemble size ...
+# Each model component has their own number of instances.
+# ==============================================================================
+
+setenv cesmtag              cesm2_1_1
+# Trim the minor version off cesmtag to get the major release
+setenv cesmtagmajor `echo ${cesmtag} | cut -c1-7`
```
timhoar, line 50: I think there is reason to source the DART_params.csh here instead of setting the cesmtag, etc.
Furthermore, line 112 requires a DARTROOT environment varriable - which is only set in DART_params.csh

models/POP/shell_scripts/cesm2_1/perfect_model.csh
timhoar: Unless the input.nml:perfect_model_obs_nml:  write_output_state_to_file = .true.    there is no output. The input.nml in the repo has this as .false.

timhoar: More information:
When `write_output_state_to_file = .true.` the `perfect_restart` is created - as a netCDF file. The input.nml could/should have a .nc on this filename.
When `single_file_out = .true.`, the `true_state.nc` file is created. (the input.nml has this as `.false.` - not sure what you want as the default action.


```
   write_output_state_to_file = .true.
   single_file_out            = .true.
   output_state_files         = 'perfect_restart'
```

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_hybrid
```diff
@@ -0,0 +1,766 @@
...
+echo '--ninst ${num_instances}'
+echo '--multi-driver'
+
+${cesmroot}/cime/scripts/create_newcase \
+--case ${CASEROOT} \
+--mach ${mach} \
+--res ${resolution} \
+--compset ${compset} \
+--run-unsupported \
+--pecount ${use_tasks_per_node}x${nthreads} \
```
timhoar, line 189: I have no idea what this does.
It does not appear to impact the compilation time - and we manually clobber the pe setup with xmlchanges ... what does this do?

timhoar: I have confirmed that this only impacts the env*xml variables that we modify later.
I would prefer to remove the --pecount option from the create_newcase. I don't think it
serves a purpose and might confuse someone with regard to the xmlchanges on lines 256 to 264 below.
My first thought was that it was going to somehow adjust something for the build environment - that was apparently wrong.

models/POP/shell_scripts/cesm2_1/HPSS_get_ensemble
```diff
@@ -38,14 +38,8 @@ cd ${save_path}
...
     # Zero pad the year string so that there are 4 digits
-    if ($iyear < 10) then
-        setenv zero_filled_year 000${iyear}
-    else if ($iyear < 100) then
-        setenv zero_filled_year 00${iyear}
-    else if ($iyear < 1000) then
-        setenv zero_filled_year 0${iyear}
-    endif
-   
+    setenv zero_filled_year = printf %04d $iyear
```
timhoar, line 41: That's not quite right either ... 
```
setenv zero_filled_year = `printf %04d $iyear`
```

timhoar: Thank You ...

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
johnsonb-ucar: Ah this was an oversight. I copied these options from Kevin's CAM ensemble setup scripts -- I typically don't set pecounts as a create_newcase option. I'm going to remove it on the pmo script if that sounds good.

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
johnsonb-ucar: I'm editing the comment to update it. The default for G compsets, e.g. those using the COREv2 and JRA-55 streamfiles, is to have NO_LEAP calendars by default while the CAM reanalysis has a GREGORIAN calendar so the vestigial comment still applies somewhat.

models/POP/shell_scripts/cesm2_1/perfect_model.csh
johnsonb-ucar: 👍

models/POP/shell_scripts/cesm2_1/DART_params.csh
johnsonb-ucar: Removed it from DART_params.csh and left it in the setup scripts.

models/POP/shell_scripts/cesm2_1/DART_params.csh
johnsonb-ucar: Made it DARTROOT in all of the scripts.

models/POP/shell_scripts/cesm2_1/DART_params.csh
timhoar: Sorry ... it looks like `nodes_per_instance` still exists in DART_params.csh ... the other thing is that setting `nodes_per_instance` here is kinda the wrong place, as people might think it has something to do with the resources needed to build CESM - which is not true.

johnsonb-ucar: Yep, I haven't pushed to my fork yet. Apologies, I have been marking them as resolved when I've made the commit locally as a way of checking off the boxes as I'm processing of the revisions. I can go back and mark them as unresolved after I've pushed to my fork.

models/POP/shell_scripts/cesm2_1/DART_params.csh
```diff
@@ -0,0 +1,216 @@
...
+# the processor layout, xml file options, etc - that you will almost certainly
+# want to change before doing a real science run.
+# ==============================================================================
+
+# The CESM compile step takes enough resource that cheyenne requires a wrapper
+# If your platform does not have this restriction, set BUILD_WRAPPER to '' 
+# setenv BUILD_WRAPPER ''
+setenv BUILD_WRAPPER "qcmd -q share -l select=1 -A ${ACCOUNT} --"
+setenv nodes_per_instance 2
+setenv number_of_threads 1
```
timhoar, line 196: `number_of_threads` should be set in the same place as `nodes_per_instance` ... and both have nothing to do with the build_wrapper ...

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
johnsonb-ucar: Added an additional environmental variable `case_string` in `DART_params`.

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_pmo
```diff
@@ -0,0 +1,763 @@
...
+
+   # ===========================================================================
+   set fname = "user_nl_cice${inst_string}"
+   # ===========================================================================
+   # CICE Namelist
+   # Note well: this is true as of CESM2_1 but may change in the future.
+   # In a hybrid case, setting restart files for CICE is a distinct process from
+   # setting restart files for POP. The restart file is set in the user_nl_cice
+   # file in the ${CASEROOT} directory using the ice_ic (ice initial condition)
+   # key in the namelist file instead of using an rpointer file in ${RUNDIR}.
```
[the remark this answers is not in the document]
johnsonb-ucar: Edited out that clause so the section reads `In a hybrid case, setting restart files for CICE is set in the user_nl_cice file in the ${CASEROOT} directory using the ice_ic (ice initial condition) key in the namelist file instead of using an rpointer file in ${RUNDIR}.

models/POP/shell_scripts/cesm2_1/user_datm.streams.txt.CPLHISTForcing.Solar_template
johnsonb-ucar: No problem.

models/POP/shell_scripts/cesm2_1/assimilate.csh
johnsonb-ucar: Yes, this code fails in the second cycle. I removed the line `${REMOVE} obs_seq.out` which was included because it copied the logic of `/models/cam-fv/shell_scripts/cesm2_1/assimilate.csh.template`

models/POP/shell_scripts/cesm2_1/assimilate.csh
johnsonb-ucar: Fixed the comment.

models/POP/shell_scripts/cesm2_1/assimilate.csh
timhoar: Thanks ... the new version removes the cookie.

models/POP/shell_scripts/cesm2_1/CESM2_1_setup_hybrid
```diff
@@ -0,0 +1,770 @@
...
+
+# ==============================================================================
+# Updating the run script to reflect queue and wallclock
+# ==============================================================================
+
+echo ""
+echo "Updating the run script to set wallclock and queue."
+echo ""
+
+./xmlchange --subgroup case.st_archive --id JOB_WALLCLOCK_TIME --val $timewall
```
timhoar, line 361: I am putting in a comment here ... the short term archiver may take a LOT longer than the assimilation.
I am going to commit this as is, but we need to revisit it before creating a release.

timhoar:
This pull request was to change Manhattan. Since it is a feature (and not a bugfix), I think it is more appropriate to change the master branch. I took this pull request and merged it onto the master branch.
Since it did not appear to resolve this pull request (which was my hope), I am simply closing the pull request without merging it onto Manhattan.

PR #10, opened 2020-03-31
Add absolute humidity observation

myying:
Add quantity QTY_ABSOLUTE_HUMIDITY and observation type MPD_ABSOLUTE_HUMIDITY, along with obs_converters (convert_to_text.py processes original netCDF files for daily MPD profiles, then text_to_obs converts text file to obs_seq files)


assimilation_code/modules/observations/DEFAULT_obs_kind_mod.F90
```diff
@@ -489,10 +489,14 @@ module obs_kind_mod
   QTY_ION_O_MIXING_RATIO            = 365, &
   QTY_ATOMIC_H_MIXING_RATIO         = 366
 
+!!MYY: absolute humidity
```
timhoar, line 492: I know lots of people habitually put their initials next to the changes, but this is completely unneccesary.
'git blame <filename>'
makes it perfectly clear who modifed what and when.

observations/forward_operators/obs_def_abs_humidity_mod.f90
```diff
@@ -0,0 +1,186 @@
+! DART software - Copyright UCAR. This open source software is provided
+! by UCAR, "as is", without charge, subject to all terms of use at
+! http://www.image.ucar.edu/DAReS/DART/DART_download
+!
+! $Id$
```
timhoar, line 5: We can remove these two lines ... the Id was svn boilerplate that is not needed with git.

observations/forward_operators/obs_def_abs_humidity_mod.f90
```diff
@@ -0,0 +1,186 @@
...
+implicit none
+private
+
+public :: get_expected_absolute_humidity
+
+! version controlled file description for error handling, do not edit
+character(len=256), parameter :: source   = &
+   "$URL$"
+character(len=32 ), parameter :: revision = "$Revision$"
+character(len=128), parameter :: revdate  = "$Date$"
```
timhoar, line 58: These can be rewritten:
```
character(len=*), parameter :: source   = 'obs_def_abs_humidity_mod.f90'
character(len=*), parameter :: revision = ''
character(len=*), parameter :: revdate  = ''
```
since the svn keywords were stripped out when files got converted to git, we need to manually do this now.
We are abandoning the revision/revdate, so those don't need to be filled in.

observations/forward_operators/obs_def_abs_humidity_mod.f90
```diff
@@ -0,0 +1,186 @@
...
+!  interpolate the temperature to the desired location
+call interpolate(state_handle, ens_size, location, QTY_TEMPERATURE, tmpk, tmpk_istatus)
+call track_status(ens_size, tmpk_istatus, ah, istatus, return_now)
+
+where (tmpk <= 0.0_r8)
+   ah = missing_r8
+   istatus = 99
+endwhere
+
+if(all(istatus /= 0)) return  ! not using return_now because where clause modifies istatus.
```
timhoar, line 114: I suspect you copied this from a file that had this comment.  Since this is a new file, the comment is not correct and should be removed.  If you want to add a comment before this line, you could explain what is happenning next:
```
! If any of the interpolations failed, just return now
```

observations/forward_operators/obs_def_abs_humidity_mod.f90
```diff
@@ -0,0 +1,186 @@
...
+else
+   ! pressure comes back in pascals (not hPa or mb)
+   call interpolate(state_handle, ens_size, location, QTY_PRESSURE, pres, pres_istatus)
+   call track_status(ens_size, pres_istatus, ah, istatus, return_now)
+
+   where (pres <= 0.0_r8 .or. pres >= 120000.0_r8)
+      ah = missing_r8
+      istatus = 99
+   end where
+   if(all(istatus /= 0)) return  ! not using return_now because where clause modifies istatus.
```
timhoar, line 129: the trailing comment should be removed - 'return_now()' is not an issue.

observations/forward_operators/obs_def_abs_humidity_mod.f90
```diff
@@ -0,0 +1,186 @@
...
+
+if (first_time_warn_low) then
+   !if (any(ah < MIN_VALUE .and. istatus == 0)) then
+   !   write(msgstring, '(A,F12.6)') 'values lower than low limit detected, e.g.', minval(ah, istatus==0)
+      write(msgstring, *) 'checking absolute humidity value computed by the forward operator'
+      call error_handler(E_MSG,'get_expected_absolute_humidity', msgstring,      &
+                         text2='all values lower than 0 will be set to 0', &
+                         text3='this message will only print once')
+      first_time_warn_low = .false.
+   !endif
```
timhoar, line 157: should the if statement be there or not ... commenting out the stuff is not in keeping with our coding practice.

observations/forward_operators/obs_def_abs_humidity_mod.f90
```diff
@@ -0,0 +1,186 @@
...
+!----------------------------------------------------------------------------
+
+end module obs_def_abs_humidity_mod
+! END DART PREPROCESS MODULE CODE
+
+! <next few lines under version control, do not edit>
+! $URL$
+! $Id$
+! $Revision$
+! $Date$
```
timhoar, line 186: You can get rid of these, too.

observations/obs_converters/MPD/text_to_obs.f90
```diff
@@ -0,0 +1,206 @@
...
+                              set_copy_meta_data, set_qc_meta_data
+
+use obs_utilities_mod, only : create_3d_obs, add_obs_to_seq
+
+use      obs_kind_mod, only : MPD_ABSOLUTE_HUMIDITY
+
+implicit none
+
+character(len=64), parameter :: text_input_file = 'text.txt'
+character(len=64), parameter :: obs_out_file    = 'obs_seq.out'
```
timhoar, line 43: FYI:  character parameters can be specified with (len=*)
This is really compact and the variables never need to be 'trimmed' when being printed.

observations/obs_converters/MPD/text_to_obs.f90
```diff
@@ -0,0 +1,206 @@
...
+      ! make an obs derived type, and then add it to the sequence
+      call create_3d_obs(lat, lon, vert, VERTISHEIGHT, abs_humid, &
+                         MPD_ABSOLUTE_HUMIDITY, terr, oday, osec, qc, obs)
+      call add_obs_to_seq(obs_seq, obs, time_obs, prev_obs, prev_time, first_obs)
+      if (debug) print *, 'added abs humidity obs to output seq'
+
+   else !WHAT???
+
+      ! no method defined to convert this type
+      write(*,*) 'no method defined to convert this type = ', otype
```
timhoar, line 184: So is this an error or what? Should the program stop?

observations/obs_converters/MPD/work/convert_to_text.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+import numpy as np
+from netCDF4 import Dataset
+
```
timhoar, line 4: Please add DART boilerplate

observations/obs_converters/MPD/work/mkmf_preprocess
```diff
@@ -0,0 +1,18 @@
...
+../../../../build_templates/mkmf -p preprocess -t ../../../../build_templates/mkmf.template \
+ -a "../../../.." path_names_preprocess
+
+exit $status
+
+# <next few lines under version control, do not edit>
+# $URL$
+# $Revision$
+# $Date$
+
```
timhoar, line 18: At some point in my life, I am going to remove all these, so we don't need to add them to new files.
Please remove (and the DART $Id line at the top).  Thanks.

observations/forward_operators/obs_def_abs_humidity_mod.f90
nancycollins: i'm not sure about removing this comment.  generally we tell people to call track_status() and then test the logical 'return_now' to decide whether all the forward operators have failed and it's time to return.  in this case, return_now is not consistent anymore with the istatus codes if more have been set to fail.  this might be a useful comment - but maybe in fact it needs more information if it didn't help tim.

myying: I am removing the comment for now. These comments are inherited from obs_def_rel_humidity_mod.f90 from which I copied.

observations/forward_operators/obs_def_abs_humidity_mod.f90
myying: also inherited from obs_def_rel_humidity_mod.f90. I turned back on the if statement.

observations/obs_converters/MPD/text_to_obs.f90
myying: this should be a warning only, so I modified the warning message

observations/obs_converters/MPD/work/mkmf_preprocess
myying: i have removed all these

myying:
> usually there is no harm in making the lines REALLY long. At one point, we were concerned about things lining up in memory, so we chose to stagger them a bit by making the character strings to NOT be a multiple of 8, hence the 129 ...

Yeah, I just tested that it doesn't work for len=*, which is intriguing

observations/forward_operators/obs_def_abs_humidity_mod.f90
timhoar: That's what I thought.

timhoar:
OK - so I'm in the process of merging the PR and realize there is one thing really missing ...
what is 'MPD'? where does the data come from? Who do I credit?  Maybe a short 'README' would be appropriate.   I need to modify the list of available converters, and I have no idea what to say.

timhoar:
When I say 'who do I credit', I mean 'We normally provide ways to acknowledge the data provider somehow.' Who provides the data? That sort of thing - I know you get the credit for the contents of the PR

observations/obs_converters/MPD/text_to_obs.f90
```diff
@@ -0,0 +1,193 @@
...
+      if (debug) print *, 'got bad read code from input file, rcio = ', rcio
+      exit obsloop
+   endif
+
+   ! pull off the first value as an integer, to decode the type
+   read(input_line, *, iostat=rcio) otype
+   if (rcio /= 0) then
+      if (debug) print *, 'got bad read code trying to get obs type, rcio = ', rcio
+      exit obsloop
+   endif
```
timhoar, line 131: If there is only 1 type in the file ... do we need this logic at all?
Will the MPD python script ever make a text file with multiple observation types in it? Should the code issue a warning that it is just going to exit now and process what it has and ignore the rest of the file (which is what it is doing).

PR #14, opened 2020-04-22
add absolute humidity observation support

timhoar:
Michael Ying added observation conversion and assimilation support for absolute humidty observa


timhoar:
The MPD/README.md  has two choices -one block is if we distribute an example dataset with DART, the other is if we put the test dataset on the web and provide a URL.

observations/obs_converters/MPD/work/convert_to_text.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
+#
+# DART software - Copyright UCAR. This open source software is provided
+# by UCAR, "as is", without charge, subject to all terms of use at
+# http://www.image.ucar.edu/DAReS/DART/DART_download
+
+import numpy as np
```
johnsonb-ucar, line 7: The libraries needed to scientific computing in python (working with large arrays and netCDF) are not part of the standard libraries. They are already installed on glade can be loaded via:
> module load python
> ncar_pylib
On your local machine they can be installed via:
> pip install numpy
> pip install netCDF4
The latter library is maintained by Unidata. Documentation here:
http://unidata.github.io/netcdf4-python/netCDF4/index.html

PR #16, opened 2020-05-15
Adding a science nugget about MPAS

timhoar:
This has the permission of So-Young, the content came from:
https://mpas-dev.github.io/dart_atmosphere/dart_atmosphere.html


docs/pages/Research.md
```diff
@@ -58,6 +59,91 @@ Kevin Raeder, raeder@ucar.edu</td>
...
+</td>
+<td style="vertical-align: top">The MPAS/DART interface is
+built on MPAS's unstructured centroidal Voronoi mesh, 
+and does not regrid to latitude-longitude grids. In MPAS, the finite-volume 
+approach based on a C-grid staggering retains prognostic equations for mass 
+at the center of finite-volume cells and for the normal component of velocity
+(<i>u</i>) at the faces (or edges in 2D) of the cells. 
+The normal component of velocity (<i>u</i>) at cell edges is then used to 
+reconstruct zonal and meridional winds at cell centers 
+(<img src="../images/MPAS_V_vector.png" style="width: 20px; height: 22px;" alt="" />)
```
johnsonb-ucar, line 114: For consistency, should we put (<i>v</i>) in this line instead of trying to link to the vector "v" png?

johnsonb-ucar:
Other than the png italicized "v" on line 114 of Research.md, this PR looks good to me.

timhoar:
I copied that from the original source.
Do you know how to make an inline 'v' that matches the figure?

Tim Hoar
Data Assimilation Research Section
National Center for Atmospheric Research
thoar@ucar.edu
303.497.1708

johnsonb-ucar:
The most commonly used library (as far as I know) for rendering mathematical symbols is called MathJax:
https://github.com/mathjax/MathJax
It will render the symbols properly once converting from markdown to HTML, but will render them in their original LaTeX syntax within Github, so it's not a perfect solution.

timhoar:
So ... since the (admittedly awkward) solution in place (using a .png) only
looks bad when you view the source, not the rendered image or the .html ...
what are you proposing?
Is embedding LaTeX any better - considering it does not render until
converted to .html?

I don't know what you want me to do here.

Tim Hoar
Data Assimilation Research Section
National Center for Atmospheric Research
thoar@ucar.edu
303.497.1708

johnsonb-ucar:
Ah, my apologies -- if the png v renders on the page, I think it's fine. When I was previewing the change the png doesn't show up so it just looks like "( )".

timhoar:
GRRRRRR.....    It renders correctly in ATOM, but it is the wrong size on
GitHub ... and changes how 'wrong' based on what size font you are using
... boo hiss grr ... I'll think of something or reword it such that
we don't need it.

Tim Hoar
Data Assimilation Research Section
National Center for Atmospheric Research
thoar@ucar.edu
303.497.1708

docs/pages/Research.md
timhoar: Guess what? Because its in an html table, none of the markdown syntax for highlighting/bolding works - I guess this is not too surprising.  So - I just reverted to html ...
<strong><i>V</i></strong>
There is no overbar to denote a vector, but it should get the point across. If this is acceptable, let me know and I'll remove the images/MPAS_V_vector.png file from the repo.

johnsonb-ucar:
The revised (**_V_**) looks fine by me. I might try to mock up some equations from DART Lab over the weekend to see if it plays nicely with Github.

PR #56, opened 2020-09-23
fix z'0123' without an explicit int() or real() intrinsic

nancycollins:
the fortran standard says defining a binary, octal or hexadecimal constant
is not allowed without using one of the numeric intrinsics like int(val, kind)
or real(val, kind) to specify the kind explicitly.  gcc10 now flags this as
an error, which is new.  fix the two places where 'z' constants were being used.

### DIFF
--- a/assimilation_code/modules/utilities/time_manager_mod.f90
+++ b/assimilation_code/modules/utilities/time_manager_mod.f90
@@ -3148,6 +3148,12 @@ function generate_seed(timestamp)
 ! expected to be used to seed a random number generator in a way
 ! that you can reproduce the same sequence if seeded again from
 ! the same time value.
+!
+! the return value needs to be an i4 since seeds are only i4. 
+! compute total number of seconds using a double integer (i8) and 
+! return the least significant 32 bits.  a simple assignment could
+! overflow an i4, and the seed needs to be as unique as possible
+! so preserving the least significant digits is the better choice.
 
 type(time_type), intent(in) :: timestamp
 integer                     :: generate_seed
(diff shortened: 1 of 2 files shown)


assimilation_code/modules/utilities/time_manager_mod.f90
```diff
@@ -3159,7 +3159,7 @@ function generate_seed(timestamp)
 
 call get_time(timestamp, seconds, days)
 
-generate_seed = iand((secs_day * days) + seconds, z'00000000FFFFFFFF')
+generate_seed = iand((secs_day * days) + seconds, int(z'00000000FFFFFFFF',i8))
```
timhoar, line 3162: A little explanation on the algorithm would be good here.  Initially, I could not figure out why it was cast as an i8, but then 'secs_day' is an i8, so I got it. Then I thought ... but generate_seed is not i8.
So the 'iand' function is just to rip off the lowest (32) bit integer portion which guarantees the i8 value won't overflow the i4 variable?

nancycollins: pretty much correct.  you can't make unique seeds for every possible time since the seed is only an i4 (32 bits) while times can have more values than that.  but in practice the times used in a single run fall into a limited range.  the mask prevents the compiler from getting to choose how to put an i8 into an i4 and prevents an overflow error if running with floating point exceptions enabled.  i'm preserving the lower bits because that's where the differences between close times will show up.      

i added a paragraph in the comment section for this routine trying to say pretty much this.  let me know if it's enough or if i can be clearer.

PR #57, opened 2020-09-25
Adding the content from public PR 56

timhoar:
This addresses the gnu 10 compiler discovery that
we were not following the standard for declaring
hexadecimal constants.


johnsonb-ucar:
Looks good.

PR #58, opened 2020-10-16
Update Publications, adding two for 2019 & 2020

johnsonb-ucar:


docs/pages/Publications.md
```diff
@@ -26,9 +26,31 @@ The Data Assimilation Research Testbed: A Community Facility.
 *Bulletin of the American Meteorological Society*, **90**, 1283-1296,
 [doi:10.1175/2009BAMS2618.1](http://dx.doi.org/10.1175/2009BAMS2618.1)
 
-----
+---
+## 2020
+---
```
timhoar, line 31: This just in ... can you please add one more:

> Article ID: JAME21256
> Article DOI: 10.1029/2020MS002176
> Internal Article ID: 16918191
> Article: An EnOI-based Data Assimilation System with DART for a High-Resolution Version of the CESM2 Ocean Component
> Journal: Journal of Advances in Modeling Earth Systems

https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/2020MS002176

johnsonb-ucar: Yes, no problem. I just added Castruccio et al. 2020 to the Publications page. Please note page numbers have yet to be assigned for Castruccio et al. 2020 and Moore et al. 2020 so we'll need to do an additional edit once those become available.

timhoar (approved):
Looks good to me.
If you have no further mods, please go ahead and  merge.

timhoar:
As to the workflow ... I would have merged the PR you initiated ... but I did not know if you were 'done'.
We should talk about who/when could/should merge PRs...

johnsonb-ucar:
Yep. I think codifying a process is a good idea.

jlaucar:
This seems like something that should be pointed to from our new Institutional Memory page.

PR #59, opened 2020-10-20
bugfixing hex constants, GSI2DART converter

timhoar:


timhoar:
This has conflicts that must be resolved (obviously) and then it will get tagged as a bugfix ... v9.8.1 ... which means the 
docs/data/mainmenu.yml   has to be updated to 9.8.1

johnsonb-ucar:
Getting to this today. Apologies, I just didn't have the bandwidth for the past two days.

PR #60, opened 2020-10-20
One-line fix to bypass posterior inflation update if RTPS option

weather4evr:
When doing posterior state-space inflation, to update inflation, filter_assim is called a second time with inflate_only = .true.  This step is unnecessary if using RTPS inflation, as there's nothing to update when using RTPS.


nancycollins (approved):
looks good to me.

PR #62, opened 2020-10-22
Adding Obs Network choice functionality to L96_inf

mgharamti:
The changes are summarized below

(i) The user can now select between 6 different obs networks:
- Full: Observe all 40 variables
- Half: Observe every other variable
- Quarter: Observer every 4th variable
- 1:20: Observe the first 20 variables
- 10:30: Observe the middle 20 variables
- 1:10; 30:40: Observe the first and last 0 variables

This should allow the users to understand the interaction between localization,
inflation and the density of the observations. A lot can be learnt from this!

(ii) Improved the functionality of the polar plot:
- The plot is now annotated from 0 to 40 rather than 0 to 360 degrees
- Removed useless annotation in the center of the plot
- The xlabel on the inflation plot is edited accordingly
- The position of the figures have been adjusted to fit the new obs functionality

(iii) Variables in the code have been edited to mimic the fortran code
- Inflation variables inside 'run_lorenz_96_inf.m' and 'update_inflate.m'
have been renamed such that they match those in 'assim_tools_mod.f90' and
'adaptive_inflate_mod.f90'


docs/DART_LAB/matlab/run_lorenz_96_inf.m
```diff
@@ -165,6 +166,101 @@
...
+    'FontWeight'        ,'bold', ...
+    'SelectionChangeFcn', @ObsNetwork_selection);
+
+% Create the four radio buttons in the button group.
+handles.ui_radio_full = uicontrol(handles.ui_button_group_obsnetwork, ...
+    'Style'             , 'radiobutton', ...
+    'Units'             , 'Normalized', ...
+    'Position'          , [0.12 0.6 0.4 0.4], ...
+    'BackgroundColor'   , atts.background, ...
+    'String'            , 'Full', ...
```
timhoar, line 197: I wonder if we should label these as
1:40:1   1:40:2  1:40:4   instead of   Full Half Quarter ... 

I also spent a little time (unsuccessfully) trying to figure out how to have tickmarks at the outer perimeter of the observation locations. It would be nice if you saw the observation locations - throughout the assimilation cycles, not just when you have observations.

docs/DART_LAB/matlab/private/polar_dares.m
```diff
@@ -142,21 +142,21 @@
                 'HandleVisibility', 'off', 'Parent', cax);
         end
         
-        % draw radial circles
-        c82 = cos(82 * pi / 180);
-        s82 = sin(82 * pi / 180);
+        % draw radial circles (remove annotation; not useful here!)
```
timhoar, line 145: Can we change this to
```
        % draw radial circles for looks, no need to annotate
```

docs/DART_LAB/matlab/private/polar_dares.m
```diff
@@ -142,21 +142,21 @@
...
+%         c82 = cos(82 * pi / 180);
+%         s82 = sin(82 * pi / 180);
```
timhoar, line 147: Delete these.

docs/DART_LAB/matlab/private/polar_dares.m
```diff
@@ -142,21 +142,21 @@
...
         rinc = (rmax - rmin) / rticks;
         for i = (rmin + rinc) : rinc : rmax - 1
             hhh = line(xunit * i, yunit * i, 'LineStyle', ls, 'Color', tc, 'LineWidth', 1, ...
                 'HandleVisibility', 'off', 'Parent', cax);
-            text((i + rinc / 20) * c82, (i + rinc / 20) * s82, ...
-                ['  ' num2str(i)], 'VerticalAlignment', 'bottom', ...
-                'HandleVisibility', 'off', 'Parent', cax);
+%             text((i + rinc / 20) * c82, (i + rinc / 20) * s82, ...
+%                 ['  ' num2str(i)], 'VerticalAlignment', 'bottom', ...
+%                 'HandleVisibility', 'off', 'Parent', cax);
```
timhoar, line 154: delete these

timhoar (approved):
This looks good ... I do have some suggestions, take them or leave them.

docs/DART_LAB/matlab/private/polar_dares.m
timhoar: I guess, I just wanted this block to look like:
```
        % draw radial circles for looks, no need to annotate
        rinc = (rmax - rmin) / rticks;
        for i = (rmin + rinc) : rinc : rmax - 1
            hhh = line(xunit * i, yunit * i, 'LineStyle', ls, 'Color', tc, 'LineWidth', 1, ...
                'HandleVisibility', 'off', 'Parent', cax);
        end
        set(hhh, 'LineStyle', '-'); % Make outer circle solid
```

timhoar:
On a ruler there are big ticks and little ticks.
On the polar plot, the radial spokes are big ticks.
I think the plot is big enough that there should be some indication of where the state variables are located - like little ticks.
I thought ticks were hard, just a symbol would be ok.
Then I thought ... maybe we should have the ticks where the observations are going to be, instead of ticking the state locations.
Just thinking out loud ... the GUI certainly stands on its own.

As an aside ...  I don't think we need the run_lorenz_96.m anymore ...

docs/DART_LAB/matlab/run_lorenz_96_inf.m
timhoar: ACK ... When I 'reset', the observation network button was still on 1:20, but I got a full observation network when assimilating.

timhoar: ALSO ... and I don't know where ... the annotated default inflation mean is 0.6 ... below the inflation min of 1.0 ...
not sure either of these is right.

timhoar (changes requested):

docs/DART_LAB/matlab/run_lorenz_96_inf.m
```diff
@@ -39,19 +39,20 @@
...
 
 LOG_FILE = strcat(mfilename, '.log');
 
 first_call_to_reset = true;
 
 atts = stylesheet; % get the default fonts and colors
 
-figWidth  = 990;   % in pixels
-figHeight = 820;    % in pixels
+figWidth  = 1000;    % in pixels
```
timhoar, line 54: Actually ... please make this 1024
on my machine, 1000 makes the text box above the 'Inf Sd.' gets clipped to simply 'Inf' ...

timhoar:
There is one more thing ... the concentric circles for delineating the observation/state value are so faint as to be useless.
In the polar_dares.m, the 'tc' variable comes out at [0.875 0.875 0.875] ... for no real good reason, in my opinion.
How do you feel about changing that to [0.7 0.7 0.7]?

It is only used by run_lorenz_96_inf.m and run_lorenz_96.m   (by using plot_polar.m)

docs/DART_LAB/matlab/run_lorenz_96_inf.m
mgharamti: Will do!

mgharamti (commented):
> There is one more thing ... the concentric circles for delineating the observation/state value are so faint as to be useless.
> In the polar_dares.m, the 'tc' variable comes out at [0.875 0.875 0.875] ... for no real good reason, in my opinion.
> How do you feel about changing that to [0.7 0.7 0.7]?
> 
> It is only used by run_lorenz_96_inf.m and run_lorenz_96.m (by using plot_polar.m)

That's a good idea.

timhoar:
One more one more thing ... 

I just advanced for 20 timesteps, changed the forcing, turned on the EAKF and then changed the localization.
Changing the localization wiped out the existing graphic (but apparently the ensemble state was saved).
I took one 'Assimilate Obs' and the ensemble was displayed (correctly, I think).  Is that OK?

mgharamti:
> One more one more thing ...
> 
> I just advanced for 20 timesteps, changed the forcing, turned on the EAKF and then changed the localization.
> Changing the localization wiped out the existing graphic (but apparently the ensemble state was saved).
> I took one 'Assimilate Obs' and the ensemble was displayed (correctly, I think). Is that OK?

Good find Tim!! It took me a little to figure a way to fix it :) Now, I believe it's OK. Please give it a try.

PR #73, opened 2020-11-18
Twopubs agu2020

timhoar:
Adding two publications and adding enough whitespace at the end of the lines to
force the title and journal to be on new lines.


kdraeder (approved):
I looked at the details of the Sampling Error Correction reference.
It looks good to me.
I also scanned through the formatting changes, which don't raise any objections.

johnsonb-ucar (approved):

PR #95, opened 2020-12-11
mpas rttov 

hkershaw-brown:
Aim of this pull request is to commit the work Nancy and Soyoung did on MPAS and RTTOV. 

Fixes https://github.com/NCAR/DART_development/issues/119

The rttov obs_def changes:
* E_ALLMSG used instead of E_MSG
* loc_sea variable used to create `sfc`, `2m`, `10m` locations relative to model surface elevation
* monotonically decreasing pressure from TOA check now has greater than or **equal to**  ( this is the bug fix)

The mpas_atm model_mod changes:
* Debugging info in the comment at the start of the module. 
* New error code for pressure not monotonically decreasing with level
* QTY_CLOUD_FRACTION added.  Force non-negative hydrometeors and for QTY_CLOUD_FRACTION <1
* `compute_scalar_with_barycentric` passing `tavs(1)`, `values(1,:)`. when `n=1`.  There is some discussion of this in https://github.com/NCAR/DART_development/issues/119.   It is a bit of a gotcha, so is on the list of things to tidy up in the mpas_atm module.
* theta_to_tk has debugging info for non-zero istatus going in

There is some discussion on VERTISUNDEFINED in the comments (whether there should be the error code 82).  

Tested these changes on Cheyenne with a 10 member version on Soyoung's mpas case. Note it was a 10 member test since each restart is 3.7GB.  `/glade/p/cisl/dares/hkershaw/mpas_rttov_testing`

### DIFF
--- a/models/mpas_atm/model_mod.f90
+++ b/models/mpas_atm/model_mod.f90
@@ -1156,7 +1156,6 @@ logical  :: goodkind, surface_obs
 real(r8) :: lpres(ens_size), values(3, ens_size)
 real(r8) :: llv(3)    ! lon/lat/vert
 integer  :: e, verttype
-integer  :: ierror(ens_size)   ! Ha
 
 if ( .not. module_initialized ) call static_init_model
 
@@ -1901,7 +1900,7 @@ if (vertical_localization_on()) then
       call convert_vert_distrib(state_handle, 1, location_arr, base_qty, vert_localization_coord, istat_arr)
       istatus1 = istat_arr(1)
       base_loc = location_arr(1)
-      if(debug > 5 .and. do_output()) then
+      if(debug > 4 .and. do_output()) then
          call write_location(0,base_loc,charstring=string1)
          call error_handler(E_ALLMSG, 'get_close_obs: base_loc',string1,source, revision, revdate)
      endif
--- a/models/mpas_atm/model_mod.f90
+++ b/models/mpas_atm/model_mod.f90
@@ -1105,7 +1105,6 @@ end subroutine find_mpas_dims
 !>       ISTATUS = 99:  general error in case something terrible goes wrong...
 !>       ISTATUS = 81:  Vertical location too high
 !>       ISTATUS = 80:  Vertical location too low
-!>       ISTATUS = 82:  Vertical location is missing
 !>       ISTATUS = 88:  this kind is not in the state vector
 !>       ISTATUS = 89:  tk cannot be computed.
 !>       ISTATUS = 11:  Could not find the closest cell center that contains this lat/lon
@@ -4593,7 +4592,7 @@ enddo
 ! if the entire ensemble has missing vertical values we can return now.
 ! otherwise we need to continue to convert the members with good vertical values.
 if (all(zin == missing_r8)) then
-   istatus(:) = 0 ! vertisundef might come here - nsc  82    ! FIXME: Ha - Why should this be zero? I created 82 if all zin are missing.
+   istatus(:) = 0 
    return
 endif
 
@@ -4937,7 +4936,7 @@ enddo
 ! if all the vertical localization coord values are missing, 
 ! we don't call this routine again, and return.
 if (all(zin == missing_r8)) then ! .or. on_bound) then
-   istatus(:) = 82    ! FIXME: Ha - Why should this be zero? I created 82 if all zin are missing.
+   istatus(:) = 0
    return
 endif
 
--- a/models/mpas_atm/model_mod.f90
+++ b/models/mpas_atm/model_mod.f90
@@ -7451,6 +7451,15 @@ integer :: e
 qv_nonzero = max(qv,0.0_r8)
 theta_to_tk = missing_r8
 
+if (any(istatus /= 0)) then
+   print *, 'theta_to_tk - nonzero istatus coming in'
+   do e = 1, ens_size
+    if (istatus(e) /= 0) then
+        write(string2, *) 'member ', e, ' incoming istatus = ', istatus(e)
+        call error_handler(E_ALLMSG, 'theta_to_tk',string2,source, revision, revdate)
+    endif  !(istatus(e) /= 0) then
+   enddo   ! ens_size
+endif
 where (istatus == 0)
 
    theta_m = (1.0_r8 + 1.61_r8 * qv_nonzero)*theta
--- a/models/mpas_atm/model_mod.f90
+++ b/models/mpas_atm/model_mod.f90
@@ -1242,7 +1242,7 @@ else
          goodkind = .true.
       case (QTY_PRESSURE)   ! surface pressure should be in the state
          goodkind = .true.
-      case (QTY_SKIN_TEMPERATURE, QTY_SURFACE_TYPE, QTY_CLOUD_FRACTION)   ! Ha: added for rttov
+      case (QTY_SKIN_TEMPERATURE, QTY_SURFACE_TYPE, QTY_CLOUD_FRACTION)   
          goodkind = .true.
       case (QTY_SPECIFIC_HUMIDITY, QTY_2M_SPECIFIC_HUMIDITY)
          goodkind = .true.
@@ -1349,7 +1349,6 @@ else if (obs_kind == QTY_GEOPOTENTIAL_HEIGHT) then
      if(istatus(e) == 0) expected_obs(e) = query_location(location_tmp(e), 'VLOC')
    enddo
 
-   print *, 'model_interpolate: GEOPOTENTIAL_HEIGHT ', istatus(1), expected_obs(1), trim(locstring)
    if ( all(istatus /= 0 ) ) goto 100
 
 else if (obs_kind == QTY_VAPOR_MIXING_RATIO      .or. obs_kind == QTY_2M_SPECIFIC_HUMIDITY   .or. &
@@ -1395,7 +1394,7 @@ else if (obs_kind == QTY_SPECIFIC_HUMIDITY) then
       print *, 'model_interpolate: ', trim(get_name_for_quantity(obs_kind)), istatus(1), &
                                       expected_obs(1), trim(locstring)
 
-! Ha: Only for the variables NOT included in the dart state vector.
+! Only for the variables NOT included in the dart state vector.
 ! Anything included in the state_handle should not go into the case right down here.
 ! ex. QTY_SURFACE_PRESSURE will be taken care of in the else statement (one further down)
 ! as a generic interpolation case thru compute_scalar_with_barycentric
@@ -1461,12 +1460,11 @@ do e = 1, ens_size
          write(string1,*) 'interp routine returned a negative status which is an illegal value'
       else if (istatus(e) /= 0 .and. expected_obs(e) /= MISSING_R8) then
          write(string1,*) 'interp routine returned a bad status but not a MISSING_R8 value'
-         expected_obs(e) = MISSING_R8  ! Ha
+         expected_obs(e) = MISSING_R8  
       else
          write(string1,*) 'interp routine returned a good status but set value to MISSING_R8'
       endif
 
-      !call error_handler(E_ERR,'model_interpolate', string1, source,revision,revdate, &
       call error_handler(E_ALLMSG,'model_interpolate', string1, source,revision,revdate, &
                          text2=string2, text3=string3)
    endif
@@ -3332,8 +3330,8 @@ call nc_get_variable(ncid, 'dcEdge',        dcEdge,        routine)
 call nc_get_variable(ncid, 'zgrid',         zGridFace,     routine)
 call nc_get_variable(ncid, 'cellsOnVertex', cellsOnVertex, routine)
 call nc_get_variable(ncid, 'xland',         xland,         routine)
-call nc_get_variable(ncid, 'seaice',        seaice,        routine)  ! Ha: added for rttov
-call nc_get_variable(ncid, 'skintemp',      skintemp,      routine)  ! Ha: added for rttov
+call nc_get_variable(ncid, 'seaice',        seaice,        routine)
+call nc_get_variable(ncid, 'skintemp',      skintemp,      routine)
 
 dxmax = maxval(dcEdge)  ! max grid resolution in meters
 
@@ -6056,7 +6054,6 @@ call find_vert_level(state_handle, ens_size, loc, nc, c, .true., lower, upper, f
 
 if (debug > 9 .and. do_output()) then
    write(string3,*) 'ier = ',ier(1), ' triangle = ',c(1:nc), ' vert_index = ',lower(1:nc, 1)+fract(1:nc, 1), ' nc = ', nc
-!   write(string3,'(A,I5,A,3I10,A,3F7.2)') 'ier = ',ier(1), ' triangle = ',c(1:nc), ' vert_index = ',lower(1:nc, 1)+fract(1:nc, 1)
    call error_handler(E_MSG, 'find_vert_indices', string3, source, revision, revdate)
 endif
 
@@ -6339,7 +6336,7 @@ select case (use_rbf_option)
                          source, revision, revdate)
 end select
 
-! Ha: Check if any of edges are located in the boundary zone.
+! Check if any of edges are located in the boundary zone.
 ! (We will skip the obs if any edges are located there.)
 if (on_boundary_edgelist(edge_list)) then
    nedges = -1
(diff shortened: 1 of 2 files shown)


models/mpas_atm/model_mod.f90
```diff
@@ -1319,40 +1349,57 @@ subroutine model_interpolate(state_handle, ens_size, location, obs_type, expecte
      if(istatus(e) == 0) expected_obs(e) = query_location(location_tmp(e), 'VLOC')
    enddo
 
-else if (obs_kind == QTY_VAPOR_MIXING_RATIO) then 
+   print *, 'model_interpolate: GEOPOTENTIAL_HEIGHT ', istatus(1), expected_obs(1), trim(locstring)
```
timhoar, line 1352: Should this _always_ print?

models/mpas_atm/model_mod.f90
```diff
@@ -1424,14 +1476,18 @@ subroutine model_interpolate(state_handle, ens_size, location, obs_type, expecte
...
 
    if (.not. expected_obs(e) < 0 .and. .not. expected_obs(e) >= 0) then
-      write(string1,*) 'member ', e, ' expected obs may be NaN: ', expected_obs(e)
-      call error_handler(E_ERR,'model_interpolate', string1, source,revision,revdate)
+      write(string1,*) 'Skip member ', e, ' expected obs may be NaN: ', expected_obs(e), &
+      ' obs_kind', obs_kind,' istatus = ', istatus(e)
+      write(string2,*) 'cellid:',cellid,' location ', trim(locstring)
+      expected_obs(e) = MISSING_R8
+      istatus(e) = 99
+      call error_handler(E_ERR,'model_interpolate', string1, source,revision,revdate,text2=string2)
```
timhoar, line 1484: Is this the standard way to check for NaN ... I thought the test was basically to see if the quantity did not equal itself, it must be a NaN.  i.e.
```
if (expected_obs(e) /= expected_obs(e)) then
   it must be a NaN.
else
   all good
endif
```

models/mpas_atm/model_mod.f90
```diff
@@ -7375,20 +7500,36 @@ subroutine compute_full_pressure(ens_size, theta, rho, qv, pressure, tk, istatus
 real(r8), dimension(ens_size), intent(in)  :: qv       ! water vapor mixing ratio [kg/kg]
 real(r8), dimension(ens_size), intent(out) :: pressure ! full pressure [Pa]
 real(r8), dimension(ens_size), intent(out) :: tk       ! return sensible temperature to caller
-integer,  dimension(ens_size), intent(in)  :: istatus_in
+integer,  dimension(ens_size), intent(inout):: istatus
 
 ! Local variables
 real(r8), dimension(ens_size) :: qv_nonzero
+!integer,  dimension(ens_size) :: ierr   ! Ha
```
timhoar, line 7507: Can we delete this line ?

models/mpas_atm/model_mod.f90
```diff
@@ -5951,12 +6054,13 @@ subroutine find_vert_indices (state_handle, ens_size, loc, nc, c, lower, upper,
 ! need vert index for the vertical level
 call find_vert_level(state_handle, ens_size, loc, nc, c, .true., lower, upper, fract, ier)
 
-if (debug > 10 .and. do_output()) then
-   write(string3,'(A,I5,A,3I10,A,3F7.2)') 'ier = ',ier(1), ' triangle = ',c(1:nc), ' vert_index = ',lower(1:nc, 1)+fract(1:nc, 1)
+if (debug > 9 .and. do_output()) then
+   write(string3,*) 'ier = ',ier(1), ' triangle = ',c(1:nc), ' vert_index = ',lower(1:nc, 1)+fract(1:nc, 1), ' nc = ', nc
+!   write(string3,'(A,I5,A,3I10,A,3F7.2)') 'ier = ',ier(1), ' triangle = ',c(1:nc), ' vert_index = ',lower(1:nc, 1)+fract(1:nc, 1)
```
timhoar, line 6059: Please remove.

models/mpas_atm/model_mod.f90
```diff
@@ -5688,7 +5790,7 @@ subroutine compute_scalar_with_barycentric(state_handle, ens_size, loc, n, ival,
       ! go around triangle and interpolate in the vertical
       ! c(3) are the cell ids
 
-      low_offset = (c(i)-1) * nvert
+      low_offset = (c(i)-1) * nvert !HK low_offset and upp_offset are the same?
```
timhoar, line 5793: Has this been answered?

models/mpas_atm/model_mod.f90
```diff
@@ -5414,8 +5508,8 @@ subroutine find_pressure_bounds(state_handle, ens_size, p, cellid, nbounds, &
 where(ier(:) == 0) ier(:) = temp_ier(:)
 
 ! Check for out of the column range
-where(p(:) < pressure(nbounds, :)) ier(:) = 81
-!where(p(:) > pressure(      1, :)) ier(:) = 80
+where(p(:) < pressure(nbounds, :)) ier(:) = 81   ! too high
+!where(p(:) > pressure(      1, :)) ier(:) = 80  ! too low -> just take the lowest level
```
timhoar, line 5512: Does this mean error code of 80 is never used ... should it not exist?

models/mpas_atm/model_mod.f90
```diff
@@ -5052,7 +5143,7 @@ subroutine convert_vert_distrib_state(state_handle, ens_size, location, quantity
 101 continue
 
      if (debug > 9 .and. do_output()) then
-        write(string2,'("zout_in_scaleheight:",F10.2)') zout
+        write(string2,'("zout_in_scaleheight:",F10.2)') zout(1)
```
timhoar, line 5146: zout is a 1D array 'ens_size', printing out just one value is not necessarily the value with the problem.
This is in a fairly high debug level ... If they are digging that deep, they may want all the values.

models/mpas_atm/model_mod.f90
```diff
@@ -5017,11 +5100,19 @@ subroutine convert_vert_distrib_state(state_handle, ens_size, location, quantity
...
+        call compute_full_pressure(ens_size, values(1,:), values(2,:), values(3,:), fullp, tk, istatus)
+        if( all(istatus /= 0) ) then
+           if (debug > 1 .and. do_output()) then
+           write(string2,'("Failed in compute_full_pressure")')
+           call error_handler(E_MSG, 'convert_vert_distrib_state',string2,source, revision, revdate)
+           endif
+        else
+           if (debug > 9 .and. do_output()) then
+           write(string2,'("zout [Pa],theta,rho,qv:",3F10.2,F12.8)') fullp(1), values(1:3,1)
+           call error_handler(E_MSG, 'convert_vert_distrib_state',string2,source, revision, revdate)
```
timhoar, line 5112: See following comment about 'zout' ... minimally, string2 should contain something about it only printing the infor for ensemble member 1.

models/mpas_atm/model_mod.f90
```diff
@@ -4870,7 +4939,7 @@ subroutine convert_vert_distrib_state(state_handle, ens_size, location, quantity
 ! if all the vertical localization coord values are missing, 
 ! we don't call this routine again, and return.
 if (all(zin == missing_r8)) then ! .or. on_bound) then
-   istatus(:) = 0
+   istatus(:) = 82    ! FIXME: Ha - Why should this be zero? I created 82 if all zin are missing.
```
timhoar, line 4942: I think this is fixed? remove the FIXME?

models/mpas_atm/model_mod.f90
```diff
@@ -4950,7 +5026,7 @@ subroutine convert_vert_distrib_state(state_handle, ens_size, location, quantity
    endif
 
    if (debug > 9 .and. do_output()) then
-      write(string2,'("zout_in_height:",F10.2)') zout
+      write(string2,'("zout[m]:",F10.2)') zout(1)
```
timhoar, line 5029: Actually `zout[1]` ....

timhoar (commented):
Could delete some commented-out lines.
I'm not a fan of the  "Ha: <comment>" ... the comments are usually not informative, and `git blame` shows which lines were modified, when, and with the commit message.

models/mpas_atm/model_mod.f90
hkershaw-brown: nope

models/mpas_atm/model_mod.f90
hkershaw-brown: removed

models/mpas_atm/model_mod.f90
hkershaw-brown: I don't know

models/mpas_atm/model_mod.f90
hkershaw-brown: I think the intent here is that the unit is meters.

models/mpas_atm/model_mod.f90
hkershaw-brown: deleted

models/mpas_atm/model_mod.f90
hkershaw-brown: removed this line

models/mpas_atm/model_mod.f90
hkershaw-brown: The comment above this:
````
! the only portable, reliable test for NaNs we know - if some number is neither
! less than nor equal to/greater than 0, it must be a NaN.  all numerical comparisons
! fail if one or more of the operands are NaN.
````

This came in with the regional mpas commit. I'm not sure what portability problems were hit.  Maybe `a /= a`  gets optimized out by some compilers?

File: models/mpas_atm/model_mod.f90 (hunk 9 above)
Comment (nancycollins) [reply models/mpas_atm/model_mod.f90:4940]: actually, this needs to be status 0.  VERTISUNDEF locations often come in with missing_r8 as the vertical.  the original code was correct.  istatus(:) = 0

File: models/mpas_atm/model_mod.f90 (hunk 8 above)
Comment (nancycollins) [reply models/mpas_atm/model_mod.f90:5110]: see my earlier comment about the dangers of printing zout vs just zout(1).  i think zout(1) is sufficient for most purposes but putting something in the string about member 1 is a good idea.

File: models/mpas_atm/model_mod.f90 (hunk 7 above)
Comment (nancycollins) [reply models/mpas_atm/model_mod.f90:5144]: see earlier comment about having to loop if printing all values.  i think labeling it as member 1 is sufficient.

File: models/mpas_atm/model_mod.f90 (hunk 2 above)
Comment (nancycollins) [reply models/mpas_atm/model_mod.f90:1482]: i believe this is a safer test than (a /= a).

File: models/mpas_atm/model_mod.f90 [hunk 11]
```diff
@@ -1167,21 +1185,31 @@ subroutine model_interpolate(state_handle, ens_size, location, obs_type, expecte
    goto 100
 endif
 
+if (debug > 1 .and. do_output()) &
+   print *, 'model_interpolate for obs_kind:',&
+             trim(get_name_for_quantity(obs_kind)),' at ',trim(locstring),' cellid:',cellid
+
+! pass surface variables for rttov - it should be ok as these are diagnostic (except for surface elevation).
+if((obs_kind == QTY_SURFACE_PRESSURE) .or. (obs_kind == QTY_SURFACE_ELEVATION)    .or. &
```
nancycollins, line 1193: now that the rttov code passes in a valid value for the vertical location for VERTISSURFACE interpolations, i believe that all of these except surface elevation could be removed from the "special case" code which avoids the test for the difference between the location and the model elevation at that location.  but in the interests of making small steps, feel free to put this on the "phase 2" code changes if you want.

models/mpas_atm/model_mod.f90
```diff
@@ -1226,15 +1254,15 @@ subroutine model_interpolate(state_handle, ens_size, location, obs_type, expecte
    end select
 endif
 
-if (debug > 10 .and. do_output()) &
+if (debug > 1 .and. do_output()) &
```
nancycollins, line 1257: i'm not sure we want to drop the debug level on this and the next test down to 1.  i think they were ok as 10 and 4.

models/mpas_atm/model_mod.f90
```diff
@@ -1319,40 +1349,57 @@ subroutine model_interpolate(state_handle, ens_size, location, obs_type, expecte
...
+   if (debug > 9 .and. do_output()) then
+      print *, 'model_interpolate: obs_kind,name,istatus,expected_obs,location =', &
+      obs_kind,trim(get_name_for_quantity(obs_kind)),istatus(1),expected_obs(1),trim(locstring)
+   endif
+   if ( all(istatus /= 0 ) ) goto 100
 
 else if (obs_kind == QTY_SPECIFIC_HUMIDITY) then
    tvars(1) = get_progvar_index_from_kind(QTY_VAPOR_MIXING_RATIO)
-   call compute_scalar_with_barycentric(state_handle, ens_size, location, 1, tvars, values, istatus)
+   call compute_scalar_with_barycentric(state_handle, ens_size, location, 1, tvars(1), values(1,:), istatus)
```
nancycollins, line 1378: nice fix.

models/mpas_atm/model_mod.f90
hkershaw-brown: un-dropped the debug level

models/mpas_atm/model_mod.f90
timhoar: not a deal breaker

models/mpas_atm/model_mod.f90
timhoar: There are many instances where it is not labelled as member 1 ... do you want me to change them?

timhoar:
I was puzzling over the skintemp interpolation 
```
       call compute_surface_data_with_barycentric(skintemp(:), location, expected_obs(1), istatus(1))
```
and am concerned that the skintemp is metadata common to all ensemble members. Ditto for seaice.
That doesn't seem right to me. Has anyone talked to Soyoung about this?

models/mpas_atm/model_mod.f90
hkershaw-brown: there is an early return in `convert_vert_distrib_state` if VERTISUNDEF so I am not sure what this check was intended to catch.  Removed the 82 iststaus.

models/mpas_atm/model_mod.f90
hkershaw-brown: I'm just changing them at the moment thanks!

hkershaw-brown: changed to member 1 in debug message

models/mpas_atm/model_mod.f90
hkershaw-brown: putting this into an issue for mpas

hkershaw-brown:
> I was puzzling over the skintemp interpolation
> 
> ```
>        call compute_surface_data_with_barycentric(skintemp(:), location, expected_obs(1), istatus(1))
> ```
> 
> and am concerned that the skintemp is metadata common to all ensemble members. Ditto for seaice.
> That doesn't seem right to me. Has anyone talked to Soyoung about this?

Adding this as a issue

models/mpas_atm/model_mod.f90
hkershaw-brown: issue #96

models/mpas_atm/model_mod.f90
```diff
@@ -5531,11 +5626,17 @@ subroutine get_interp_pressure(state_handle, ens_size, pt_offset, density_offset
 
 ! Convert theta, rho, qv into pressure
 call compute_full_pressure(ens_size, pt(:), density(:), qv(:), pressure(:), tk(:), ier) !HK where is tk used?
-
-if((debug > 11) .and. do_output()) then
+if( all(ier/= 0) ) then
+    if (debug > 1 .and. do_output()) then
+        write(string2,'("Failed in compute_full_pressure")')
+        call error_handler(E_MSG, 'get_interp_pressure',string2,source, revision, revdate)
```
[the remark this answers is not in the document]
hkershaw-brown: done

models/mpas_atm/model_mod.f90
```diff
@@ -7335,33 +7438,55 @@ function theta_to_tk (ens_size, theta, rho, qv, istatus_in)
...
 real(r8), dimension(ens_size) :: theta_m    ! potential temperature modified by qv
 real(r8), dimension(ens_size) :: exner      ! exner function
 real(r8), dimension(ens_size) :: qv_nonzero ! qv >= 0
+integer :: e
 
 qv_nonzero = max(qv,0.0_r8)
 theta_to_tk = missing_r8
 
-where (istatus_in == 0)
+if (any(istatus /= 0)) then
```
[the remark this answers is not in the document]
hkershaw-brown: done

Issue #93, opened 2020-12-10
Update POP initial ensemble scripts for cesm2_1


amrhein:
The setup script

${DARTROOT}/models/POP/shell_scripts/cesm2_1/HPSS_get_ensemble

to stage a POP ensemble targeted Who Kim's JRA-forced POP runs on HPSS that have since been moved.

I plan to write a Python script that copies these from campaign storage, unzipping as necessary, and renames files to give 80 initial ensemble members staged in directory 

'/glade/scratch/'+USER+'/'+save_dir+'/'+'POP_ensemble_dir'

hkershaw-brown:
Pull request https://github.com/NCAR/DART/pull/97 closes this issue.

PR #97, opened 2020-12-14
Update POP ensemble initialization

amrhein:
Added a Python script to copy restart files from Who Kim's JRA-forced POP runs on campaign storage, decompress if necessary, and rename to ensemble members 1-80. Also removed script HPSS_get_ensemble because it is no longer needed.

This fixes Github NCAR/DART Issue #93.

The changes were tested by building a hybrid case using CESM2_1_setup_hybrid and ensuring that the case ran to completion. Results here:
glade/work/damrhein/cases/GNYF.f09_g17_e80/GNYF.f09_g17_e80.run.o5430028


johnsonb-ucar:
The code works as described and is consistent with the [official python style guide](https://www.python.org/dev/peps/pep-0008/).

johnsonb-ucar (approved):

Issue #99, opened 2020-12-14
AIRS AMSUA observation converter ... Zeeman coefficients and channels


timhoar:
The `observations/obs_converter/AIRS/convert_amsu_L1.f90` ultimately requires two pieces of microwave metadata: `mag_field` and `cosbk`. The `amsua_netCDF_support_mod.f90` has a section that is supposed to fill in the microwave metadata for the `mag_field` and `cosbk` variables ... It is only executed if the `obs_def_rttov_nml:use_zeeman = .true.`. I have no idea which channels to support for this option, nor do I know the proper values to supply. 

I am really confused by the contraditions in the 
the RTTOV v12 User Guide Section 8.12 _**Zeeman effect for SSMIS and AMSU-A**_
and the information about the AMSU-A mission from Remote Sensing Systems.

In one case, only Channel 14 is affected, in the other, Channels 4-14 are affected.

http://www.remss.com/missions/amsu/ states that:

> Molecular Oxygen has a complex of relatively strong absorption lines near 60 GHz. ...
> The AMSU instruments are composed of 2 sub units, AMSU-A and AMSU-B.  AMSU-B is a humidity sounder
> (not discussed further), and AMSU-A is a 15-channel temperature sounder similar to MSU.
> Of the 15 channels, 11 (Channels 4 through 14) are located in the 60 GHz absorption complex and thus are
> most closely related to atmospheric temepratures at various heights above the surface.  

Which is true ... Channels 4-14 are 52.8, 53.596, 54.4, 54.94, 55.5, 57.29034, 57.29034, 57.29034, 57.29034, 57.29034, 57.29034 GHz, according to the `center_freq` variable in the data files from the NASA Earthdata Data Access Services website.

The RTTOV v12 user guide states that:

> For microwave sensors that have high peaking weighting functions in the mesosphere such as SSMIS,
> channels close to lines of molecular oxygen may be significantly affected by the redistribution of line intensity
> through Zeeman splitting as described in the RTTOV v10 Science and Validation Report. The absorption for the
> affected channels will depend on the strength and orientation of the magnetic field. You must specify two input
> variables for the geomagnetic field in the rttov_profiles structure, these being the magnitude, Be, of the field
> and the cosine, cosbk, of the angle between the field vector and the viewing path considered. For SSMIS,
> values will be available with the satellite data stream, and will therefore already match the geographical location
> and orientation of the viewing path. For AMSU-A, this is not the case, but the values may be obtained from
> a pre-computed look-up table.
[...]

PR #101, opened 2020-12-18
AIRS/AMSU-A Level 1 radiance observation converter

timhoar:
The datset of interest is: "AIRS/Aqua L1B AMSU (A1/A2) geolocated and calibrated brightness temperatures V005 (AIRABRAD) at GES DISC" The _short name_ for this dataset is 'AIRABRAD'.

The download instructions and much more are in the README.md file in the AIRS directory.

The data are distributed in HDF-EOS(2) format, which is in near-direct contradition to the HDF5 requirements of the RTTOV libraries. The RTTOV libraries are needed for radiance observation metadata, so the conversion to DART format is a two-step procedure. 

There is a curiosity when bulding the L1_AMSUA_to_netcdf.f90 program in that it builds just fine when I run quickbuild from the command line. When the 'test_dart.csh` routine is run, the build for L1_AMSUA_to_netcdf complains that the jpeg library is not found ....

Issue #99 is the conversation about the need for this converter.


hkershaw-brown:
Is this pull request meant for the master branch?

PR #102, opened 2020-12-18
AMSU-A radiance observation converter

timhoar:
The last PR was against Manhattan ... that was a mistake, should be against 'master'.
The datset of interest is: "AIRS/Aqua L1B AMSU (A1/A2) geolocated and calibrated brightness temperatures V005 (AIRABRAD) at GES DISC" The short name for this dataset is 'AIRABRAD'.

The download instructions and much more are in the README.md file in the AIRS directory.

The data are distributed in HDF-EOS(2) format, which is in near-direct contradition to the HDF5 requirements of the RTTOV libraries. The RTTOV libraries are needed for radiance observation metadata, so the conversion to DART format is a two-step procedure.

There is a curiosity when bulding the L1_AMSUA_to_netcdf.f90 program in that it builds just fine when I run quickbuild from the command line. When the 'test_dart.csh` routine is run, the build for L1_AMSUA_to_netcdf complains that the jpeg library is not found ....


assimilation_code/modules/utilities/netcdf_utilities_mod.f90
```diff
@@ -64,7 +62,8 @@ module netcdf_utilities_mod
           nc_close_file,                  &
           nc_begin_define_mode,           &
           nc_end_define_mode,             &
-          nc_synchronize_file
+          nc_synchronize_file,            &
+          NF90_MAX_NAME, NF90_MAX_VAR_DIMS
```
timhoar, line 66: Passing these variables through allows us to declare variables of the right length without having a `use netcdf` or `use typesizes` statement.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
```diff
@@ -84,8 +83,10 @@ module netcdf_utilities_mod
 interface nc_get_global_attribute
    module procedure nc_get_global_char_att
    module procedure nc_get_global_int_att
-   module procedure nc_get_global_real_att
-   module procedure nc_get_global_real_array_att
+   module procedure nc_get_global_float_att
+   module procedure nc_get_global_double_att
+   module procedure nc_get_global_float_array_att
+   module procedure nc_get_global_double_array_att
```
timhoar, line 89: The AMSUA netCDF files that is the CF-compliant version of the HDF-EOS file has global attributes that are 'float' and 'double'. Since the netCDF reader is a clone of the HDF reader, the variables were declared identically to the provided HDF example. 64-bit reals as well as 32-bit reals are needed, so there needed to be more routines in the interface.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
```diff
@@ -128,34 +131,43 @@ module netcdf_utilities_mod
 
 interface nc_put_variable
    module procedure nc_put_char_1d
-   module procedure nc_put_single_int_1d
-   module procedure nc_put_int_1d
-   module procedure nc_put_single_real_1d
-   module procedure nc_put_real_1d
```
timhoar, line 134: the 'nc_put_single_xxxx' procedures are renamed `nc_put_xxx_0d` in keeping with the 1d,2d,3d, notation.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
```diff
@@ -128,34 +131,43 @@ module netcdf_utilities_mod
...
+   module procedure nc_put_real_0d
+   module procedure nc_put_real_1d
+   module procedure nc_put_real_2d
+   module procedure nc_put_real_3d
    module procedure nc_put_real_4d
+   module procedure nc_put_double_0d
+   module procedure nc_put_double_1d
+   module procedure nc_put_double_2d
+   module procedure nc_put_double_3d
+   module procedure nc_put_double_4d
```
timhoar, line 149: needed 32-bit and 64-bit versions.  the question is why these are 'real' and 'double' when the others are 'float' and 'double'

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
```diff
@@ -437,7 +485,7 @@ subroutine nc_add_char_att_to_var(ncid, varname, attname, val, context, filename
 ret = nf90_inq_varid(ncid, varname, varid)
 call nc_check(ret, routine, 'inquire variable id for '//trim(varname), context, filename, ncid)
 
-ret = nf90_put_att(ncid, varid, attname, val)
+ret = nf90_put_att(ncid, varid, attname, adjustl(val))
```
timhoar, line 488: I was surprised by this, but it seemed to be needed.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
```diff
@@ -1168,25 +1263,24 @@ end subroutine nc_put_int_1d
 
 !--------------------------------------------------------------------
 
-subroutine nc_put_single_real_1d(ncid, varname, varindex, varval, context, filename)
+subroutine nc_put_real_0d(ncid, varname, varval, context, filename)
```
timhoar, line 1266: In GitHub, this lines up with the wrong routine. For 0d variables, there is no 'start,count' so there is no need for a 'varindex'.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
```diff
@@ -1407,6 +1625,32 @@ end subroutine nc_put_real_4d
 ! ever want to support these - just punt and make the caller drop down
 ! into direct calls to the netcdf lib.
 
+!--------------------------------------------------------------------
+
+subroutine nc_get_short_0d(ncid, varname, varval, context, filename)
+
+integer,          intent(in)  :: ncid
+character(len=*), intent(in)  :: varname
```
timhoar, line 1633: This is a mistake ... I have to add a routine for i2 as well as i4 ... stay tuned.

build_templates/mkmf.template.rttov.intel
```diff
@@ -129,29 +129,35 @@ LD = ifort
...
-HDFLIB = $(HDFDIR)/lib
-HDFLIBS = -lhdf5hl_fortran -lhdf5_hl -lhdf5_fortran -lhdf5 
+#HDFINC = /glade/u/home/thoar/hdf-eos/include
+#HDFLIB = /glade/u/home/thoar/hdf-eos/lib
+#HDFLIBS = -lhdfeos -lGctp -lmfhdf -ldf -ljpeg -lz
+
+#HDF5DIR = /glade/u/apps/ch/opt/hdf-eos5/5.1.16/intel/19.0.5
+#HDF5INC = $(HDF5DIR)/include
+#HDF5LIB = $(HDF5DIR)/lib
+#HDF5LIBS = -lhdf5hl_fortran -lhdf5_hl -lhdf5_fortran -lhdf5
```
timhoar, line 139: I'm looking for advice on how to handle the HDF dependencies across platforms ...

models/gitm/gitm_to_netcdf.f90
```diff
@@ -202,7 +202,7 @@ program gitm_to_netcdf
 call get_time(delta_time, nsecs, ndays)
 real_time = ndays + (real(nsecs, r8) / 86400.0_r8)
 
-call nc_put_variable(ncid, 'time', 1, real_time)
+call nc_put_variable(ncid, 'time', real_time)
```
timhoar, line 205: This now resolves to the '0d' variant, scalars do not need to supply a length.

models/lorenz_96_2scale/model_mod.f90
```diff
@@ -436,7 +436,8 @@ subroutine nc_write_model_atts(ncid, domain_id)
 integer :: i, xs, xe, ys, ye
 integer(i8) :: indx
 type(location_type) :: lctn
-real(r8) :: loc
+
+real(r8), allocatable :: xloc(:),yloc(:)
```
timhoar, line 440: Rather than make repeated calls to a 1d put_var, I decided to fill the array and make a single call to the 1d put_var.

observations/forward_operators/obs_def_rttov_mod.f90
```diff
@@ -4118,8 +4118,7 @@ function get_rttov_option_logical(field_name) result(p)
    integer,          parameter   :: duc = ichar('A') - ichar('a')
    character(len=:), allocatable :: fname
    
-   character :: ch
-   integer   :: slen, i
+   integer   :: slen
```
timhoar, line 4121: cleanup .. unused vars removed.

observations/obs_converters/AIRS/BUILD_HDF-EOS.sh
```diff
@@ -0,0 +1,270 @@
+#!/bin/sh
+# 
+# updated 4 Dec 2020
+
```
timhoar, line 4: AIRS/BUILDME is now instructions for building HDF-EOS, so I renamed it when I updated the content describing where to get the source, what was needed, etc.

observations/obs_converters/AIRS/L1_AMSUA_to_netcdf.f90
```diff
@@ -0,0 +1,157 @@
+! DART software - Copyright UCAR. This open source software is provided
+! by UCAR, "as is", without charge, subject to all terms of use at
+! http://www.image.ucar.edu/DAReS/DART/DART_download
+
+program L1_AMSUA_to_netcdf
```
timhoar, line 5: This program is not finished. At some point, if someone wanted to take a run at making a converter that reads HDF-EOS and only converts what is needed for the DART observation - here is the place to start.

There is an error_handler() EXIT right at the beginning with a warning message.

observations/obs_converters/AIRS/airs_JPL_mod.f90
```diff
@@ -150,28 +150,23 @@ subroutine airs_ret_rdr(file_name, airs_ret_gran, ver)
...
-      integer :: he5_swclose        ! closing the file
-
-      fid = he5_swopen(file_name, 1)
+      integer :: swopen         ! open a swath file
+      integer :: swattach       ! attatch to a swath object
+      integer :: swinqswath     ! retieves number and names of swaths in file
+      integer :: swrdfld        ! read data from a data field
+      integer :: swrdattr       ! read swath attribute
+      integer :: swdetach       ! detatching from the swath object
+      integer :: swclose        ! closing the file
```
timhoar, line 160: At one point, this module used HDF-EOS  calls and I converted them to use HDF-EOS5 calls when I thought it had to use the RTTOV libraries.  Turns out the converter that uses this module (`convert_airs_L2.f90`) does not use radiances, so I am
reverting back to the aa9e25892ffc  version.

observations/obs_converters/AIRS/airs_obs_mod.f90
```diff
@@ -72,7 +71,8 @@ module airs_obs_mod
 
 contains
 
-!-------------------------------------------------
+!-------------------------------------------------------------------------------
+!>
```
timhoar, line 75: No functional changes in this file ... just code cleanup and removing unused variables.

observations/obs_converters/AIRS/amsua_bt_mod.f90
```diff
@@ -0,0 +1,1100 @@
+
+module amsua_bt_mod
+
+! the contents of this file are a consolidation of:
+!    amsua_bt_typ.inc
+!    amsua_bt_struct.inc
+! modified to use fortran 90, removed redundant type declarations
```
timhoar, line 7: These original files came from NASA and were really unneccessarily complex F77 files.
I have modified these substantially, probably enough to put the DART copyright on them.  Thoughts?

observations/obs_converters/AIRS/work/input.nml
```diff
@@ -6,55 +6,79 @@
...
+#               57.29034, 57.29034, 57.29034, 57.29034, 89 ;
+#
+# All these are identical:
+# channel_list       = 3,15
+# channel_list       = 'A1-1','A1-13'
+# channel_list       = 50.3,89
+
+&convert_amsu_L1_nml
+   l1_files           = '../data/AIRS.2019.06.22.236.L1B.AMSU_Rad.v5.0.0.0.G19174110442.nc',
+                        '../data/AIRS.2019.06.22.237.L1B.AMSU_Rad.v5.0.0.0.G19174110446.nc'
```
timhoar, line 48: These data files exist in `DART_test_cases/test_build_datasets.tar`

observations/obs_converters/AIRS/work/mkmf_L1_AMSUA_to_netcdf
```diff
@@ -0,0 +1,29 @@
...
+else if ( -e /glade/work/thoar/hdf-eos/lib/libhdfeos.a ) then
+   set HDFEOS_INCLUDE = "/glade/work/thoar/hdf-eos/include"
+   set HDFEOS_LIBROOT = "/glade/work/thoar/hdf-eos/lib"
+   set HDFEOS_LIBRARIES = "-lhdfeos -lGctp -lmfhdf -ldf -ljpeg -lz"
+
+else
+   set HDFEOS_INCLUDE = "/Users/thoar/gnu/gnu_9.2.0/hdf-eos/include"
+   set HDFEOS_LIBROOT = "/Users/thoar/gnu/gnu_9.2.0/hdf-eos/lib"
+   set HDFEOS_LIBRARIES = "-lhdfeos -lGctp -lmfhdf -ldf -ljpeg -lz"
+endif
```
timhoar, line 21: Again ... why this works interactively but not when run as a batch job is beyond me.
Note the subtlety in that _there is no hdf-eos module_ on cheyenne, so the first 'if' block is moot.

timhoar (commented):
I have left comments throughout. Hopefully these will explain the thought process.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
nancycollins, line 1266: hi tim - have you tried compiling all the converters with this version of the netcdf utilities?  i believe the "put single 1d" routines were not intended to write scalars, but to overwrite a single value in a 1d array.  thus varindex is needed.  if the overloading doesn't break, perhaps you need to keep both? 

your question about using names with real vs float is a valid one.  i was inconsistent in terminology.  float and double are C conventions, and match how netcdf names them i think; real is fortran but doesn't indicate size.  if you want to rename them all to float/double i think that's transparent to the calling code.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
nancycollins: this is great - i didn't think to do that and it does let you set the lengths using the actual netcdf variables.  one question: would it be less confusing to define our own parameters which are initialized to these lengths and then export those?  otherwise it could be unclear how they get values if the user isn't using the actual netcdf module?

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
nancycollins: in your calling code how were you creating the val string that you passed in here?  if using write() without a format string it might have added an initial blank?  i'm not sure if this is a good idea to muck with what the user gives us here.  (e.g. what if they want 8 initial blanks for some reason?)

observations/obs_converters/AIRS/work/mkmf_L1_AMSUA_to_netcdf
nancycollins: are these setting internal make variables, shell variables (which kind - tcsh or ksh), or environment variables?  what gets preserved might be different depending on interactive vs batch system, maybe?

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
nancycollins: random fluctuations in the quantum continuum?  :)

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
nancycollins: see my other comment for why these are actually different routines, and if the interface overloading still works, that we might need both.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
timhoar: I have run this through the entire test suite, which did check this code.  There is still a nc_put_real_1d() routine, which could be used to stuff an array of length one (technically not a scalar) into some slot in a netCDF array variable. Stuffing a scalar into  a slot in an array is - I think - unsupported as is, but the only place that was happening was in the lorenz_96_2scale model - which now writes an array into an array. None of the observation converters had anything to say about it. I suspect 99% of them are still using the netCDF routines in obs_converters/utilities/obs_utilities_mod.f90

I'll tackle the renaming at a future time, this PR is/was spiraling out of control and I'm already at the limit of the scope I wanted.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
timhoar: I'm still a bit torn. If this thing is identically described in the netCDF Fortran manuals, I'd like to keep it the same. If someone web-searched it, it would be clear. I guess I'd rather not create something exactly the same and just call it something different.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
timhoar: Understood ... but I'm not sure we need a routine to stick a scalar (as opposed to a length-1 array) into a particular slot in a netCDF array variable. We do not do that  in any of the code tested by test_dart.csh

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
timhoar: That wouldve been my guess ... just didn't want to seem ... _**obvious**_ ...

timhoar (commented):
I'm still checking on the `attribute string that starts with a blank` issue.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
nancycollins: ok.  i see the original usage in lorenz_96_2scale.  you changed it to put the whole arrays in a single call?  sounds good.  (i couldn't imagine that i'd invent this variation unless it was used someplace.)

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
nancycollins: agreed - now that you pointed out that it was only used in a single place (which has been changed to put the entire array) i'm happy to remove that variant.  (and sorry - i should have opened a review on this instead of nickel and dime-ing my comments.

observations/obs_converters/README.md
```diff
@@ -281,7 +281,7 @@ observations into the format required by DART.
 
 The current list of converters (some directories contain multiple converters) include:
 
--   [AIRS](AIRS/AIRS.html)
+-   [AIRS](AIRS/README.md) [atmospheric variables](AIRS/AIRS.html) and [AMSUA radiances](AIRS.README.md)
```
hkershaw-brown, line 284: Typo in link here: AIRS.README.md  -> AIRS/README.md

hkershaw-brown: I corrected this typo.

models/lorenz_96_2scale/model_mod.f90
hkershaw-brown: this is bitwise with Manhattan & does not change the runtime (unscientific test, it is faster 25.4s vs 25.9s).

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
hkershaw-brown: `use netcdf_utilities_mod,  only :` I think covers where NF90_MAX_NAME comes from.

timhoar: Helen, are you agreeing that we should keep the name the same? Or are you raising a different issue?

observations/obs_converters/README.md
```diff
@@ -292,6 +292,7 @@ The current list of converters (some directories contain multiple converters) in
 -   [DWL](DWL/dwl_to_obs.html)
 -   [GMI](GMI/README.md)
 -   [GOES](GOES/README.md)
+-   [GMI](GMI/README.md)
```
timhoar, line 295: This is in here twice ... and is not in alphabetical order ...

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
hkershaw-brown: Agreeing that we keep the same name.

models/gitm/model_mod.f90
```diff
@@ -2101,19 +2101,17 @@ subroutine unpack_data0d(data0d, ivar, ncid, define)
       call error_handler(E_MSG,routine,string1,source,revision,revdate)
    end if
 
-   call nc_define_double_variable(ncid, gitmvar(ivar)%varname, (/ LON_DIM_NAME, LAT_DIM_NAME, ALT_DIM_NAME /) )
```
hkershaw-brown, line 2104: I'm missing why this line is being removed.  Is this related to the AIRS converter?

observations/obs_converters/AIRS/README.md
```diff
@@ -0,0 +1,245 @@
...
+[HDF-4](http://www.hdfgroup.org) format, using
+some additional conventions for metadata called
+[HDF-EOS](http://hdfeos.org/software.php).
+There is a basic library for accessing data in hdf files, and a variety of
+[generic tools](http://www.hdfgroup.org/products/index.html)
+that work with hdf files.
+The specific library we use is the
+[HDF-EOS2](http://hdfeos.org/software/library.php#HDF-EOS2)
+The web page has a link to specific build
+instructions.  Also, see [Build](#build") on this page
```
hkershaw-brown, line 60: broken link here [Build](#build")   I think should be [Build](#build)

hkershaw-brown: fixed this since it was a simple type

models/gitm/gitm_to_netcdf.f90
hkershaw-brown: nice

observations/obs_converters/AIRS/L1_AMSUA_to_netcdf.f90
hkershaw-brown: @timhoar I've put in an issue (https://github.com/NCAR/DART/issues/105) to finish the L1_AMSUA_to_netcdf program.  I probably missing the subtleties of the various AMSUA issues.

observations/obs_converters/AIRS/README.md
```diff
@@ -0,0 +1,245 @@
+
```
hkershaw-brown, line 1: nice detailed build instructions.

models/gitm/model_mod.f90
timhoar: The declaration of the scalar variable as a 3D variable is simple misleading. The thing is read as a scalar from a namelist, so it will never be 3D ...

build_templates/mkmf.template.rttov.intel
```diff
@@ -129,29 +129,35 @@ LD = ifort
...
+RTLIBDIRS = -L$(RTLIB) 
 
-INCS = -I$(NETCDF)/include $(RTINCS)
-LIBS = -L$(NETCDF)/lib -lnetcdff -lnetcdf $(RTLIBDIRS) $(RTLIBS) $(HDFLIBS)
+INCS = -I$(NETCDF)/include $(RTINCS) 
+LIBS = -L$(NETCDF)/lib -lnetcdff -lnetcdf \
+       -L$(RTLIB)   $(RTLIBS) 
 
-FFLAGS  = -O -assume buffered_io $(INCS)
+# FFLAGS  = -O0 -assume buffered_io $(INCS)
```
hkershaw-brown, line 153: This should be -0 (-02)

build_templates/mkmf.template.rttov.intel
```diff
@@ -129,29 +129,35 @@ LD = ifort
...
 
 # for development or debugging, use this instead:
-# FFLAGS = -g -C -check noarg_temp_created -fpe0 \
-#          -fp-model precise  -ftrapuv -traceback \
-#          -warn declarations,uncalled,unused $(INCS)
-#
+ FFLAGS = -g -C -check noarg_temp_created -fpe0 \
+          -fp-model precise  -ftrapuv -traceback \
+          -warn declarations,uncalled,unused $(INCS)
+
```
hkershaw-brown, line 160: should have the production options, rather that the debugging options commented in.

hkershaw-brown (changes requested):
Changes requested: 
mkmf.template.rttov.intel production options rather than debugging options.

build_templates/mkmf.template.rttov.intel
timhoar: Yup ... I'm on it.

build_templates/mkmf.template.rttov.intel
timhoar: agreed

build_templates/mkmf.template.rttov.gfortran
```diff
@@ -68,35 +68,29 @@ LD = gfortran
...
-# You will likely not need to modify below this line
+HDFEOS5 = /Users/thoar/gnu/gnu-9.2.0/
 
-HDFINC = $(HDFDIR)/include
-HDFLIB = $(HDFDIR)/lib
-HDFLIBS = -lhdf5hl_fortran -lhdf5_hl -lhdf5_fortran -lhdf5
-
-RTMOD = $(RTTOV)/mod
-RTINC = $(RTTOV)/include
-RTLIB = $(RTTOV)/lib
```
timhoar, line 89: I thought this was a bit overly complicated and made it hard for the Intel environment on Cheyenne to look similar.

build_templates/mkmf.template.rttov.intel
```diff
@@ -117,41 +117,37 @@ LD = ifort
 # for more information on installing RTTOV.
 
 # MODIFY THE FOLLOWING VARIABLES FOR YOUR SYSTEM:
-# If your NETCDF, HDFDIR, and RTTOV environment variables are not set,
+# If your NETCDF, HDFEOS5, or RTTOV environment variables are not set,
```
timhoar, line 120: This whole section is now very similar to the gfortran version and works on cheyenne with or without the hdf-eos module, as long as the HDFEOS5 variable is set correctly (i.e. consistent with that in the hdf-eos module)

observations/obs_converters/AIRS/work/mkmf_L1_AMSUA_to_netcdf
timhoar: So ... what is happenning is that the build works fine - but the LD_LIBRARY_PATH is not necessarily set consistent with these. I made notes in the affected mkmf files (see lines 7-10 here) and in the `run_tests.csh`  script.

observations/obs_converters/quikscat/work/mkmf_convert_L2b
```diff
@@ -4,31 +4,30 @@
 # by UCAR, "as is", without charge, subject to all terms of use at
 # http://www.image.ucar.edu/DAReS/DART/DART_download
 #
-# DART $Id$
+# This converter needs the HDF-EOS2 libraries which are built on the HDF4
+# libraries. If any of these libraries are not found during execution,
+# make sure your [DY]LD_LIBRARY_PATH contains $HDFEOS_LIBROOT
```
timhoar, line 9: This is now consistent with the other programs that use hdf ...

timhoar (commented):
Just one last review and it is done.

hkershaw-brown (approved):
looks good.  Let's merge!

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
timhoar: I removed the adjustl() and nothing broke ... I think the code I was using that exercised that is no longer relevant.

PR #116, opened 2021-01-12
Adding Enhanced Adaptive Inflation in DART_LAB + Improving RTPS functionality

mgharamti:
1- Added the recent enhanced adaptive inflation functionality in the DART_LAB GUIs:

Previously, the adaptive inflation algorithm was based on Anderson 2009 work that assumes a Gaussian distribution for the inflation. The new algorithm is based on the work of El Gharamti 2018 that relaxes this assumption and imposes an inverse Gamma distribution on the inflation field. This is especially useful when performing both inflation and deflation (or in general 'variance correction'). The inverse Gamma (I-Gamma) behaves better than Gaussian near zero. Note that deflation is not recommended in Anderson 2009. Another advantage of the new scheme is that during the update, the standard deviation of the inflation may be increase and decrease (unlike the Gaussian case which only force it to decrease). 

The user now will have the option to choose between both scheme through radio buttons in the Lorenz96 (inf) GUI. The bets way to asses the difference between the 2 schemes is to look at the RMSE and Spread. In the fortran world of DART, the Gaussian variant is termed as 'flavor 2' and the enhanced I-Gamma one is termed as 'flavor 5'.  

Supporting the new inflation functionality required adding 2 files: [1] change_GA_IG.m and [2] enh_compute_new_density.m
There are also a lot of changes in 'run_lorenz_96_inf.m' and 'update_inflate.m' I've also added some bullet-proofing if statements within 'update_inflate.m' to match the fortran modules. To keep up with the changes to 'update_inflate.m', I had to change the function call in 'oned_model_inf.m'   

Future recommendations: 
- Add some sort of a plot to show the pdf of the inflation (in addition to the time-series of the inflation value)
- It's probably time to get rid of 'run_lorenz_96.m' I just don't see any reason to keep it in the repo. It's obsolete now!

 - - - -

2- Improved the support for RTPS (inf_flavor 4) posterior inflation: 

Previously, one could only decide on a single parameter (say, alpha) in the input nml (as inf_initial(2)) without being able to see what is the effective posterior inflation the RTPS scheme is actually using. In other words, if you check the resulting inflation in preassim.nc for instance (variable state_postinf_mean) it would be the same as inf_initial(2). That's incorrect. 

To clarify, there is nothing wrong in the RTPS implementation, however, reporting the output has an issue. The following update corrects the issue and updates the output files with the right posterior inflation values. See the figure below for illustration. The left hand side plot is the value of state_postinf_mean in preassim.nc (or analysis.nc) after running filter. The right hand side is the plot after the proposed code change. Note that both runs use inf_flavor(2) = 4 and inf_initial(2) = 0.3

Screen Shot 2021-01-13 at 10 28 30 AM

Keep in mind that RTPS inflation does not use any files (i.e., inflation has no-memory between DA cycles).


hkershaw-brown:
I've only looked at the Fortran changes at the moment. 

For the RTTPS factor there are two comments that seem to contradict each other. In filter_mod.f90

````
1606            !Reset the RTPS factor to the given input.nml value
1607           ens_handle%copies(inflate_copy, j) = inf_initial(2)
````
so in the example you have plotted, inf_initial=0.3

Then in adaptive_inflate_mod.f90, there is a comment and code setting `inflate_handle%inflate = 1.0_r8`

````
242   !Overwriting the initial value of inflation with 1.0
243   !as in other inflation flavors (usually start from 1). 
244   !This is required for RTPS because inf_initial(2)
245   !is not really the inflation factor but rather the weighting
246   !parameter, say alpha: 
247   !RTPS: lambda = alpha * (sd_b - sd_a) / sd_a + 1
248   !where; sd_b (sd_a): prior (posteriro) spread
249   if(inf_flavor == 4) inflate_handle%inflate = 1.0_r8
````

So my question is should the RTPS start at 1, or start at whatever the namelist input for inf_initial(2) is. 


Second comment:
`ens_handle%copies(inflate_copy, j) = inf_initial(2)` looks like it can be moved outside the `j = 1, ens_handle%my_num_vars` loop because there is no dependence on `j`:

````
!Reset the RTPS factor to the given input.nml value
ens_handle%copies(inflate_copy, :) = inf_initial(2)

         do j = 1, ens_handle%my_num_vars
`````

Third comment:
I am not sure whether RTPS and the "inflate each group separately" functionally are ever intended to be used at the same time.  If they are, then this code will only track the RTPS factor for the last group.

File: docs/DART_LAB/matlab/oned_model_inf.m [hunk 1]
```diff
@@ -1386,10 +1386,15 @@ function step_ahead(~, ~)
...
+                    % switch between 'Gaussian' [Anderson 2009] and 'Gamma'
+                    % [El Gharamti 2018]. The GUI option is only available
+                    % in the Lorenz'96 section. 
                     [lambda, handles.adap_inf_Std] = ...
                         update_inflate(mean(ens), var(ens), observation, obs_error_sd^2, inf_prior, ...
-                        handles.inflation, handles.adap_inf_Min, handles.adap_inf_Max, ...
-                        1, handles.adap_inf_Std, handles.adap_inf_Std_Min);
+                        handles.inflation, handles.adap_inf_Std, handles.adap_inf_Min, handles.adap_inf_Max, ...
+                        1, handles.adap_inf_Std_Min, handles.ens_size, 'Gaussian');
+
```
timhoar, line 1397: Unfortunately, the matching Fortran routine is named 'update_inflation' not 'update_inflate' ... should we (being me) rename `private/update_inflate.m` to   `private/update_inflation.m` to match the Fortan? I think that's in the spirit of this pull request.

timhoar: Actually, this is not a one-to-one comparison. Forget it.

docs/DART_LAB/matlab/private/change_GA_IG.m
```diff
@@ -0,0 +1,42 @@
...
+% obtained as a "real" solution to a cubic polynomial.
+AA = mode_p(4) * sqrt((var_p(2) + 47*var*mode_p(2) + 3*mode_p(4)) / var_p(3));
+BB = 75*var_p(2)*mode_p(5);
+CC = 21*var*mode_p(7);
+DD = var_p(3)*mode_p(3);
+EE = (CC + BB + DD + mode_p(9) + 6*sqrt(3)*AA*var_p(3)) / var_p(3);
+
+beta = (7*var*mode + mode_p(3))/(3*var)                               + ...
+       EE^(1/3)/3 + mode_p(2)*(var_p(2) + 14*var*mode_p(2) + ...
+       mode_p(4)) / (3*var_p(2)*EE^(1/3));
```
timhoar, line 37: While it doesn't make a different to Matlab's numerics ... the Fortran version of this routine uses real number declarations ... (1.0/3.0)/3.0  ... so I am changing them in the Matlab code.

docs/DART_LAB/matlab/private/update_inflate.m
```diff
@@ -23,26 +29,62 @@
...
+    fac2 = -1 / ens_size;
+    
+    if fac1 < abs(fac2), fac2 = 0; end
+    
+    theta_bar_2 = (fac1+fac2) * sigma_p_2 + sigma_o_2;
+end
+
 theta_bar    = sqrt(theta_bar_2);
 u_bar        = 1 / (sqrt(2 * pi) * theta_bar);
 like_exp_bar = - 0.5 * dist_2 / theta_bar_2;
```
timhoar, line 46: This line does not match what I think it should in adaptive_inflate_mod:linear_bayes()
```
! Compute exponent of likelihood at lambda_bar
like_exp_bar = dist_2 / (-2.0_r8 * theta_bar_2)
```

docs/DART_LAB/matlab/private/update_inflate.m
```diff
@@ -1,10 +1,16 @@
...
+% Based on the algorithm in the following articles:
+%
+% [falvor: 2, Gaussian inflation pdf] 
 % Anderson, J. L., 2009: Spatially and temporally varying adaptive covariance 
 % inflation for ensemble filters. Tellus A, 61, 72-83. doi: 10.1111/j.1600-0870.2008.00361.x 
 %
+% [flavor: 5, I-Gamma inflation pdf] 
+% El Gharamti, M., 2018: Enhanced Adaptive Inflation Algorithm for Ensemble Filters. 
+% Monthly Weather Review, 146, 623-640. doi: 10.1175/MWR-D-17-0187.1
+%
```
timhoar, line 13: So ... the Matlab private/update_inflate.m  actually does the job of two Fortran routines (adaptive_inflate_mod.f90  linear_bayes() and enh_linear_bayes().

Granted, update_inflate.m  has fewer lines than the two Fortran routines, but I think the point is that we are trying to make it easy for the reader to go between the Matlab and Fortran worlds.   I'd like to split update_inflate.m into a
simple routine that calls one of two functions (to match the Fortran) ... those routines would be Matlab 'helper' functions in update_inflate.m
Thoughts?

timhoar (changes requested):
I think the update_inflate.m  code does not parallel the Fortran version enough.
I vote to make two helper functions to ease the transition between Fortran and Matlab.
The two routines should parallel the Fortran linear_bayes() and enh_linear_bayes()

mgharamti:
@hkershaw-brown 

- Regarding your first comment: 
To explain what's going on, I need to give some background. In order to use RTPS inflation, we need a single parameter from the user. Let's call this parameter "alpha". Other flavors (for instance 2 and 5) require a lot more parameters (damping, lower and upper bounds, ...). It's important to keep in mind that this alpha parameter is not the inflation factor that RTPS is effectively using to inflate the posterior covariance! In fact, the actual value of the inflation is the one I report on line 247 of adaptive_inflate_init as `lambda = alpha * (sd_b - sd_a) / sd_a + 1`
However, the person who coded this initially perhaps didn't want to add an additional entry in the input namelist to request this parameter. He/She decided to provide this alpha parameter through `inf_initial(2)`. That's fine. The issue come when we write out the value of `state_postinf_mean`. Previously, alpha was written out as the `state_postinf_mean` and that's incorrect (in fact in this case 0.3 is intuitively not an inflation value because it's smaller than 1).  
I only set it to 1 in the `adaptive_inflate_init` routine so that the `state_postinf_mean` initially (at the first time) is equal to 1 and not alpha i.e., no inflation. If I don't do that, that at time 1 `stat_postinf_mean` would be alpha and correct for the rest of the cycles. To answer your question, the RTPS inflation (and any other inflation as a matter of fact) should start from 1. Again, `inf_initial(2)` here in flavor 4 is not the actual inflation. 

- Second comment:
You're right I can reset the `ens_handle%copies(inflate_copy, j) = inf_initial(2)` outside the do loop. I'll do that and push it.  

- Third comment: 
I don't think anyone would combine splitting the ensemble into groups with RTPS. This is a design issue for whoever coded this initially though I'm not concerned.

docs/DART_LAB/matlab/private/change_GA_IG.m
mgharamti: haha, FINE :)

docs/DART_LAB/matlab/private/update_inflate.m
mgharamti: I'm not sure I follow here. Isn't `- 0.5 * dist_2 / theta_bar_2` equivalent to `like_exp_bar = dist_2 / (-2.0_r8 * theta_bar_2)` ??

docs/DART_LAB/matlab/private/update_inflate.m
mgharamti: Good observation. My objective was to bring the functionality of flavor 5 just like what I did with flavor 2 around 3 years ago. See even when I brought in flavor 2 into DART_LAB, I didn't explicitly called the function linear_bayes() or adaptive_inflate_mode .. My intention was to make sure we are replicating the functionality but not mirroring the code. I think having the same parameter/variable names in enough for users to relate the fortran modules to the matlab functions.

mgharamti:
> @hkershaw-brown
> 
> * Regarding your first comment:
>   To explain what's going on, I need to give some background. In order to use RTPS inflation, we need a single parameter from the user. Let's call this parameter "alpha". Other flavors (for instance 2 and 5) require a lot more parameters (damping, lower and upper bounds, ...). It's important to keep in mind that this alpha parameter is not the inflation factor that RTPS is effectively using to inflate the posterior covariance! In fact, the actual value of the inflation is the one I report on line 247 of adaptive_inflate_init as `lambda = alpha * (sd_b - sd_a) / sd_a + 1`
>   However, the person who coded this initially perhaps didn't want to add an additional entry in the input namelist to request this parameter. He/She decided to provide this alpha parameter through `inf_initial(2)`. That's fine. The issue come when we write out the value of `state_postinf_mean`. Previously, alpha was written out as the `state_postinf_mean` and that's incorrect (in fact in this case 0.3 is intuitively not an inflation value because it's smaller than 1).
>   I only set it to 1 in the `adaptive_inflate_init` routine so that the `state_postinf_mean` initially (at the first time) is equal to 1 and not alpha i.e., no inflation. If I don't do that, that at time 1 `stat_postinf_mean` would be alpha and correct for the rest of the cycles. To answer your question, the RTPS inflation (and any other inflation as a matter of fact) should start from 1. Again, `inf_initial(2)` here in flavor 4 is not the actual inflation.

To remove the confusion, I think it might be a good idea supply alpha through `inf_damping(2)` rather than `inf_initial(2)`. I think this would clarify things a bit (because we use `inf_damping` to compute inflation value)

Thoughts?

> * Second comment:
>   You're right I can reset the `ens_handle%copies(inflate_copy, j) = inf_initial(2)` outside the do loop. I'll do that and push it.
> * Third comment:
>   I don't think anyone would combine splitting the ensemble into groups with RTPS. This is a design issue for whoever coded this initially though I'm not concerned.

Issue #79, opened 2020-11-19
gendocs script in DART/docs/doxgen does not work as is


hkershaw-brown:
This is a small fix (changing CHANGELOG to CHANGELOG.md)
However the output produced is uninspiring and I would argue counter-productive since it is not complete.  
Propose removing the `DARTdocs/doxygen` directory.

side note: I am now able to fry eggs on my laptop after running dot on every DART program.

PR #120, opened 2021-01-12
Remove doxygen directory

hkershaw-brown:
Fixes: https://github.com/NCAR/DART/issues/79

No longer using doxygen for API documentation.


timhoar (approved):
Made changes to two files that referenced doxygen.

PR #143, opened 2021-02-02
Referencing correct location of input.nml.template

timhoar:
Thanks to Ilya Livshits for reporting the error.


hkershaw-brown (approved):
looks good.

g-romine (commented):
Agree this looks correct.

PR #150, opened 2021-02-05
clean up namelists 

timhoar:
The namelists have been checked twice ... once during development and again with test_dart.csh
we have to modify the (few) namelists in `test_build_datasets.tar` to reflect the new layout.
My test_dart.csh execution used those (few) old namelists.  I am building an alternate tar file with namelists
appropriate for the new preprocess strategy.


timhoar:
The (seven) namelists that need to be updated are:

1. models/clm/work/input.nml.testing
2. models/gitm/work/input.nml.testing
3. models/wrf/work/input.nml.testing
4. observations/obs_converters/GMI/work/input.nml.testing
5. observations/obs_converters/GOES/work/input.nml.testing
6. observations/obs_converters/GSI2DART/work/input.nml.testing
7. observations/obs_converters/MADIS/work/input.nml.testing

timhoar:
The test tarfile appropriate for the new preprocess strategy is

/glade/p/cisl/dares/DART_test_cases/test_build_datasets_preprocess.tar

(the old one remains unchanged)
The test_dart.csh is in the machine.

timhoar:
If you are OK with these changes, I can either push them directly to the public preprocess_back_comp branch or we can merge them this way. I believe we agreed on one of these methods.

timhoar:
After the update to Catalina, I had to (basically) reinstall my MacPorts compilers and mpich. This now means I am using `GNU Fortran (MacPorts gcc10 10.2.0_5) 10.2.0` which seems to be VERY picky and will not compile things like the NCEP BUFR libs, some of the HDF-EOS libs ... 
```
Error: Rank mismatch between actual argument at (1) and actual argument at (2) (rank-1 and scalar)
ufbtab.f:436:36:
```
and
```
Error: Type mismatch between actual argument at (1) and actual argument at (2) (INTEGER(4)/CHARACTER(*)).
../../../../observations/obs_converters/AIRS/amsua_bt_mod.f90:347:41:
```

Anyone explore the compile options for the gfortran compiler?

hkershaw-brown (approved):

assimilation_code/modules/observations/oned_quantities_mod.f90
```diff
@@ -14,13 +14,8 @@
  
 ! BEGIN DART PREPROCESS QUANTITY DEFINITIONS
 !
-!  ! QTY_STATE_VARIABLE is also predefined in the preprocess program.
-!  ! it is being repeated here because for some classes of models
-!  ! this is the only quantity that they will use.  it must be the
-!  ! the first entry in this file and must not be altered.
+!  ! QTY_STATE_VARIABLE is predefined in the preprocess program.
```
nancycollins: for all the other quantity files, removing QTY_STATE_VARIABLE seems right.

however, in 1D models this is probably the only QTY a model_mod is going to use.  i'm ok with removing it as long as the comment is a bit more prominent when someone looks at this file.

i wanted to try that suggest thing but i can't see where helen clicked before.  

maybe something like:

! ! QTY_STATE_VARIABLE is a generic quantity which is predefined
! ! by the system and available for code to use without needing
! ! a separate definition here..  It is frequently used in models 
! ! where the items in the state vector do not directly represent 
! ! a physical quantity in the model.    
! !
! ! These additional QTYs are available for more specialized cases.
! !

nancycollins (commented):
i added one comment for the oned quantity file, but otherwise the cleanup looks very nice.

hkershaw-brown:
-fallow-argument-mismatch ?

PR #154, opened 2021-02-08
support integers or strings to specify inflation flavor

timhoar:
The `filter_nml:inf_flavor` variable has always been an integer ... making it harder than need be to specify the inflation scheme.
This PR allows for the use of character strings as well as integers to specify `inf_flavor`, and uses integer parameters in the code rather than hardcoded "magic" integers. In the adaptive_inflate_mod.f90, I also made an effort to use integer parameters 'PRIOR' and 'POSTERIOR' rather than '1' or '2'.

```
 inf_flavor                  = 2,                       0
```
and
```
 inf_flavor                  = 'VARYING_SS_INFLATION',                       0
```
are identical, for example.

I have not modified any of the input.nml's to use the character strings ... should I?

There is a comment from PR 116 about contradictory comments in adaptive_inflate_mod.f90 line 249 (old) that I have ignored, but can be and should be addressed in the PR to replace PR # 116.

```
integer, parameter :: NO_INFLATION               = 0
integer, parameter :: OBS_INFLATION              = 1
integer, parameter :: VARYING_SS_INFLATION       = 2
integer, parameter :: SINGLE_SS_INFLATION        = 3
integer, parameter :: RELAXATION_TO_PRIOR_SPREAD = 4
integer, parameter :: ENHANCED_SS_INFLATION      = 5
```

If people like these strings, I can update the documentation as appropriate.

This has been tested with the Intel and gfortran compilers.


assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
```diff
@@ -423,27 +468,27 @@ subroutine validate_inflate_options(inf_flavor, inf_damping, inf_initial_from_re
...
                         'RTPS inflation (type 4) disables posterior inf_damping')
-   inf_damping(2) = 1.0_r8  ! no damping
+   inf_damping(POSTERIOR) = 1.0_r8  ! no damping
 endif
 
 ! enhanced inflation checks - this is before we set the subflavor in the structure.
-if (inf_flavor(1) == 5 .or. inf_flavor(2) == 5) then
+if (inf_flavor(PRIOR) == ENHANCED_SS_INFLATION .or. inf_flavor(POSTERIOR) == ENHANCED_SS_INFLATION) then
    ! check inf_sd_max_change() for valid range
    do i=1, 2
```
mgharamti, line 493: A suggestion: to be consistent with the notation, the do-loop could be re-written
> do i = PRIOR, POSTERIOR

assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -2719,6 +2730,69 @@ subroutine set_copies(cnum, STAGE_COPIES)
...
+!> or the integer associated with the type of inflation. The string names of the
+!> inflation algorithms is based on what is declared in the adaptive_inflate_mod.f90
+!> which is repeated here for reference. 
+!>
+!> NO_INFLATION               = 0
+!> OBS_INFLATION              = 1    observation-space inflation (deprecated)
+!> VARYING_SS_INFLATION       = 2    spatially-varying state-space inflation
+!> SINGLE_SS_INFLATION        = 3    spatially-constant state-space inflation
+!> RELAXATION_TO_PRIOR_SPREAD = 4    (available only with posterior inflation)
+!> ENHANCED_SS_INFLATION      = 5    Inverse Gamma version of VARYING_SS_INFLATION
```
nancycollins, line 2746: tim, can we talk about whether it's feasible to move some of this routine into the utilities module so we don't end up replicating this code (upper case, select, etc) in every module we want to make this change?  maybe it's impractical but it's worth a conversation, i think.

assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -394,29 +400,31 @@ subroutine filter_main()
...
-call adaptive_inflate_init(post_inflate, inf_flavor(2), inf_initial_from_restart(2),  &
+call adaptive_inflate_init(post_inflate, inflation_flavor(2), inf_initial_from_restart(2),  &
    inf_sd_initial_from_restart(2), output_inflation, inf_deterministic(2),            &
    inf_initial(2),  inf_sd_initial(2), inf_lower_bound(2), inf_upper_bound(2),        &
    inf_sd_lower_bound(2), inf_sd_max_change(2), state_ens_handle,                     &
    allow_missing, 'Posterior')
 
 if (do_output()) then
-   if (inf_flavor(1) > 0 .and. inf_damping(1) < 1.0_r8) then
+   if (inflation_flavor(1) > NO_INFLATION .and. inf_damping(1) < 1.0_r8) then
```
mgharamti, line 423: Why not also set 1 to PRIOR and 2 to POSTERIOR here?

nancycollins (commented):
the changes to the adaptive_inflate_mod look great - the parameters make it much easier to read.

i do have a question about parsing strings in the filter namelist.  if you input:

 inf_flavor = 0, 0

and inf_flavor changes from int to string, does it read them as "0" and "0" without adding quotes?  so it understands comma is a delimiter for a string?

and i left a comment about trying to make at least part of this code into a single routine in the utilities mod so if we end up having to modify it we could do it in a single place instead of cutting and pasting similar code into all the modules where we're hoping to make this same change.

assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -2719,6 +2730,69 @@ subroutine set_copies(cnum, STAGE_COPIES)
...
+integer :: i
+character(len=32) :: string
+
+do i = 1,2   ! one is prior, two is posterior
+
+   string = adjustl(flavor_string(i))
+   call to_upper(string)
+
+   select case(trim(string))
+      case ('0','NO_INFLATION')
```
mgharamti, line 2762: Should we add more options? For instance: 
- case('0', 'NO_INFLATION', 'no inf') 
-case('1', 'OBS_INFLATION', 'obs inf') 
- case('2', 'VARYING_SS_INFLATION', 'A09') 
- case('5', 'ENHANCED_SS_INFLATION', 'E18')?
I understand this might complicate the documentation a bit. I just like to write a short version of the word "inflation" as "inf". A09 and E18 refer to Anderson 2009 and El Gharamti 2018, respectively.

mgharamti (commented):
This is a great enhancement to our code. Great job!
I like the names you chose.

assimilation_code/modules/assimilation/filter_mod.f90
timhoar: I ran some tests as to what is valid input for a variable defined to be a character(len=?) ... which surprised me a little.  I always knew that if you wanted to include a slash '/' in a character variable, you had to protect the entire string with quotes to prevent the namelist parser from thinking the slash was the end of the namelist. Seems gfortran requires alphabetic characters to be quoted, while SINGLE numeric characters do not need to be quoted.  Examples of syntaxes that work:

```
   inf_flavor                  = 0,          'RTPS'
   inf_flavor                  = '0',          'RTPS'
   inf_flavor                  = 'NO_INFLATION',          'RTPS'
   inf_flavor                  = 'NO_INFLATION',          0
```
and some syntaxes that fail:
```
   inf_flavor                  = 0,          RTPS,
   inf_flavor                  = 00,          0,
   inf_flavor                  = 'NO_INFLATION',          'RTPS'
```
Not sure if that is definitive enough to answer the question about whether the comma is a sufficient delimeter ...

assimilation_code/modules/assimilation/filter_mod.f90
timhoar: That's just a test to see if you were actually looking  :)    ... clearly missed the one in adaptive_inflate_mod.f90 and didn't know how far to go in the filter_mod.f90 ... there are MANY places  where PRIOR and POSTERIOR could be used.
I had a bit of a dilemma using PRIOR and POSTERIOR for the loop bounds, that only works if PRIOR and POSTERIOR are adjacent numbers ... and actually ONLY if PRIOR and POSTERIOR and 1 and 2, respectively.  If, for some reason, someone wanted to change PRIOR, POSTERIOR to 3 and 7 (or something) the whole system fails.  I've put in a comment to mention that the values of PRIOR and POSTERIOR MUST be 1 and 2 - which clears the way for using PRIOR and POSTERIOR for loop bounds.

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
timhoar: Done.

assimilation_code/modules/assimilation/filter_mod.f90
timhoar: Done.

assimilation_code/modules/assimilation/filter_mod.f90
timhoar: This is discussion for the group ... at the standup today.

assimilation_code/modules/assimilation/filter_mod.f90
timhoar: I created   utilities_mod:flex_parser() that does what I understood from my talk with Nancy. Hopefully, this is along the lines of what she meant!

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
```diff
@@ -400,21 +447,23 @@ subroutine validate_inflate_options(inf_flavor, inf_damping, inf_initial_from_re
 end do
 
 ! Check to see if state space inflation is turned on
-if (inf_flavor(1) > 1) do_prior_inflate     = .true.
-if (inf_flavor(2) > 1) do_posterior_inflate = .true.
+if (inf_flavor(PRIOR)     > OBS_INFLATION)  do_prior_inflate     = .true.
+if (inf_flavor(POSTERIOR) > OBS_INFLATION)  do_posterior_inflate = .true.
```
nancycollins, line 451: this code isn't safe - it assumes OBS_INFLATION has value 1 and would break if changed.  to be more clear, maybe define a separate parameter, value = 2 , which is named 'STATE_SPACE_INFLATION_START' or something and make this code >=

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
```diff
@@ -400,21 +447,23 @@ subroutine validate_inflate_options(inf_flavor, inf_damping, inf_initial_from_re
...
 
 ! Relaxation-to-prior-spread (RTPS) is only an option for posterior inflation
-if(inf_flavor(1) == 4) call error_handler(E_ERR, 'validate_inflate_options', &
+if(inf_flavor(PRIOR) == RELAXATION_TO_PRIOR_SPREAD) &
+   call error_handler(E_ERR, 'validate_inflate_options', &
    'RTPS inflation (type 4) only supported for Posterior inflation', source, revision, revdate)
 
 ! Cannot select posterior options if not computing posterior
-if(.not. compute_posterior .and. inf_flavor(2) > 0) then
+if(.not. compute_posterior .and. inf_flavor(POSTERIOR) > NO_INFLATION) then
```
nancycollins, line 466: this one is clear - we can assume NO_INFLATION is 0.

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
```diff
@@ -1200,23 +1262,27 @@ subroutine set_from_string(from_restart, single_file, from_string)
...
+!>
 
 function nvalues_to_log(inflation_handle, from_string)
 type(adaptive_inflate_type), intent(in) :: inflation_handle
 character(len=*),            intent(in) :: from_string
 integer :: nvalues_to_log
 
-if ((inflation_handle%inflation_flavor == 3) .or. &
-    (inflation_handle%inflation_flavor == 2 .and. from_string == 'from namelist')) then
+if ((inflation_handle%inflation_flavor == SINGLE_SS_INFLATION) .or. &
```
nancycollins, line 1274: by the way, changing the magic numbers in this code to parameters makes it all much clearer.  good change.

assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -2719,6 +2730,50 @@ subroutine set_copies(cnum, STAGE_COPIES)
...
+                                            'RELAXATION_TO_PRIOR_SPREAD',&
+                                            'RTPS                      ',&
+                                            'ENHANCED_SS_INFLATION     ' /)
+
+flavors(PRIOR)     = flex_parser(flavor_string(PRIOR),     &
+                          int_options, string_options, 'input_nml:inf_flavor(1)')
+flavors(POSTERIOR) = flex_parser(flavor_string(POSTERIOR), &
+                          int_options, string_options, 'input_nml:inf_flavor(2)')
+
+end function set_inflation_flavor
```
nancycollins, line 2774: this looks good - the calling code isn't too messy and now you can use the same code for all the places where integers should be strings.

assimilation_code/modules/utilities/utilities_mod.f90
```diff
@@ -492,6 +493,101 @@ subroutine check_namelist_read(iunit, iostat_in, nml_name)
...
+   if (uppercase == possibility) then
+      flex_parser = integer_options(iopt)
+      exit LOOP
+   endif
+
+enddo LOOP
+
+if (flex_parser == MISSING_I) call error_report(context)
+
+contains
```
nancycollins, line 561: i like this code, but i'm not a fan of 'contains' in subroutines.  it confuses debuggers and profilers, and it prevents an embedded routine from being used anywhere else.  i realize this routine is specific to the flex_parser() code, but still i would move it to be a fully fledged subroutine but i wouldn't put it in the public list.  i might also rename it flex_parser_error() or something to tie it more closely to the parent routine.

assimilation_code/modules/utilities/utilities_mod.f90
```diff
@@ -492,6 +493,101 @@ subroutine check_namelist_read(iunit, iostat_in, nml_name)
...
+
+      call error_handler(E_MSG, 'flex_parser', msgstring1, source, &
+                 text2=msgstring2, text3=msgstring3)
+
+   do iopt = 1,size(integer_options)
+      write(msgstring1,*) integer_options(iopt), &
+                          ' = "'//trim(string_options(iopt))//'"' 
+      call error_handler(E_MSG,'flex_parser',msgstring1)
+   enddo
+   call error_handler(E_ERR,'flex_parser','Stopping.',source)
```
nancycollins, line 583: i actually never print 'stopping' because the error_handler routine could, in theory, be set to continue after E_ERR.  (i know that would break a lot of code - but it's possible.) 

i might replace line 583 with lines 567-571, so the last message is 'no valid option found'.  users often start at the last line in the error message and read back (or not).  the first msg string could start with line 572 - input is ... , then valid values, ... then something like 'not a valid option' as the last error message.  

(is this clear?  i feel like i might not have had enough coffee yet...)

nancycollins (commented):
looks good overall - i added a few small comments.

timhoar (commented):
Unfortunately ... I seem to have screwed up the RTPS ... the lorenz_63 model is throwing an error:
```
forrtl: severe (408): fort: (3): Subscript #1 of the array COPIES has value -1 which is less than the lower bound of 1

Image              PC                Routine            Line        Source             
filter             0000000000AF34E6  Unknown               Unknown  Unknown
filter             00000000005E41AD  ensemble_manager_        1468  ensemble_manager_mod.f90
filter             0000000000564CE4  filter_mod_mp_fil         834  filter_mod.f90
filter             0000000000404317  MAIN__                     20  filter.f90
filter             0000000000404262  Unknown               Unknown  Unknown
libc.so.6          00007EFC5426A6E5  __libc_start_main     Unknown  Unknown
filter             0000000000404169  Unknown               Unknown  Unknown
```

assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -2345,13 +2357,12 @@ function count_state_ens_copies(ens_size, post_inflate, prior_inflate) result(nu
...
-!   applying the inflation algorithm; hence we must save the prior ensemble spread in a different copy
-if ( inf_flavor(2) == 4 ) then ! CSS
-   SPARE_PRIOR_SPREAD = next_copy_number(cnum)
-endif 
+! If Whitaker/Hamill (2012) relaxation-to-prior-spread (rpts) inflation (inflation_flavor = 4)
+! then we need an extra copy to hold (save) the prior ensemble spread
+! ENS_SD_COPY will be overwritten with the posterior spread before
+! applying the inflation algorithm; hence we must save the prior ensemble spread in a different copy
+
+if (do_rtps_inflate(post_inflate)) SPARE_PRIOR_SPREAD = next_copy_number(cnum)
```
timhoar, line 2365: ### WHY IS THIS NOT WORKING?!

the `do_rtps_inflate(post_inflate)  call is returning FALSE even though `inf_flavor(2) == 4`

timhoar: Don't care anymore ... reverted to previous method.

nancycollins: don't forget to change filter_mod.dopplerfold.f90 accordingly if you're updating filter_mod.f90

assimilation_code/modules/utilities/utilities_mod.f90
```diff
@@ -480,6 +481,111 @@ subroutine check_namelist_read(iunit, iostat_in, nml_name)
...
+           text2=msgstring2, text3=msgstring3)
+
+do iopt = 1,size(integer_options)
+   write(string1,*) integer_options(iopt), ' = "'//trim(string_options(iopt))//'"' 
+   call error_handler(E_MSG,'flex_parser',string1)
+enddo
+
+! Repeat the leading message (as an E_ERR) to help delineate the problem.
+call error_handler(E_ERR, 'flex_parser', msgstring1, source, &
+           text2=msgstring2, text3=msgstring3)
```
nancycollins, line 584: this should only print msgstring1.  msgstring2 and 3 don't fit here.

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
```diff
@@ -29,8 +29,10 @@ module adaptive_inflate_mod
           get_is_prior,             get_is_posterior,       do_ss_inflate,      &
           set_inflation_mean_copy,  set_inflation_sd_copy,  get_inflation_mean_copy, &
           get_inflation_sd_copy,    do_rtps_inflate,        validate_inflate_options, &
-          print_inflation_restart_filename
-          
+          print_inflation_restart_filename, &
+          PRIOR, POSTERIOR, NO_INFLATION, OBS_INFLATION, VARYING_SS_INFLATION, &
```
nancycollins, line 33: prior and posterior are pretty generic names to be in a public list.  if they're specifically for the arrays of 2 arguments in the inflation namelist items, i'd suggest they be PRIOR_INF and POSTERIOR_INF.  PRIOR and POSTERIOR have lots of meanings to filter in many other contexts other than inflation (state copies, for example) and making these public might end up with making confusing collisions.

assimilation_code/modules/assimilation/filter_mod.dopplerfold.f90
```diff
@@ -392,30 +398,32 @@ subroutine filter_main()
...
 
 ! Initialize the adaptive inflation module
-call adaptive_inflate_init(prior_inflate, inf_flavor(1), inf_initial_from_restart(1), &
-   inf_sd_initial_from_restart(1), output_inflation, inf_deterministic(1),            &
-   inf_initial(1), inf_sd_initial(1), inf_lower_bound(1), inf_upper_bound(1),         &
-   inf_sd_lower_bound(1), inf_sd_max_change(1), state_ens_handle,                     &
+call adaptive_inflate_init(prior_inflate, inflation_flavor(PRIOR), inf_initial_from_restart(PRIOR), &
+   inf_sd_initial_from_restart(PRIOR), output_inflation, inf_deterministic(PRIOR),            &
+   inf_initial(PRIOR), inf_sd_initial(PRIOR), inf_lower_bound(PRIOR), inf_upper_bound(PRIOR),         &
+   inf_sd_lower_bound(PRIOR), inf_sd_max_change(PRIOR), state_ens_handle,                     &
```
nancycollins, line 411: longer term, the inflation options should move to their own namelist.  chunks of code like this are the argument for why being in the filter namelist isn't the right place for them.

assimilation_code/modules/utilities/utilities_mod.f90
```diff
@@ -73,6 +73,7 @@ module utilities_mod
           register_module, &
           find_namelist_in_file, &
           check_namelist_read, &
+          flex_parser, &
```
nancycollins, line 76: i get the intent, but "flex_xx" isn't very specific about what it's flexible about.  it's less catchy, but maybe 'int_from_string_value()' would make the code read more clearly?

int_option = int_from_string_value(string, ...)

seems pretty self-documenting?

assimilation_code/modules/utilities/utilities_mod.f90
```diff
@@ -480,6 +481,111 @@ subroutine check_namelist_read(iunit, iostat_in, nml_name)
...
 
+
+!-----------------------------------------------------------------------
+!> convert integers or strings describing options into integer. 
+!> 
+!> Some namelist items have historically been integers to describe inflation
+!> algorithms or ... which is a terrible thing to do to the users.
+!> This routine allows for backward compatibility and still provides
+!> the option to specify   inf_flavor = 'VARYING_SS_INFLATION' for example.
+
```
nancycollins, line 492: this is a routine in the utilities module, so it needs to be pretty general.  i know it was first implemented for inflation, but it needs to clearly be documented as working for anything.

```suggestion
!> Some namelist items have historically been integers which do not help describe the
!> option they are selecting.  They are being converted to descriptive strings.
!> During a transition period before the integers are deprecated this routine will return 
!> an integer value if the input string is either an integer or a string.  
!> For strings, the corresponding integer values are input as a 1-to-1 array with the
!> valid strings.   On error this routine prints an error message and does not return.
!> If needed, this routine could be changed to take an optional return code, which if 
!> present would return to the calling code without printing or calling the error 
!> handler so the caller can take an alternative code path.

```

nancycollins (commented):
the named parameters are much better than magic numbers, but the changes in the adaptive_inflate_mod are pretty extensive.  this needs a lot of careful testing for all inflation types to be sure that none of these changes alter the answers or provoke unexpected errors.  (i'm not aware of many good inflation test cases, to be honest.)

assimilation_code/modules/utilities/utilities_mod.f90
timhoar: I disagree. I chose to repeat the message string since the intervening messages can be quite extensive and I wanted to make sure they recognize it is all part of the same problem.

nancycollins: tim, please read the messages it's going to be printing?  strings 2 and 3 say "the following list" of items is valid.  the list precedes this message so it will sound like there's missing information.  repeating string 1 makes perfect sense to end on.

timhoar: You're right ... I can delete msgstring2, msgstring3 from that message.

timhoar:
> the named parameters are much better than magic numbers, but the changes in the adaptive_inflate_mod are pretty extensive. this needs a lot of careful testing for all inflation types to be sure that none of these changes alter the answers or provoke unexpected errors. (i'm not aware of many good inflation test cases, to be honest.)

OK - so I ran the L63 workshop_setup.csh on both the Manhattan release and the flex_filter_input  branch with 5 sets of namelist settings:

1. inf_flavor = 0, 0
2. inf_flavor = 2,2
3. inf_flavor = 3,3
4. inf_flavor = 2,4
5. inf_flavor = 5,5

In all tests, the obs_seq.final files were identical. Keep in mind the observation sequence file for that has 200 timesteps. The cutoff is crazy small, but the obs are located on the state variable locations (even thought they are not technically identity obs). Howzat?

jlaucar:
Inflation depends on the rest of the namelist parameters and the ensembles in the adaptive case. For all the adaptive algorithms there may be no inflation required if there is sufficient variability in the priors. The extreme localization means that obs are doing very little so this is a possibility.

nancycollins:
it may be too late for this pull request, and my apologies if you've already discussed this!  it seems like this might be better if split into 2 separate pull requests?  one change is new functionality (strings vs numbers in namelists) that needs new test cases, and one is a refactoring of code that should always reproduce previous results (changing magic numbers to named parameters).  both good and needed changes, but can be easily separated.  in the old svn days we did huge commits but i guess the better (newer, shinier) method is smaller, simpler changes in steps?  again, apologies if you've already discussed this - i don't mean to start trouble.  (really!  don't look at me like that!)

timhoar:
> Inflation depends on the rest of the namelist parameters and the ensembles in the adaptive case. For all the adaptive algorithms there may be no inflation required if there is sufficient variability in the priors. The extreme localization means that obs are doing very little so this is a possibility.

OK - so ... how would you test it?

jlaucar:
You need a case where we expect the inflation to do something. Maybe that's the case you have and the test is showing that things aren't working. We know that the appropriate localization for well-observed L96 cases is somewhere in the 0.2 range and that applications like that will turn on the adaptive inflation from time to time. The question remains, what are you trying to test? Are you trying to show that things are bitwise between the old and the new for the different flavors? 

Another way to guarantee that inflation turns on is to introduce model error in L96, say change the forcing to 5 or 12.

timhoar:
I created a small test case for L96 by creating a small observation sequence file with 1 identity observation (-10) and manually biased the observations. 

[input.nml.inf_test.txt](https://github.com/NCAR/DART/files/6029869/input.nml.inf_test.txt)
[obs_seq.out.inf_test.txt](https://github.com/NCAR/DART/files/6029870/obs_seq.out.inf_test.txt)
[CheckInflation.m.txt](https://github.com/NCAR/DART/files/6029943/CheckInflation.m.txt)

I ran the same inflation configurations as before (0,0 ; 2,2 ; 3,3 ; 2:4 ; 5,5) and then compared the entire contents of the  filter_output.nc files from Manhattan and from flex_filter_input.  All (inflated) cases had identical  inflation values except for the RTPS inflation. I'm looking into the difference.


```
>> CheckInflation
Comparing inflation case inf00 
                  MemberMetadata is a char variable.
                        inputnml is a char variable.

Comparing inflation case inf22 
                  MemberMetadata is a char variable.
                        inputnml is a char variable.

Comparing inflation case inf33 
                  MemberMetadata is a char variable.
                        inputnml is a char variable.

Comparing inflation case inf24 
                  MemberMetadata is a char variable.
                        inputnml is a char variable.
              state_postinf_mean has min/max differences of -3.41261 0.199984

Comparing inflation case inf55 
                  MemberMetadata is a char variable.
                        inputnml is a char variable.
```

assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -231,12 +234,13 @@ module filter_mod
                       obs_sequence_out_name = "obs_seq.final",  &
                       adv_ens_command       = './advance_model.csh'
 
-!                  == './advance_model.csh'    -> advance ensemble using a script
+! The inflation flavors are defined in adaptive_inflate_mod.
+! There are accessor routines such that filter_mod should not need to know
+! the  integer values relating to the flavor. It is an open question
+! as to whether or not we should 'use' the integer parameters for PRIOR
+! and POSTERIOR from adaptive_inflate_mod.
```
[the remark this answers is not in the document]
nancycollins: here are lines from the top of filter_mod.f90 which is the model for my suggestion that you use PRIOR_INF and POSTERIOR_INF in the adaptive inflation module:

> ! Defining whether diagnostics are for prior or posterior
> integer, parameter :: PRIOR_DIAG = 0, POSTERIOR_DIAG = 2

the actual question is not whether to 'use' them in the filter_mod - you must to be consistent - but whether to rename them to something more specific or not.

nancycollins:
a suggestion?  your test cases - both for the new int/string handling, and how you verify the inflation regression - should be added to the developer test directory.  i don't know how the testing framework is going to evolve, but we should try to collect as many tests as possible so we have them for testing future changes/regressions.

nancycollins (commented):
i'm sorry if these are duplicate comments.  if so, definitely my user error on git reviews.  someday i'll get the process down.  (several times i've started on the 'conversation' page instead of the 'files changed' one.)

timhoar:
## Testing

I created a wiki page describing the testing process:
https://github.com/NCAR/DART/wiki/Inflation-Testing

Note, this uses the integer representation of the inflation settings. It is trivial to repeat any of the tests with the character string specifications for the inflation algorithm. The bits needed to test are attached:

[inflation_test_pieces.tar.gz](https://github.com/NCAR/DART/files/6093238/inflation_test_pieces.tar.gz)

## Better Late than Never:
 - ensure that replacing the magic integers with strings reproduces when using integer specifications for inflation algorithms (i.e. `inf_flavor = 2,4`) for all possible inflation combinations.
 - ensure that  `inf_flavor = "VARYING_SS_INFLATION", "ENHANCED_SS_INFLATION"`  reproduces results from `inf_flavor = 2, 5` such that at least the character strings for the 6 inflation options work.
 - Keep in mind that OBS_INFLATION (1) is deprecated and terminates without running ({Manhattan})
 - Keep in mind that RTPS cannot be used as a prior inflation option.
 - Manhattan has a reporting error for RTPS such that the state-space output simply reports the namelist values rather than the computed inflation values. This has been fixed on ({master}) such that the state-space inflation values for RTPS tests across those branches are expected to generate differences. The observation sequence files (which include posterior values computed after RTPS inflation) are (still) identical.

## Documentation required
- [_not done_] specify this in the Changelog
- [_not done_] specify this in the filter.html - or do I wait till after the upcoming switch to .rst

## Outside the scope
 - incorporating test into the developer test suite
 - Checking consistency of Matlab inflation calculations with those from the Fortran

hkershaw-brown (changes requested):
looks good.  There are two changes I have added as suggestions. 

1. Is a definite change: minmax_sd(2)  is the max sd, not the posterior inflation sd. 

2.   I think the original message which stated that RTPS overrides the posterior inflation restart file is still valid.

Because RTPS is going to force the following:
inf_initial_from_restart(POSTERIOR_INF) = false.
inf_sd_initial_from_restart(POSTERIOR_INF) = .false.

I have put in a suggestion combining the new and old messages.

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
```diff
@@ -426,56 +444,67 @@ subroutine validate_inflate_options(inf_flavor, inf_damping, inf_initial_from_re
...
-   inf_initial_from_restart(2) = .false.    ! Get parameter from namelist inf_initial(2), not from file
-   inf_sd_initial_from_restart(2) = .false. ! inf_sd not used in this algorithm
+! Do not read in any files.  Also, no damping.  
+! Warn the user if they try to set different values in the namelist.
+if (inf_flavor(POSTERIOR_INF) == RELAXATION_TO_PRIOR_SPREAD) then
+
+   if ( inf_sd_initial_from_restart(POSTERIOR_INF)) &
+       call error_handler(E_MSG, 'validate_inflate_options:', &
+         'RTPS inflation (type 4) does not use "inf_sd_inifial_from_restart" values.', &
+         text2='values of "sd" in output inflation variables are not meaningful.')
```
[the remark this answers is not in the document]
timhoar: I disagree ... the message stated that "... overrides the posterior inflation restart **file** with value in namelist".
the RTPS PR changed that ... the posterior inflation file now has the actual inflation values - no matter what the namelist values were.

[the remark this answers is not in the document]
timhoar: Since RTPS does not use the inf_sd values at all ... it does parrot the sd namelist values into the output inflation file.
No matter what the 'mean' inflation values are in the namelist - or the input inflation files - they are completely disjoint from the inflation values calculated and used by RTPS ...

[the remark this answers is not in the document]
hkershaw-brown: The message is a warning about reading *from* inflation files,   if you have 

````
&filter_nml
inf_initial_from_restart = .true.
````

RTPS will not read the values from the file:

````
   ! Get parameter from namelist inf_initial(2), not from file
   ! inf_sd not used in this algorithm
      inf_initial_from_restart(POSTERIOR_INF) = .false.
     inf_sd_initial_from_restart(POSTERIOR_INF) = .false.
````

I think changing error messages for RTPS is outside the scope of this pull request,  I recommend 
switching out '2' for POSTERIOR_INF and leaving the rest of the message unchanged.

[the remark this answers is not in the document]
Comment (mgharamti) [reply assimilation_code/modules/assimilation/adaptive_inflate_mod.f90]: Helen, perhaps as you say this might not be in the scope of this PR. However, Tim is right. The message is wrong. After the correction that I made, this error message should be updated as Tim suggests. In fact, if I was more careful I should've done it in my PR originally.

File: assimilation_code/modules/assimilation/adaptive_inflate_mod.f90 [hunk 19]
```diff
@@ -622,8 +651,9 @@ end subroutine update_inflation
 !> Uses one of 2 algorithms in references on DART web site to update the 
 !> distribution of inflation:  Anderson 2007, 2009 or Gharamti 2017
 
-subroutine bayes_cov_inflate(ens_size, inf_type, x_p, sigma_p_2, y_o, sigma_o_2, lambda_mean, lambda_sd, &
-   gamma_corr, sd_lower_bound_in, sd_max_change_in, new_cov_inflate, new_cov_inflate_sd)
+subroutine bayes_cov_inflate(ens_size, inf_type, x_p, sigma_p_2, y_o, sigma_o_2, &
```
nancycollins, line 654: if this is just a line-wrap change, next time could this be deferred until a separate pull request?  i just spent time looking for the change in variable names for this routine.  (i already have 'ignore white space changes' set on.)

assimilation_code/modules/assimilation/filter_mod.dopplerfold.f90
```diff
@@ -228,10 +234,12 @@ module filter_mod
                       obs_sequence_out_name = "obs_seq.final",  &
                       adv_ens_command       = './advance_model.csh'
 
-! Inflation namelist entries follow, first entry for prior, second for posterior
-! inf_flavor is 0:none, 1:obs space, 2: varying state space, 3: fixed state_space,
-! 4: rtps (relax to prior spread), 5: enhanced state space
-integer  :: inf_flavor(2)                  = 0
+! The inflation flavor parameters are defined in adaptive_inflate_mod.
+! It is an open question as to whether or not we should 'use' the
```
nancycollins, line 238: i don't think that's an open question - to be consistent it's a good idea.  this comment could go.  it was an open question when they were simply called PRIOR and POSTERIOR.

assimilation_code/modules/assimilation/filter_mod.dopplerfold.f90
```diff
@@ -392,30 +402,54 @@ subroutine filter_main()
 
 call trace_message('Before initializing inflation')
 
-call validate_inflate_options(inf_flavor, inf_damping, inf_initial_from_restart, &
+inflation_flavor = set_inflation_flavor(inf_flavor)
```
nancycollins, line 405: why is a separate variable needed now?  maybe this needs a comment to explain why using inf_flavor isn't sufficient anymore

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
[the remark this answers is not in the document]
hkershaw-brown: Moha, can you create an issue for what the problem with the message is. At the moment it is not clear to me how your correction https://github.com/NCAR/DART/pull/165 relates to warning users about inflation files not being read in.   I thought it was a key point for users to be aware that
"RTPS inflation does not use any files (i.e., inflation has no-memory between DA cycles).:

Let's leave any RTPS changes out of this pull request and fix and RTPS message problems separately.

assimilation_code/modules/utilities/utilities_mod.f90
nancycollins: i still think this is not a good name for this routine.  flex doesn't say what it does, and technically it's not a parser - there is no syntax involved (like "a = b", "bob = call fred()", etc).  i wouldn't bother if it was a module private routine, but this is going into the utilities module for anyone to use and it's just going to propagate a confusing name into more code.

to be clear - i'm fine with the code itself, just not the name of the routine.

nancycollins (changes requested):
i went through this carefully and didn't see any bugs.  i'd like a comment about why there's a new variable needed in filter instead of using the namelist item; and i'd like to change name of the new routine that reads a string or integer.  but otherwise this looks good.

[whine: even with 'ignore white space', the line wrapping changes were hard to validate - that nothing was added or removed, nothing was renamed.  those changes mixed with real code changes have a cost in review time.]

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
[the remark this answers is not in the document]
nancycollins: i think these messages are about the *INPUT* files ( "from_restart" ).  didn't moha's changes only apply to the files it's writing out?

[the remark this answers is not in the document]
nancycollins: p.s. didn't we remove the namelist options to control writing out inflation restart files?  you can control if the values are read from a file or from the namelist, but if inflation is on you get restart files whether you want them or not.

assimilation_code/modules/assimilation/filter_mod.dopplerfold.f90
timhoar: OK - I'll remove the comment.

assimilation_code/modules/assimilation/filter_mod.dopplerfold.f90
timhoar: I cannot change the namelist variable name ... so 'inf_flavor' must now be a character instead of integer variable.
I needed an integer variable everywhere the old 'inf_flavor' was used, so I had to make a new variable.

assimilation_code/modules/utilities/utilities_mod.f90
timhoar: someone else is going to have to make the call on this.

assimilation_code/modules/assimilation/filter_mod.dopplerfold.f90
nancycollins: perfect explanation.  so:

```suggestion
! inf_flavor from the namelist is now a character string.  
! the inflation-related subroutines require an integer so the
! variable inflation_flavor is added and is type integer.
inflation_flavor = set_inflation_flavor(inf_flavor)
```

timhoar: Since I had to add it to both filter_mod.f90 and filter_mod.f90 ... I did this on the command line and pushed.

hkershaw-brown (approved):
outstanding issues reviewed in person yesterday and finished today.

nancycollins (approved):

PR #160, opened 2021-02-17
mirroring the Fortran algorithms for SS inflation

timhoar:
The GHA2017 inflation algorithm has been in Fortran for a while,
this takes the dartlab_rtps branch for PR 139 and make some changes
to increase the code similarity between adaptive_inflate_mod.f90 and
update_inflate.m


timhoar:
This is just a a debacle. The merge-point between the two branches is just wrong, wrong, wrong ... so nuisance changes are dominating the request.

PR #161, opened 2021-02-17
write FO values with obs_sequence_tool

criedel40:
The obs_sequence_tool was not writing out FO values when working with obs_seq.out files that contain FO values. I modified the obs_sequence_tool code along with DEFAULT_obs_sequence_mod.F90 to read the namelist option that users select to have FO values outputted.

### DIFF
--- a/assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
+++ b/assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
@@ -37,7 +37,8 @@ use obs_sequence_mod, only : obs_sequence_type, obs_type, write_obs_seq, &
                              get_obs_key, copy_partial_obs, &
                              delete_obs_from_seq, get_next_obs_from_key, &
                              delete_obs_by_qc, delete_obs_by_copy, &
-                             select_obs_by_location, set_obs_values, set_qc
+                             select_obs_by_location, set_obs_values, set_qc, &
+                             set_obs_seq_precomputed_FOs
 
 implicit none
 
@@ -740,6 +741,9 @@ if (.not. print_only) then
 else
    print*, 'Total number of selected obs in all files :', get_num_key_range(seq_out)
 endif
+
+!setting flag to output precomputed FOs based on namelist
+call set_obs_seq_precomputed_FOs(seq_out,write_external_FOs)
 
 call print_obs_seq(seq_out, filename_out)
 if (.not. print_only) then
--- a/observations/forward_operators/DEFAULT_obs_def_mod.F90
+++ b/observations/forward_operators/DEFAULT_obs_def_mod.F90
@@ -839,7 +839,7 @@ if ( obs_def%has_external_FO .and. obs_def%write_external_FO ) then
          source, revision, revdate, text2='observation type '//trim(get_name_for_type_of_obs(obs_def%kind)))
    endif
    if (is_ascii) then
-      write(ifile, 12) obs_def%ens_size, obs_def%external_FO_key
+      write(ifile, 12) obs_def%ens_size, key
       write(ifile, *) (obs_def%external_FO(ii), ii=1,obs_def%ens_size)
    else
       write(ifile)    obs_def%ens_size, EXTERNAL_PRIOR_CODE
--- a/assimilation_code/modules/observations/obs_sequence_mod.f90
+++ b/assimilation_code/modules/observations/obs_sequence_mod.f90
@@ -29,7 +29,8 @@ use      obs_def_mod, only : obs_def_type, get_obs_def_time, read_obs_def, &
                              write_obs_def, destroy_obs_def, copy_obs_def, &
                              interactive_obs_def, get_obs_def_location, &
                              get_obs_def_type_of_obs, get_obs_def_key,  &
-                             operator(==), operator(/=), print_obs_def
+                             operator(==), operator(/=), print_obs_def, &
+                             set_obs_def_write_external_FO
 
 use     obs_kind_mod, only : write_type_of_obs_table, &
                              read_type_of_obs_table, &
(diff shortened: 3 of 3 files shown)


observations/forward_operators/DEFAULT_obs_def_mod.F90
```diff
@@ -839,7 +839,7 @@ subroutine write_obs_def(ifile, obs_def, key, fform)
          source, revision, revdate, text2='observation type '//trim(get_name_for_type_of_obs(obs_def%kind)))
    endif
    if (is_ascii) then
-      write(ifile, 12) obs_def%ens_size, obs_def%external_FO_key
+      write(ifile, 12) obs_def%ens_size, key
```
nancycollins, line 842: remind me why this change is needed?  is this the main obs key (1,2,3, etc in the sequence regardless of obs type) or is this the per-type key?  i know it will get ignored on read, but for tracking problems is this going to be per-type key that would be printed out by debugging statements during a run?

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
```diff
@@ -741,6 +742,9 @@ program obs_sequence_tool
    print*, 'Total number of selected obs in all files :', get_num_key_range(seq_out)
 endif
 
+!setting flag to output precomputed FOs based on namelist
+call set_obs_seq_precomputed_FOs(seq_out,write_external_FOs)
```
nancycollins, line 746: with the team's approval, i'd like to leave write_external_FOs as a local logical but remove it from the namelist and add one of these two: 

discard_precomputed_FOs or 
remove_precomputed_FOs 

as string arrays, before this gets entrenched in too many namelists.   the initial values are "" (empty string) and the only valid value accepted is "all" for now.  if it's "", write_external_FOs stays true (should be the default value) and if it's "all" then it changes to false.  i think then the rest of the code works as-is.

at some future point someone can add support for getting the index values from a list of obs_type names and turning them off individually.  but this is sufficient for now, i'd think.

nancycollins (commented):
i've got a question, and one suggestion - but otherwise i think these changes look good.

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
timhoar: What I implemented supports getting index values for a list of obs_type names and turns the write flag ON for them.
You can check out the branch I supplied and compare those differences to any branch you like with the following syntax:
```
git fetch upstream
git branch -b tims_crazy_branch upstream/criedel40-write_FOs_seq_tool
git diff upstream/master
```

If you want more surgical control, you can do the following:
```
git diff upstream/master -- obs_sequence_tool.f90
```

If you have the 'gitdiff' alias, you can replace `git diff` with `gitdiff` and it whips up the differences in diffuse.

The present default behavior is to suppress writing out the precomputed FOs (i.e. they are already discarded or removed), so I think the namelist variable should have the opposite implication: preserve_precomputed_FOs

nancycollins: i actually think the default behavior for this tool should be to default to preserving the info, and letting the user choose to discard it if they have a specific need.

timhoar:
I am reviewing Chris' branch 'as is' and can verify that 'obs_sequence_tool' and 'filter' work as expected. Perhaps a silly test, but 'perfect_model_obs' fails. How much weight do we want to put on this? Why would anyone want to run 'perfect_model_obs' to harvest synthetic observations when they have precomputed values? To compare forward operators? I'm playing devils' advocate here ...  if you turn off the 'use_precomputed_FOs_these_obs_types', 'perfect_model_obs' works as expected.

timhoar:
The problem is in get_expected_obs ... some arrays have length 'ensemble size',
which may not match what is declared in the number of ensemble members in the precomputed FO. 

```
   ! the decision process here is that if some external source computed
   ! the prior forward operator values then we can use them or not, but
   ! if we use them there is no way to compute a consistent posterior.
   ! so the posteriors are always marked as 'failed forward operator'.
   if (use_precomputed_FO) then
      if (isprior) then
         if ( obs_def%has_external_FO ) then
            expected_obs(:) = obs_def%external_FO(:)
            istatus = 0
         else
            call error_handler(E_ERR, 'get_expected_obs_from_def', &
                  'Attempt to access an external FO that is not present in the observation information.', &
                   source, revision, revdate, text2='observation type '//trim(get_name_for_type_of_obs(obs_def%kind)))
         endif
      else ! posterior - missing value
         expected_obs(:) = missing_r8
         istatus = 1
      endif
```

nancycollins:
i'm surprised.  i would think it shouldn't be an error to run with _fewer_ than the number of precomputed ensemble members, only _more_.  this code should copy (1:ens_size) on both arrays (assuming ens_size is available at this point in the code?)

if there are no earlier tests to be sure the number of FOs is >= ens_size, then i guess i'd test it here before doing the assignment.

timhoar:
As far as I can tell, even on the master branch, there is no error-checking to see if the number of precomputed values is consistent with the ensemble size of the program. I have tested the following in `DEFAULT_obs_def_mod:get_expected_obs_from_def()` and it seems to have the desired behavior (even pmo works).  But ... since this is a separate bugfix, I am considering putting it on a separate pull request.

```
      if (obs_def%ens_size < ens_size) then
         write(string1,'(A,1x,I4,1x,A,1x,I4)')'The number of precomputed forward operators (', &
                      obs_def%ens_size, ') is smaller than the ensemble size of ', ens_size
         write(string2,*)'observation type '//trim(get_name_for_type_of_obs(obs_def%kind))
         call error_handler(E_ERR, 'get_expected_obs_from_def', string1, source, text2=string2)
      endif

      if (isprior) then
         if ( obs_def%has_external_FO ) then
            expected_obs(:) = obs_def%external_FO(1:ens_size)
            istatus = 0
         else
  ...
```

nancycollins:
great idea to make this a separate pull request.  i've been saying things should be broken up into smaller requests but it's hard to remember to actually do it.

timhoar (approved):
Testing revealed an unrelated sub-optimal behavior in that pmo bombs ungracefully if there is more than 1 precomputed FO value and filter bombs ungracefully when the number of precomputed FO values is less than the ensemble size.

To be clear, bounds checking has to be on for filter to bomb. Otherwise, (the gfortran compiler, at least) just sticks in bogus values for the 'extra' ensemble members.

Making a separate Issue.

hkershaw-brown:
@criedel40 thanks Chris! you are officially a DART contributor.

criedel40:
Yay!! Life goal check mark.
Have a good weekend everyone!
-Chris

observations/forward_operators/DEFAULT_obs_def_mod.F90
nancycollins: this is now a merged pull request, but did anyone answer why the key on this line changed?

timhoar:
Chris indicated that the external_FO_key was not correct and did not help
locate the observation in the file. Using 'key' did.
Chris talked with Glen about this and Chris reported to me that Glen was
aware of the problem but since it did not cause an issue, he did not feel
compelled to change it.

PR #163, opened 2021-02-19
Bug fix to allow reflectivity forward operator limit check.

weather4evr:
In observations/forward_operators/obs_def_radar_mod.f90, if reflectivity (QTY_RADAR_REFLECTIVITY) is in the state, right now, the subroutine get_expected_radar_ref immediately "return"s.  However, by returning, the subroutine never gets to apply the "apply_ref_limit_to_fwd_op" check, which is problematic.

This simple fix actually allows the "apply_ref_limit_to_fwd_op" check to happen if reflectivity is in the state.


hkershaw-brown (approved):
Sensible fix, thanks!

@hkershaw-brown  todo: put on master branch also [done]

Issue #162, opened 2021-02-18
source code variants out of sync


hkershaw-brown:
There are two files in DART which have 2 variants.  The variants are have not been updated along with the regular filter_mod.f90 and assim_tools_mod.f90.  The only differences should be the intentional algorithm differences.  

1. filter_mod.f90   filter_mod.dopplerfold.f90  

2. assim_tools_mod.f90   assim_tools.pf.f90

For details on why these variants exist:
https://wiki.ucar.edu/display/DIME/Updating+Variants+in+Source+Code

PR #164, opened 2021-02-19
Keep variants in sync

hkershaw-brown:
Fixes issue #162

assim_tools_mod.pf.f90 is whitespace changes and one comment block.  These are a bit of a nuisence to review here, but should make future comparisons easier.  Probably the best way to review assim_tools_mod.pf.f90 is to compare 
`assim_tools_mod.pf.f90` with `assim_tools_mod.f90` and confirm that the only changes are the pf filter. 

filter_mod.dopplerfold.f90 has code changes to keep up with filter_mod.f90
style guide updates:
       dart_time  (time is a fortran intrinsic)
       character len

code updates:
       if statement with  .not. do_rtps_inflate(post_inflate))
       allocation fix for intel 19
       create_ensemble from single file uses dart iunit


timhoar (approved):
This all looks right to me.

nancycollins (commented):
these changes all look fine.

i want to point out that changing the name of a subroutine argument is not completely transparent to the user if they have code that uses the calling syntax:   
   call bob(arg1 = 1, arg2 = "fred", ...)

i understand 'time' is a fortran intrinsic and to avoid confusion on the programmer's side it's clearer not to use it.  but the fortran standard allows you to use intrinsic names as variable names and the compilers have no problem knowing the difference.
 
in this case it seems safe to assume no one is using the keyword syntax to call these subroutines and i approve this change.  but at some future time, especially if there are optional arguments to routines, then there *is* a good chance callers may be using the name=value syntax in the calling code, so i wouldn't change those even if the variable name is the same as an intrinsic.   </lecture>

hkershaw-brown:
Thanks for this point Nancy.  I'll add to the style guide because it is an important consideration when making changes to variable names. 
The `time` argument was changed to dart_time` in filter_mod.f90 in 2017, but I'll be on the lookout for filter.doppler users who would be affected by this.

PR #165, opened 2021-02-19
Improved the support for RTPS (inf_flavor 4) posterior inflation: 

hkershaw-brown:
From Moha:

Improved the support for RTPS (inf_flavor 4) posterior inflation: 

Previously, one could only decide on a single parameter (say, alpha) in the input nml (as inf_initial(2)) without being able to see what is the effective posterior inflation the RTPS scheme is actually using. In other words, if you check the resulting inflation in preassim.nc for instance (variable state_postinf_mean) it would be the same as inf_initial(2). That's incorrect. 

To clarify, there is nothing wrong in the RTPS implementation, however, reporting the output has an issue. The following update corrects the issue and updates the output files with the right posterior inflation values. See the figure below for illustration. The left hand side plot is the value of state_postinf_mean in preassim.nc (or analysis.nc) after running filter. The right hand side is the plot after the proposed code change. Note that both runs use inf_flavor(2) = 4 and inf_initial(2) = 0.3

Screen Shot 2021-01-13 at 10 28 30 AM

Keep in mind that RTPS inflation does not use any files (i.e., inflation has no-memory between DA cycles).

### DIFF
--- a/assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
+++ b/assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
@@ -250,6 +250,15 @@ inflate_handle%allow_missing_in_clm = missing_ok
 inflate_handle%mean_from_restart    = mean_from_restart
 inflate_handle%sd_from_restart      = sd_from_restart
 
+!Overwriting the initial value of inflation with 1.0
+!as in other inflation flavors (usually start from 1). 
+!This is required for RTPS because inf_initial(2)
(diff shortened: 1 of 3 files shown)


nancycollins (approved):
these changes all look good to me.

mgharamti (commented):
Thanks Helen for breaking my messy PR into pieces. And thanks also for adding the change to the filter-mod-doppler version. Everything looks good.

PR #167, opened 2021-02-24
Modify logic for outlier_threshold < 0

weather4evr:
Currently, to disable the outlier check, we set outlier_threshold < 0 in the namelist.
However, doing so effectively eliminates logic related to **enable_special_outlier_code**; if outlier_threshold < 0, quality_control_mod.f90 never gets to the part where it checks if enable_special_outlier_code is enabled.

This is problematic for someone who might want to disable the outlier check for all obs types _except for _some__.  I came across this issue when I was using precomputed forward operators for all observation types except for radar obs, and I wanted to keep DART's outlier check for radar obs but wanted to eliminate the check for everything else.

Suggested fix restores functionality of enable_special_outlier_code when outlier_threshold < 0

### DIFF
--- a/assimilation_code/modules/assimilation/quality_control_mod.f90
+++ b/assimilation_code/modules/assimilation/quality_control_mod.f90
@@ -270,7 +270,8 @@ logical  :: failed
 ! only if it is still successful (assim or eval, 0 or 1), then check
 ! for failing outlier test.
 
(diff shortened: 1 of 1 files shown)


nancycollins (approved):
these changes look fine to me.

hkershaw-brown:
@hkershaw-brown todo: cherry-pick to master

PR #169, opened 2021-02-26
checking ens_size versus number of FOs

timhoar:
Explicity using the first 'ens_size' precomputed values.
Adding and error check if there are not enough precomputed values for the ens_size.

### DIFF
--- a/observations/forward_operators/DEFAULT_obs_def_mod.F90
+++ b/observations/forward_operators/DEFAULT_obs_def_mod.F90
@@ -524,6 +524,8 @@ integer             :: obs_key
 real(r8)            :: error_var
 logical             :: use_precomputed_FO
 
+character(len=512) :: string1, string2, string3
+
 ! Load up the assimilate and evaluate status for this observation kind
 assimilate_this_ob = assimilate_this_type_of_obs(obs_kind_ind)
 evaluate_this_ob = evaluate_this_type_of_obs(obs_kind_ind)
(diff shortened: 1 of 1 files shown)


observations/forward_operators/DEFAULT_obs_def_mod.F90
```diff
@@ -542,9 +544,18 @@ subroutine get_expected_obs_from_def_distrib_state(state_handle, ens_size, copy_
...
+         write(string2,*)'observation type '//trim(get_name_for_type_of_obs(obs_def%kind))
+         write(string3,*)'precomputed value(1) ', obs_def%external_FO(1)
+         call error_handler(E_ERR, 'get_expected_obs_from_def', string1, source, &
+                    text2=string2, text3=string3)
+      endif
+
       if (isprior) then
          if ( obs_def%has_external_FO ) then
-            expected_obs(:) = obs_def%external_FO(:) 
+            expected_obs(:) = obs_def%external_FO(1:ens_size)
```
timhoar, line 558: Explicitly use 'N' of them (satisfies bounds-checking for both pmo and filter).

observations/forward_operators/DEFAULT_obs_def_mod.F90
```diff
@@ -542,9 +544,18 @@ subroutine get_expected_obs_from_def_distrib_state(state_handle, ens_size, copy_
...
    if (use_precomputed_FO) then 
+      if (obs_def%ens_size < ens_size) then
+         write(string1,'(A,1x,I4,1x,A,1x,I4)')'The number of precomputed forward operators (', &
+                      obs_def%ens_size, ') is smaller than the ensemble size of ', ens_size
+         write(string2,*)'observation type '//trim(get_name_for_type_of_obs(obs_def%kind))
+         write(string3,*)'precomputed value(1) ', obs_def%external_FO(1)
+         call error_handler(E_ERR, 'get_expected_obs_from_def', string1, source, &
+                    text2=string2, text3=string3)
+      endif
+
```
timhoar, line 555: Checks to make sure there are enough precomputed values for the desired ensemble size.

nancycollins (approved):
this looks like it fixes a bug that's been around for a while.

hkershaw-brown (approved):
This particular change is fine, because it fixes a bounds error.

However, 
````
expected_obs(:) = obs_def%external_FO(1:ens_size) 
````
I don't think is correct if `distributed_state = .false.`
Each task doing a forward operator will get `obs_def%external_FO(1:1) ` which gives the same FO for each ensemble member. I'll make a new issue for this, because I'm not sure of the design history of the external FO code.

PR #170, opened 2021-02-27
selectively write FO values  based on obs type

timhoar:
The obs_sequence tool had an 'all-or-nothing' approach to writing out the precomputed forward operators values.
This change allows people to select which types of observations should write out their precomputed forward operator values.

The calling structure is a little simpler (in my opinion) in that the obs_sequence_mod.f90 is the same as it was in ec49b7c30e7ec9ed658148d0e4f8cb19cca022ee


assimilation_code/modules/observations/obs_sequence_mod.f90
```diff
@@ -29,8 +29,7 @@ module obs_sequence_mod
                              write_obs_def, destroy_obs_def, copy_obs_def, &
                              interactive_obs_def, get_obs_def_location, &
                              get_obs_def_type_of_obs, get_obs_def_key,  &
-                             operator(==), operator(/=), print_obs_def, &
-                             set_obs_def_write_external_FO
+                             operator(==), operator(/=), print_obs_def
```
timhoar, line 32: The obs_sequence_mod is reverted to ec49b7c30e7ec9ed658148d0e4f8cb19cca022ee

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
```diff
@@ -8,21 +8,30 @@
...
+                             do_nml_file, do_nml_term, set_filename_list, &
+                             to_upper
+
 use     location_mod, only : location_type, get_location, set_location, &
                              LocationName !%! , vert_is_height 
-                             ! see comment in select_gps_by_height() for explanation of !%!
+                             ! comment in select_gps_by_height() explains !%!
+
 use      obs_def_mod, only : obs_def_type, get_obs_def_time, get_obs_def_type_of_obs, &
                              get_obs_def_location, set_obs_def_write_external_FO
```
timhoar, line 25: The obs_sequence_tool already uses the `set_obs_def_write_external_FO` routine.

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
```diff
@@ -8,21 +8,30 @@
...
+
+use     obs_kind_mod, only : max_defined_types_of_obs, get_name_for_type_of_obs, &
+                             get_index_for_type_of_obs
+
 use time_manager_mod, only : time_type, operator(>), print_time, set_time, &
                              print_date, set_calendar_type, GREGORIAN
+
 use obs_sequence_mod, only : obs_sequence_type, obs_type, write_obs_seq, &
-                             init_obs, assignment(=), get_obs_def, &
+                             init_obs, assignment(=), get_obs_def, set_obs_def, &
```
timhoar, line 34: After you've used the `set_obs_def_write_external_FO` routine, you need to be able to set the observation value to the new obs_def, so you need 'set_obs_def'

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
```diff
@@ -79,6 +87,13 @@ program obs_sequence_tool
 integer :: copy_index_len      = 0
 integer :: qc_index_len        = 0
 
+
+! variables to manage whether or not the precomputed forward operator values
+! are written out
+
+integer :: FO_write_types(max_obs_input_types) = MISSING_I
+integer :: num_FO_types_to_write
```
timhoar, line 95: Since the existing behavior is to not write out any observation types ... I chose to let them choose which ones to write out (rather than which ones to suppress).

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
```diff
@@ -141,7 +155,7 @@ program obs_sequence_tool
          min_gps_height, edit_copy_metadata, new_copy_index,                 &
          edit_qc_metadata, new_qc_index, synonymous_copy_list,               &
          synonymous_qc_list, edit_copies, edit_qcs, new_copy_metadata,       &
-         new_qc_metadata, new_copy_data, new_qc_data, write_external_FOs
+         new_qc_metadata, new_copy_data, new_qc_data, external_FO_write_list
```
timhoar, line 158: The write_external_FOs variable was an all-or-nothing logical. I think this is more descriptive.
AFAIK, there is no namelist in the repo that actually had the 'write_external_FOs' variable specified.

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
```diff
@@ -642,6 +666,9 @@ program obs_sequence_tool
          obs_out = obs_in
       endif
 
+      ! changing metadata to output precomputed FOs based on namelist
+      call set_FO_metadata(obs_out)
+
```
timhoar, line 671: Here's the magic ... one observation at a time ... this is the first one.

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
```diff
@@ -659,7 +686,10 @@ program obs_sequence_tool
             endif
          endif
 
-         obs_in     = next_obs_in   ! essentially records position in seq_out
+         obs_in = next_obs_in   ! essentially records position in seq_out
+
+         !setting flag to output precomputed FOs based on namelist
+         call set_FO_metadata(obs_in)
```
timhoar, line 692: And here's all the rest.

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.nml
```diff
@@ -35,6 +35,7 @@
    synonymous_qc_list   = ''
    print_only           = .false.
    gregorian_cal        = .true.
+   external_FO_write_list = 'ALL'
```
timhoar, line 38: As an example.

timhoar (commented):
I have tested it with perfect_model_obs, filter, and obs_sequence_tool (on observation sequence files with observations that have some observations with FOs and some without).

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
nancycollins: i like the magic.  

99% of obs aren't going to have any external FOs or if they do, this tool won't want to change what's there. i guess i'd suggest setting some module global flag during the namelist processing that lets you avoid calling this routine if no changes are going to be made. it simplifies the code path if someone is trying to trace through here and doesn't know or care what FO metadata is. something like:

if (deleting_external_FOs) call set_FO_metadata(obs)

nancycollins (commented):
if you want to talk about changing the default behavior, i'm happy to.   it seems like filter (and PMO) removing it is the unusual case.  the rest of the tools should preserve the external FOs by default.

nancycollins:
github is still a mystery to me.  that is all.

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90
```diff
@@ -94,6 +109,7 @@ program obs_sequence_tool
 integer  :: last_obs_days     = -1
 integer  :: last_obs_seconds  = -1
 
+character(len=obstypelength) :: external_FO_write_list(max_obs_input_types) = ''
```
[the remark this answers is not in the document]
hkershaw-brown: The comments are in markdown, so if you want to quote code use backticks  around the code: `hello nancy`.

For several lines use "```"
```
Hello Nancy!
How are you?
```

nancycollins:
is this an example of needing some kind of issue/proposed solution/design workflow thingie before doing the code writing?  i feel awkward about asking for a bunch of changes on already-written and tested code, which doesn't seem like the best way to keep the code base self-consistent and solving the right problems.

hkershaw-brown:
Here is the spec Tim and I hashed out for this pull request


--

## Proposal for obs\_sequence\_tool

Terminology:
`precomputed_FO` rather than `external_FO`.  This is to match the filter namelist option
`use_precomputed_FOs_these_obs_types` 

Default behavior:
 
* write out all precomputed_FO values
* write out the key - the key is only relevant for the output obs_sequence file. 


User options:

 * switch off writing all precomputed\_FO values
 * switch off writing precomputed\_FO values by observation type


### Namelist options: 


#### new:
````
&obs_sequence_tool_nml
  remove_precomputed_FO_values = ‘all’
                                 ‘a’, ‘b’, ‘c’
````

where ‘a’, ‘b’, ‘c’ are observation types. 

'all' switches off writing all precomputed\_FO values

'a' switches off writing precomputed\_FO values for observation type 'a'

#### removed:
````
&obs_sequence_tool_nml
  write_external_FOs
````

### User documentation required:

For now, the documentation is being added to the DART wiki.  To be moved when the DART
docs are in .rst form. 

1.   A description for obs_def for a precomputed\_FO
    
````
     OBS            2
       253.050362373913     
      0.000000000000000E+000
               1           3          -1
    obdef
    loc3d
         1.468188332803687         1.311137934389169         0.000000000000000     -1
    kind
              17
    external_FO      79       2
````

External_FO `ensemble size`  `key`

2.  New namelist option `remove_precomputed_FO_values`


### Things that are outside the scope of this pull request:
* Dealing with obs_sequence files that have differing ensemble sizes for FO. 
* Test suite for obs\_sequence\_tool

File: assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90 (hunk 6 above)
Comment (hkershaw-brown) [reply assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.f90:671]: This is a good idea, chatted with Tim about this. It is a non-user facing change so we can put it on the todo list for the next round of improvements to obs_sequence_tool.

**Comment by nancycollins** [comment]
are you soliciting comments on the proposal?  if not, ignore the rest of this.  if so, read on.

this looks good.  it summarizes what is going to change and what behavior is expected.  it also describes what isn't going to be part of this pull request, so the pull request is well defined.  i like leaving a list of what needs to be addressed in the future.

the tool namelist changes and options definitely need updated docs.  also, the obs_sequence_tool html page had an extensive examples section - adding a very small example there might be appropriate. 

a full test suite for the tool is definitely outside the scope of this pull request.  did you talk about adding a small test case w/ a 2-3 entry obs_seq file including one obs with and one without precomputed FOs for the developer_tests directory?  or is that spreading the focus too broad?

**Comment by hkershaw-brown** [comment]
Yes, definitely soliciting comments. 

Re: small test case in the developer_tests directory
Morally, yes a small test in developer_tests is within the scope of this pull request. 
Practically, this pull request has been hand tested.  I'm going to take the todo of adding an automated test for obs_sequence_tool. 

For upcoming features, we do:
 spec -> review -> tests -> code -> review -> victory

Good point about an example in the obs_sequence_tool.html.

**Comment by timhoar** [comment]
I wrote a Wiki page: https://github.com/NCAR/DART/wiki/Precomputed-Forward-Observations-(FOs)    which incorporates information from Nancy (via email) and outlines my testing procedure.

File: assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.html [hunk 10]
```diff
@@ -790,6 +803,35 @@ <h3>Subset by Observation or QC Value</h3>
...
+# keep all precomputed values from all observations with precomputed values (the default):
+   remove_precomputed_FO_values = ''
+
+# remove all precomputed values from all observations with precomputed values:
+   remove_precomputed_FO_values = 'ALL'
+
+# remove all precomputed values for specific observation types (case does not matter):
+# The observations themselves will still be present in the output, given no other
+# subsetting processing.
+   remove_precomputed_FO_values = 'RADIOSONDE_temperATURE', 'AMDAR_U_WIND_COMPONENT'
```
hkershaw-brown, line 832:
```suggestion
   remove_precomputed_FO_values = 'RADIOSONDE_TEMPERATURE', 'AMDAR_U_WIND_COMPONENT' 
```

hkershaw-brown (approved):
looks great!

There's one small typo: `RADIOSONDE_temperATURE`.  I put in a suggestion to fix it. 

Also really nice wiki documentation.

nancycollins (approved):
looks really good, thanks.

assimilation_code/programs/obs_sequence_tool/obs_sequence_tool.html
timhoar: The bizarre case was intended to demonstrate that case does not matter. I agree it looks funny.

PR #179, opened 2021-03-11
Documenting RTPS inflation algorithm

mgharamti:
Added a user-guide on how to use RTPS inflation. Also,
added few sentences to update already existing text with
recent research. Missing references have also been appended.


nancycollins (approved):
looks good moha.  thanks for doing this.  i have one typo suggestion to fix.

someone does need to keep track that this new content gets migrated into the rst version of the docs once those are ready.

hkershaw-brown (approved):
looks great

Issue #173, opened 2021-03-05
note on external FO and distributed_state not being compatible. 


hkershaw-brown:
Here is a test case (run with lorenz_96)

Incorrect forward operator values when:
* Using external FO 
* `distributed_state = .false.`  
* more than 1 processor

````
 obs_sequence
obs_kind_definitions
           1
           2  RAW_STATE_VARIABLE
  num_copies:            1  num_qc:            1
  num_obs:          1  max_num_obs:            1
observation                                                     
Data QC                                                         
  first:            1  last:          1
 OBS            1
    2.376289999
  0.000000000000000E+000
          -1           -1          -1
obdef
loc1d
   0.3900425101203420
kind
          2
external_FO      20       2
   1.00    2.00	    3.00
   4.00    5.00	    6.00
   7.00    8.00	    9.00
   10.00   11.00    12.00
   13.00   14.00    15.00
   16.00   17.00    18.00
   19.00   20.00  
        3600            0
   1.0000000000000000
````

for 1 processor, out_seq.final is as expected (1-20).  For 4 processes, get 1-5 four times.  

[obs_seq.FO.txt](https://github.com/NCAR/DART/files/6093442/obs_seq.FO.txt)
[obs_seq.final.n1.txt](https://github.com/NCAR/DART/files/6093426/obs_seq.final.n1.txt)
[obs_seq.final.n4.txt](https://github.com/NCAR/DART/files/6093427/obs_seq.final.n4.txt)

hkershaw-brown:
Fix for this on bugfix/external_FO_non_distrib

Each task has every copy of the precomputed FO because the whole ensemble is in obs_def%external_F0

Each task calculates the forward operator for the copies it has. 
For `distributed_state = true` this is the whole ensemble
For `distributed_state = false` this is a subset of the ensemble.  

bugfix/external_FO_non_distrib adds a loop in `get_expected_obs_from_def_distrib_state` to grab the correct subset of copies from  obs_def%external_F0.

PR #180, opened 2021-03-12
bug fix for external forward operators when using distributed_state=.false.

hkershaw-brown:
Fixes https://github.com/NCAR/DART/issues/173

Each task has every copy of the precomputed FO because the whole ensemble is in `obs_def%external_F0`

Each task calculates the forward operator for the copies it has.
For distributed_state = true this is the whole ensemble
For distributed_state = false this is a subset of the ensemble.

This fix adds a loop in `get_expected_obs_from_def_distrib_state` to grab the correct subset of copies from `obs_def%external_F0.`


nancycollins (approved):
this is the kind of pull request i can honestly say i enjoy reviewing!!  :)

PR #196, opened 2021-04-05
Introduction of variational bias correction for radiances

weather4evr:
This is a big PR.  Sorry!  However, I'm not really sure how to break it into smaller chunks.

Overall, these modifications implement the variational bias correction capability for radiance observations from the GSI-based EnSRF into DART.  The newly added code has been tested against the GSI-EnSRF and replicates GSI-EnSRF results to at least 3 decimal places, which is good enough (some small differences should be expected).  This code has also been used to cycle MPAS-DART for 4 weeks (with external forward operators and GSI-based bias correction coefficients), and the bias correction seems to be working.  The important part about this capability is that it updates the bias correction _coefficients_; no longer is a standalone variational DA system required to update these coefficients each assimilation cycle.
 
Changes to assim_tools_mod.f90 are fairly self-contained, and moreover, general.  Right now, only bias correction based on GSI conventions will work, but adding bias correction from different sources is theoretically possible.  In fact, with some luck, no changes should be required in assim_tools_mod.f90 to add bias correction from a different source (including DART's RTTOV implementation); adding a new source is basically an I/O issue handled in radiance_bias_mod.f90.   The bias correction is general and does not know anything about the physical meanings of specific predictors.  Predictors for each radiance observation are provided by the obs_def structure, which I extended.

A somewhat unintended consequence of the modifications is that all programs relying on assim_tools_mod.f90 now depend on two new added modules, so all path_names* files with assim_tools_mod.f90 need to be modified, and since the new modules explicitly have MPI routines (DART doesn't seem to have the needed wrapper), MPI compilation must be used on programs dependent on assim_tools_mod.f90.  I've dealt with this issue for WRF and MPAS_atm (hence all the modified files in the "work" directories).  Perhaps there's a way around this altogether?

You probably won't like many of the changes, and I'm happy to work with you to find solutions.  As hopefully more folks use RTTOV within DART, this capability might become very useful for DART users.  The good news is that these changes are invisible to users who aren't interested in this capability.


hkershaw-brown:
Hi Craig, do you have a small(ish) test case for this on Cheyenne?

weather4evr:
Yes: /glade/scratch/schwartz/JTTI/VarBC_test2/2017042300/dart_analysis/d01/working

Just copy everything someplace and qsub submit.csh should run it.

nancycollins:
hi craig.  which mpi routines do you need that don't have wrappers in the mpi_utilities_mod.f90 file?  i looked through assim_tools_mod.f90 but didn't see any calls that i didn't recognize.  the usual way we handle this is to add a cover routine to both mpi_utilities_mod.f90 and to null_mpi_utilities_mod.f90.  the latter is usually a no-op or a copy or whatever makes sense if you're running with a single task.  then the program should compile both ways.

nancycollins:
if this is a general bias correction facility, should the module maybe not have radiance in the name?  and if it introduces any additional dependencies which a different model (e.g. lorenz 96, compiled with 1D locations and no real observation types) would not compile with, it might be necessary to also introduce a dummy module with the same module name and the same entry points, which does nothing if that's not harmful, or calls the error handler if it's called and expected to return something.  that follows how we avoid mpi and window dependencies.

weather4evr:
@nancycollins There are "mpi_allreduce" calls in radiance_bias_mod.f90 for a vector operation (so DART's "sum_across_tasks" won't work for it, I think).  I think if a cover routine were added to both mpi_utilities_mod.f90 and to null_mpi_utilities_mod.f90 to effectively do an mpi_allreduce for vectors, that might fix this issue about everything being MPI? 

As for the name, that's a fair point.  Since it's untested for anything besides radiances, it might be safer to keep radiances in the name for now, but I don't have a strong opinion on naming.

nancycollins:
a discussion comment:

much like externally precomputed FOs, there are implications here for the entire system that should be discussed.  (e.g. until helen fixed it, external FOs didn't work correctly if distributed FOs were not enabled in the namelist).   this is a great proof-of-concept and i'm sure a feature users want/need.  but it needs a design discussion about the bigger picture to be sure this works with all options of dart in all cases and that this is the way the group wants to implement and support this code.

jlaucar:
Craig,
This is amazing work. Thanks so much for doing it and especially for
sending it back for incorporation into the DART repository.

As you noted, it's a big set of changes. In addition, our entire team is in
the middle of a web/documentation upgrade process that is time critical and
messy. This means that the team will not be able to focus fully on the
request for a couple of weeks.

On a similar timescale, I'd like to chat with you about a variety of things
including this work and bigger picture things about your plans and MMM's
plans. That discussion will require clarification on Glen's status which I
anticipate receiving very soon.

I'm looking forward to chatting soon. Thanks again for your work, Jeff

assimilation_code/modules/assimilation/radiance_bias_mod.f90
```diff
@@ -0,0 +1,759 @@
...
+
+   integer :: i, nn, base_obs_kind, base_obs_type
+   real(digits12) :: base_rad, elapsed
+   real(r8) :: obs_value(1)
+   real(r8), allocatable :: mean_temp(:), qc_temp(:)
+
+   ! Populate mean posterior observation for ALL obs (including non-radiances) on this processor
+   call compute_copy_mean(obs_ens_handle, 1, ens_size, OBS_MEAN_COPY) ! Mean of the members sum(H(x_i))/N
+
+   call all_copies_to_all_vars(obs_ens_handle) ! get all observations onto each processor for a subset of copies
```
nancycollins, line 287: this comment is just to mark this place for future discussions.

weather4evr:
@jlaucar Thanks, Jeff.  No problem with me if this PR hangs out for awhile; it's big and I fully expected that things will have to change to make it compatible with all the facets of DART.  Look forward to chatting with you soon about this and bigger picture things.  I'm around.

hkershaw-brown:
moved this development to https://github.com/NCAR/DART/tree/radianceBiasCorrection

Test case is here: 
/glade/p/cisl/dares/hkershaw/pull_196_test_case/dares_version

PR #198, opened 2021-04-12
Put QC values and meanings into a table

kdraeder:


hkershaw-brown (approved):
much better than it was.  

If you want the table to have alternating colors (white, grey), the right side of the table needs editing:

Use | at the end of the line
```
--+
  |
--+
```

Rather than +

```
---+
   +
---+
```

Issue #192, opened 2021-03-31
Dependabots alerts for wrf hydro


hkershaw-brown:
Dependabot alerts for models/wrf_hydro/python/experiment/requirements.txt on the Manhattan branch.

Possible solution:
Have dependabot automatically issue pull requests.

I'm not sure if anyone is using Manhattan wrf_hydro so other solutions:
* dismiss the alerts as 'vulnerable code not actually used'

The alerts are only on default branch (which is Manhattan currently).  The master branch does not contain the wrf hydro requirement file.

timhoar:
I'd like to see the Manhattan branch get updated ... sooner rather than later. When that happens, the requirememnts.txt file goes away and this Issue is moot. 

Are we planning to update Manhattan (9.10.0?) or are we going for a new major release?

PR #203, opened 2021-04-20
Update  wrf hydro py  requirements.txt to address vulnerabilities 

hkershaw-brown:
Update versions in wrf hydro py requirements. 
Note, the latest version (master branch) does not have this list of requirements.

### DIFF
--- a/models/wrf_hydro/python/experiment/requirements.txt
+++ b/models/wrf_hydro/python/experiment/requirements.txt
@@ -73,10 +73,10 @@ prettytable==0.7.2
 progressbar2==3.34.3
 prompt-toolkit==1.0.15
 ptyprocess==0.5.2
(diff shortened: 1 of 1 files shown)


kdraeder:
If I understand the proposal here and in #192, this will fix the requirements file (temporarily),
instead of removing it as Tim preferred (?).  In the current documentation sprint, that seems reasonable.
Is it more secure to have no requirements file, or just more convenient for DART's github?

hkershaw-brown:
This is to stop the github alerts.   The alerts are only for the default branch. 
The new release branch (new default) will be branched of the master, which already has the requirements file removed.

kdraeder (approved):
This seems like the right strategy.  I approve the PR.

PR #213, opened 2021-05-04
AMSUA bugfix

timhoar:
This addresses Issue 208.  The AMSUA converter looped over multiple input files but only wrote out observations for the last input file. This corrects that error.

Since there are hundreds of files per day for this instrument, appending to an existing file is supported.
Because there are hundreds of files per day, a verbosity flag was added.


observations/obs_converters/AIRS/amsua_netCDF_support_mod.f90
```diff
@@ -1090,6 +1097,112 @@ subroutine fill_amsua_variables(granule, ncid, context)
...
+                   num_files / shrinkage
+
+end function max_possible_obs
+
+
+!-------------------------------------------------------------------------------
+!> Given two observation sequences, insert the observations from the first
+!> sequence into the second sequence.
+
+function combine_sequences(seq_in, seq_out, filename)
```
nancycollins, line 1134: i'm a little concerned about the performance of this approach.  as the output files get larger the cost of inserting an obs in the file grows.  the original code added new obs to the eventual output sequence.  this version creates a separate sequence (one insert here) and then combines them (another insert).  could this code pass the eventual output sequence into the fill routine and have it put the obs where they are eventually going to be?

observations/obs_converters/AIRS/convert_amsu_L1.f90
```diff
@@ -107,31 +122,81 @@ program convert_amsu_L1
...
+! It is impossible to predict how the lat/lon subsetting may affect this
+! number, but the thinning and channels can be taken into account.
+
+max_num = max_possible_obs(filecount, cross_track_thin, along_track_thin, use_channels)
+
+! either read existing obs_seq or create a new one
+
+inquire(file=outputfile, exist=file_exist)
+
+if ( file_exist .and. append_output ) then
```
nancycollins, line 141: the original main execution loop in this code was very clean.  could this code (~lines 139-157) be moved to a subroutine?  putting if statements into the main section of code obscures the high level logic.

observations/obs_converters/AIRS/convert_amsu_L1.f90
```diff
@@ -107,31 +122,81 @@ program convert_amsu_L1
...
+! release the file sequence memory in preparation for creating it new next iteration
+
 do ifile = 1,filecount
 
-   ! read from HDF file into a derived type that holds all the information
    call read_amsua_bt_netCDF_granule(l1_files(ifile), granule)   
 
-   ! convert derived type information to DART sequence
-   call make_obs_sequence(seq, granule, lon1, lon2, lat1, lat2, &
+   call make_obs_sequence(small_sequence, granule, lon1, lon2, lat1, lat2, &
```
nancycollins, line 168: if the create step is moved out of the loop and obs are added to the same sequence then you don't have to combine them as a separate step.  if there's a reason for having to do it this way there needs to be some comment before this loop saying why this choice was made.  it is going to have worse performance for large obs counts.

observations/obs_converters/AIRS/convert_amsu_L1.f90
```diff
@@ -107,31 +122,81 @@ program convert_amsu_L1
...
+! Report on the current status of the entire observation sequence
+if (verbose > 0) then
+   call print_obs_seq_summary(big_sequence)
+endif
+
+! Write the sequence to a disk file
+if ( get_num_obs(big_sequence) > 0 )  then
+   call write_obs_seq(big_sequence, outputfile) 
+else
+   call error_handler(E_ERR,'NO OBSERVATIONS TO WRITE',source)
```
nancycollins, line 193: i'm not sure i'd make this choice to error out if no obs are there.  in a big batch conversion, it's possible there are missing files, etc.  printing a msg is great but an error stops things abruptly.

nancycollins (commented):
most of the changes look great.  i do have a concern about adding more conditional code (if statements) to the original simple main logic loop.  also the process of moving obs from one sequence to another is going to cost more in execution time than adding the obs directly to the output sequence.

observations/obs_converters/AIRS/amsua_netCDF_support_mod.f90
timhoar: I understand ... mostly.  The separate obs sequence (the one for each file) is essentially guaranteed to be an append for each observation - not a searching insert. That should be about as fast as possible.  By creating a separate sequence for each file, it is possible to report on the full characteristics of THAT sequence, as opposed to the incremental change to the output sequence. Maybe that's not worthwhile ... dunno. 
The combination insert is where the inefficiency (if any) lies. If the user (and that's a big if) puts the input files in temporally ascending order, these are all appends - again - about as efficient as possible. If the files listed are out-of-order, they are their own worst enemy to begin with. 
The logic of the code got a bit confusing when I tried to use a single sequence, and had a pretty large footprint on the code.

observations/obs_converters/AIRS/convert_amsu_L1.f90
timhoar: Agreed ... I was going to do that after it passed, and have forgotten. Thanks for the reminder.

observations/obs_converters/AIRS/convert_amsu_L1.f90
timhoar: There are about 240 files per day, and using two channels from each file results in 2700 obs per file, 648000 obs per day.
With gfortran -O2 optimization on my laptop, it took less than 30 seconds to convert all 240 files in a single run - and that was using the (obs_def_rttov_mod:MAXrttovkey) initial metadata length of 100,000.

I guess I just felt the code was pretty clear by adding the extra routine and leaving the existing working code as untouched as possible.

observations/obs_converters/AIRS/convert_amsu_L1.f90
timhoar: These filenames are pretty unpredictable so the only way I know of to create them is to capture a listing - so I'm not really concerned about missing files ... now ... files with no valid observations is another thing. I haven't run into that and am not sure what the intent is.  I'm willing to let the end user modify this to a message if they want. I don't want the default behavior to be to just blow by 'empty/bad' files - particularly in a batch job where no one actually looks at the output.

observations/obs_converters/AIRS/amsua_netCDF_support_mod.f90
nancycollins: i might assume that someone who wanted to print detailed output about the input files might use a different program (obs_info, for example).  printing 240 detailed summaries in a single output file seems hard to find what you are looking for.   but i can see the logic in making separate sequences and concatinating them.  so as long as the performance is ok then this seems ok to me.

observations/obs_converters/AIRS/convert_amsu_L1.f90
nancycollins: i don't know that leaving the existing code untouched is justification for this, but i am convinced by the performance argument.   someone doing a long assimilation may be converting months (years) of these obs, but a run time of 30 seconds a day seems like it would be acceptable in a batch job.

observations/obs_converters/AIRS/convert_amsu_L1.f90
nancycollins: in batch jobs i actually always grepped the output logs because real world obs files are such a crapshoot.  i don't agree with making this an error because if i've waited in the batch queue for an hour and i'm converting 30 days of data and the first one has a problem i get nothing for my trouble.  but like you said, the user can always modify the code.  i'd let this go if you add a comment above the error line saying something like "change E_ERR to E_MSG to allow this program to exit without returning an error code."

nancycollins:
this is just for the record - not asking for any changes here.

the reason i look at efficiency in obs converters is because of history.  several of the madis obs converters starting taking upwards of 5 or 10 *minutes* to create output files.  the culprit was the large number of obs and them not being in time order in the input netcdf files.  my solution was to index sort them by time order after reading them in but before adding them to the output obs_seq.  doing that made the run time drop back down to something reasonable.   the NCEP BUFR converter also will run pretty slow if the obs counts get high.  BUFR obs are never in time order.   you have to call a BUFR read routine to get the next obs so there is never a time when you have them all available to do a time sort on them.  small output files don't have this problem, it only shows up when the obs counts get large.  you are correct that combining sequences is fast if both sequences are already sorted in time order.   if something related to creating obs files starts to get slow, look at obs times and order of insertion.  it is true that the code appends new obs to the end of the physical file so that existing obs don't have to be copied to insert a new obs.  but the linked list still has to be searched to know where to insert the new obs in the list to maintain the logical time order.  that search can get slow.

observations/obs_converters/AIRS/convert_amsu_L1.f90
timhoar: As another data point, I 'appended' the exact same 240 files to itself, which has to be a rather awful insertion pattern. The run-time (only did it once, so take it with a grain of salt) ... was 57 seconds.  Seemed to me that it took a bit of time to read the existing sequence, and write the 2X sequence ... as opposed to reading and inserting the new obs. for what its worth ...

observations/obs_converters/AIRS/amsua_netCDF_support_mod.f90
```diff
@@ -131,7 +138,8 @@ subroutine initialize_amsua_netcdf(filename)
 if ( AMSUA_BT_WARMPRTA2   /= WARMPRTA2   ) mismatch = .true.
 
 if (mismatch) then
-   call error_handler(E_ERR,'initialize_amsua_netcdf','assumptions wrong',source)
+   call error_handler(E_ERR,'initialize_amsua_netcdf','assumptions wrong',source,&
+              text2='dimensions in file do not match expected dimensions')
```
timhoar, line 142: FYI only: this particular check makes it impossible to use a netCDF file with a single pixel as test input

observations/obs_converters/AIRS/convert_amsu_L1.f90
```diff
@@ -107,31 +116,73 @@ program convert_amsu_L1
...
+! are no viable observations for a particular collection of files.
+! If you want to allow this program to exit without returning an error code, 
+! change E_ERR to E_MSG in the following call to error_handler().
+
+! Write the sequence to a disk file
+if ( get_num_obs(big_sequence) > 0 )  then
+   call write_obs_seq(big_sequence, outputfile) 
+else
+   call error_handler(E_ERR,'NO OBSERVATIONS TO WRITE',source, &
+              text2='see the DISCUSSION comment in the source to exit without error.')
```
timhoar, line 179: FYI: when the error_handler exits with an error in this case, the program status is 99.
When it exits with E_MSG, the program status is 0.

timhoar (commented):
I think it is ready for testing.
Just so people know ... I committed version 2da9c972f9  which writes out exactly 1 observation per file. I used the following to extract the same observations (all channels, however) so it is possible to confirm that appending,etc. is being done correctly. The debug version can be used in conjunction with the output from these commands:
```
#!/bin/csh
@ fnum = 1
foreach FILE ( AIRS.2019.06.22.23?.*.nc )
   ncks -O -d GeoTrack,10 -d GeoXTrack,20 $FILE output_${fnum}.nc
   ncdump -v brightness_temp output_${fnum}.nc | tail -n 5
   @ fnum ++
end
```
I have put some test datafiles in _/glade/work/thoar/git/DART_public/observations/obs_converters/AIRS/data_

observations/obs_converters/AIRS/convert_amsu_L1.f90
timhoar: Done ... the routine is called **append_or_create()** and is in the amsua_netCDF_support_mod.

observations/obs_converters/AIRS/convert_amsu_L1.f90
timhoar: I added comments and made the output string mention the comment in the code, so they know which one to edit.

observations/obs_converters/AIRS/convert_amsu_L1.f90
nancycollins: yes, that's exactly what i would expect.

nancycollins (approved):
was there a doc file for this converter?  if so, the updated namelist item needs to be added.  but i know there are several converters w/o doc files, so i'm approving this pull request even if there isn't one. (maybe we should be adding skeletal doc files saying 'documentation TBD' or something? so users know they shouldn't keep looking for a non-existent file?)

timhoar:
> was there a doc file for this converter? if so, the updated namelist item needs to be added. but i know there are several converters w/o doc files, so i'm approving this pull request even if there isn't one. (maybe we should be adding skeletal doc files saying 'documentation TBD' or something? so users know they shouldn't keep looking for a non-existent file?)

Actually, there is a _README.rst_ and an _AIRS.rst_ in the AIRS directory. Consolidating these is already an issue - it is confusing to have two. The _AIRS.rst_ is the converted contents of the AIRS.html, which does not have any information about the `convert_amsu_L1.f90` program. The _README.rst_ does not have any namelist information in it at all. Consolidating these two needs to be done - soon - and would be the right time to document the _convert_amsu_L1_ namelist. Should this be part of the PR?

hkershaw-brown:
> > was there a doc file for this converter? if so, the updated namelist item needs to be added. but i know there are several converters w/o doc files, so i'm approving this pull request even if there isn't one. (maybe we should be adding skeletal doc files saying 'documentation TBD' or something? so users know they shouldn't keep looking for a non-existent file?)
> 
> Actually, there is a _README.rst_ and an _AIRS.rst_ in the AIRS directory. Consolidating these is already an issue - it is confusing to have two. The _AIRS.rst_ is the converted contents of the AIRS.html, which does not have any information about the `convert_amsu_L1.f90` program. The _README.rst_ does not have any namelist information in it at all. Consolidating these two needs to be done - soon - and would be the right time to document the _convert_amsu_L1_ namelist. Should this be part of the PR?

I think it is a really good time to update/correct the documentation.  It seems well within the scope: fix the converter, documentation fix for people to use the converter correctly.

timhoar:
> > > was there a doc file for this converter? if so, the updated namelist item needs to be added. but i know there are several converters w/o doc files, so i'm approving this pull request even if there isn't one. (maybe we should be adding skeletal doc files saying 'documentation TBD' or something? so users know they shouldn't keep looking for a non-existent file?)
> > 
> > 
> > Actually, there is a _README.rst_ and an _AIRS.rst_ in the AIRS directory. Consolidating these is already an issue - it is confusing to have two. The _AIRS.rst_ is the converted contents of the AIRS.html, which does not have any information about the `convert_amsu_L1.f90` program. The _README.rst_ does not have any namelist information in it at all. Consolidating these two needs to be done - soon - and would be the right time to document the _convert_amsu_L1_ namelist. Should this be part of the PR?
> 
> I think it is a really good time to update/correct the documentation. It seems well within the scope of this PR: fix the converter, documentation fix for people to use the converter correctly.

OK - I'm on it ... combining AIRS.rst into  README.rst  ... please do not merge the pull request yet!

observations/obs_converters/AIRS/work/input.nml
```diff
@@ -47,10 +47,20 @@
...
+                        '../data/AIRS.2019.06.22.235.L1B.AMSU_Rad.v5.0.0.0.G19174110442.nc'
+                        '../data/AIRS.2019.06.22.236.L1B.AMSU_Rad.v5.0.0.0.G19174110442.nc'
                         '../data/AIRS.2019.06.22.237.L1B.AMSU_Rad.v5.0.0.0.G19174110446.nc'
-   l1_file_list       = ''
+                        '../data/AIRS.2019.06.22.238.L1B.AMSU_Rad.v5.0.0.0.G19174110427.nc'
+                        '../data/AIRS.2019.06.22.239.L1B.AMSU_Rad.v5.0.0.0.G19174110745.nc'
+
+&convert_amsu_L1_nml
+   l1_files           = ''
+   l1_file_list       = 'three_input_files.txt'
```
nancycollins: is this still an appropriate filename?  and is this file included or is it just an example?

nancycollins (approved):
all the docs look great.  i had one question about the namelist but it's not that big a deal, i'm just curious.  hurrah.

observations/obs_converters/AIRS/work/input.nml
```diff
@@ -47,20 +47,10 @@
...
-                        '../data/AIRS.2019.06.22.237.L1B.AMSU_Rad.v5.0.0.0.G19174110446.nc'
-                        '../data/AIRS.2019.06.22.238.L1B.AMSU_Rad.v5.0.0.0.G19174110427.nc'
-                        '../data/AIRS.2019.06.22.239.L1B.AMSU_Rad.v5.0.0.0.G19174110745.nc'
-
 &convert_amsu_L1_nml
-   l1_files           = ''
-   l1_file_list       = 'three_input_files.txt'
+   l1_files           = '../data/AIRS.2019.06.22.236.L1B.AMSU_Rad.v5.0.0.0.G19174110442.nc',
+                        '../data/AIRS.2019.06.22.237.L1B.AMSU_Rad.v5.0.0.0.G19174110446.nc'
+   l1_file_list       = ''
```
nancycollins, line 53: this looks good.

nancycollins (approved):

Issue #224, opened 2021-05-12
Feature request: support CLM on public branch


timhoar:
### Use case  
Even through v9.10.z, the CLM support on the public repository is limited to versions up to 8.y.z
The CLM support for versions 9.y.z are on a private repository and can now be made fully public.

### Is your feature request related to a problem?
There are some potential complications pertaining to the handling of the Time dimension as many use-case scenarios for CLM use history file variables (which have multiple timesteps) as part of the DART state. Additional complications may occur as different ensemble members of CLM may have different numbers of snow layers, which means we have to fully support the use of missing values in the state to keep the size and shape of the DART state vectors consistent across all members. To my knowledge, this has not been fully tested with the refactoring that occurred with v9.y.z

Furthermore, the partitioning function that takes snow water equivalent (SWE) and updates the prognostic variables (which have layers) consistent with the update to SWE have not been fully supported or tested with recent versions of CLM. Depending on the invasiveness, this may be a separate issue.

### Describe the your prefered solution  
Create a feature branch from the most current private branch, merge the public main branch onto the feature branch and develop, test, repeat. Compare the results to a known solution that we believe to be correct (from the private repo). That will be using CLM release-clm5.0.06 (circa August 2018) .  Supporting the newest CLM (release-clm5.0.34) is a subsequent issue.

### Describe any alternatives you have considered
N/A

hkershaw-brown:
fixed in #313

PR #229, opened 2021-05-24
CLM update 1

timhoar:
## Description:

First wave of changes to support CLM on main branch. Changes primarily revolve around the separate handling of `_FillValue` and `missing_value` attributes and takes pains to ensure that they are identical if both present. 

### Fixes issue
Relates to #224 - but does not fix it entirely.

### Types of changes

- [ ] Documentation update

### Tests
I ran the test_dart.csh suite on Cheyenne with the environment below and nothing broke. Additionally, I ran tests on the bgrid model, changing various attributes to trigger the various code blocks.

```
Currently Loaded Modules:
  1) ncarenv/1.3           4) mpt/2.22       7) nco/4.9.5      10) hdf-eos5/5.1.16
  2) intel/19.0.5          5) netcdf/4.7.4   8) ncview/2.1.7
  3) ncarcompilers/0.5.0   6) mkl/2019.0.5   9) diffuse/0.4.8
```

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [ ] No dataset needed

Rather than test what I tested, I'd like you to test what you think needs testing. We might test different things ...

### DIFF
--- a/assimilation_code/modules/assimilation/filter_mod.f90
+++ b/assimilation_code/modules/assimilation/filter_mod.f90
@@ -7,7 +7,7 @@ module filter_mod
 !------------------------------------------------------------------------------
 use types_mod,             only : r8, i8, missing_r8, metadatalength, MAX_NUM_DOMS, MAX_FILES
 
-use options_mod,           only : get_missing_ok_status, set_missing_ok_status
+use options_mod,           only : get_missing_ok_status
 
 use obs_sequence_mod,      only : read_obs_seq, obs_type, obs_sequence_type,                  &
                                   get_obs_from_key, set_copy_meta_data, get_copy_meta_data,   &
@@ -104,7 +104,7 @@ public :: filter_sync_keys_time, &
 character(len=*), parameter :: source = 'filter_mod.f90'
 
 ! Some convenient global storage items
-character(len=512)      :: msgstring
+character(len=512) :: msgstring, string2, string3
 
 integer :: trace_level, timestamp_level
 
@@ -255,10 +255,7 @@ real(r8) :: inf_sd_lower_bound(2)          = 0.0_r8
 
 ! Some models are allowed to have MISSING_R8 values in the DART state vector.
 ! If they are encountered, it is not necessarily a FATAL error.
-! Most of the time, if a MISSING_R8 is encountered, DART should die.
-! CLM should have allow_missing_clm = .true.
-logical  :: allow_missing_clm = .false.
-
+! The model is responsible for calling options_mod:set_missing_ok_status()
 
 namelist /filter_nml/ async,     &
    adv_ens_command,              &
@@ -308,8 +305,7 @@ namelist /filter_nml/ async,     &
    output_mean,                  &
    output_sd,                    &
    write_all_stages_at_end,      &
-   write_obs_every_cycle,        & 
-   allow_missing_clm
+   write_obs_every_cycle
 
 
 integer :: inflation_flavor(2)
@@ -396,7 +392,6 @@ call error_handler(E_MSG,'filter_main:', msgstring, source)
 ! See if smoothing is turned on
 ds = do_smoothing()
 
-call set_missing_ok_status(allow_missing_clm)
 allow_missing = get_missing_ok_status()
 
 call trace_message('Before initializing inflation')
--- a/assimilation_code/modules/io/direct_netcdf_mod.f90
+++ b/assimilation_code/modules/io/direct_netcdf_mod.f90
@@ -72,6 +72,9 @@ use utilities_mod,        only : error_handler, file_to_text, &
                                  find_textfile_dims, file_exist, &
                                  E_MSG, E_ALLMSG, E_ERR, E_DBG, E_WARN
 
+!>@todo FIXME Fully implement the rest of the routines in netcdf_utilities_mod.
+! This will change all the error messages.
+
 use netcdf_utilities_mod, only : nc_check
 
 use mpi_utilities_mod,    only : task_count, send_to, receive_from, my_task_id, &
@@ -89,6 +92,7 @@ use state_structure_mod,  only : get_num_variables, get_sum_variables,  &
                                  get_units, get_long_name, get_short_name, &
                                  get_has_missing_value, get_FillValue, &
                                  get_missing_value, get_add_offset, get_xtype, &
+                                 get_has_FillValue, &
                                  get_index_start, get_index_end , get_num_dims, &
                                  create_diagnostic_structure, &
                                  end_diagnostic_structure
@@ -820,10 +824,21 @@ end subroutine write_augmented_state
 
 
 !-------------------------------------------------------------------------------
-!> Read in variables from start_var to end_var
+!> Read in variables from start_var to end_var.
+!>
+!> If the model supports missing values (set in the model_mod), and 
+!> if the variable has a _FillValue or missing_value attribute, those values
+!> are read and matching data values are replaced with the DART 'missing' flag.
+!> If the variable specifies BOTH a _FillValue and missing_value attribute,
+!> they must be the same. DART only has one missing value and it we are unable
+!> to guarantee that we replace the DART missing value with the correct _FillValue
+!> or missing_value if the original values are different.
+!>
+!>@todo FIXME: the scale_factor and offset are unsupported.
+!>
 !>@todo FIXME: At the moment, this code is assuming that the variables in the state start
 !> at (1,1,1) and that the whole variable is read. This is not the case for
-!> TIEGCM and CLM. 
+!> TIEGCM and CLM (history files). 
 
 subroutine read_variables(ncfile_in, var_block, start_var, end_var, domain)
 
@@ -833,20 +848,27 @@ integer,  intent(in)    :: start_var
 integer,  intent(in)    :: end_var
 integer,  intent(in)    :: domain
 
+character(len=*), parameter :: routine = 'read_variables'
+
 integer :: i
 integer :: istart, iend
 integer :: var_size
 integer, allocatable :: dims(:)
 integer :: ret, var_id
+character(len=NF90_MAX_NAME) :: varname
 
-logical :: missing_possible
+logical :: allow_missing
+logical :: failure
 
-missing_possible = get_missing_ok_status()
+failure = .false.
+
+allow_missing = get_missing_ok_status()
 
 istart = 1
 
 do i = start_var, end_var
 
+   varname  = get_variable_name(domain, i)
    var_size = get_variable_size(domain, i)
    iend = istart + var_size - 1
 
@@ -856,12 +878,28 @@ do i = start_var, end_var
    dims = get_io_dim_lengths(domain, i)
 
    ret = nf90_inq_varid(ncfile_in, get_variable_name(domain, i), var_id)
-   call nc_check(ret, 'read_variables: nf90_inq_varid',trim(get_variable_name(domain,i)) )
+   call nc_check(ret, 'read_variables: nf90_inq_varid', varname)
+
+   ! check for unsupported features
+   ! In order to support them, we'd have to apply them when we write as well,
+   ! and we do not ... yet. Error out if we find them.
+   ret = nf90_inquire_attribute(ncfile_in, var_id, 'scale_factor')
+   if (ret == NF90_NOERR) failure = .true.
+
+   ret = nf90_inquire_attribute(ncfile_in, var_id, 'offset')
+   if (ret == NF90_NOERR) failure = .true.
+
+   if (failure) then
+      write(msgstring2,*)'the use of scale_factor and/or offset is not supported.'
+      call error_handler(E_ERR,routine,varname, source, text2=msgstring2)
+   endif
+
+   ! Finally read the variable.
 
    ret = nf90_get_var(ncfile_in, var_id, var_block(istart:iend), count=dims)
-   call nc_check(ret, 'read_variables: nf90_get_var',trim(get_variable_name(domain,i)) )
+   call nc_check(ret, 'read_variables: nf90_get_var', varname)
 
-   if (missing_possible) call set_dart_missing_value(var_block(istart:iend), domain, i)
+   if (allow_missing) call set_dart_missing_value(domain, i, var_block(istart:iend))
 
    istart = istart + var_size
 
@@ -916,8 +954,7 @@ COPIES: do copy = 1, state_ens_handle%my_num_copies
    iend = istart + block_size -1
 
    if (query_read_copy(name_handle, copy)) then
-      call read_variables(ncfile, vector, 1, get_num_variables(domain), domain)
-      ! close netcdf file
+      call read_variables(ncfile, vector, start_var, get_num_variables(domain), domain)
       ret = nf90_close(ncfile)
       call nc_check(ret, 'read_transpose_single_task: closing', netcdf_filename)
       state_ens_handle%copies(copy, istart:iend) = vector
@@ -1422,6 +1459,8 @@ end subroutine transpose_write_multi_task
 !> If the variable is unbounded, this routine returns .FALSE.
 !> The return value is not an indication of whether or not the values have
 !> actually been modified.
+!> "clamp_variable()" must be called after the MISSING_R8 has been replaced 
+!> by the 'native' missing value "set_model_missing_value()"
 !-------------------------------------------------------------------------------
 
 subroutine clamp_variable(dom_id, var_index, variable)
@@ -1431,8 +1470,9 @@ integer,     intent(in) :: var_index   ! variable index
 real(r8), intent(inout) :: variable(:) ! variable
 
 real(r8) :: minclamp, maxclamp, my_minmax(2)
-character(len=NF90_MAX_NAME) :: varname ! for informational log messages
-logical  :: allow_missing ! used in CLM for state variables
+character(len=NF90_MAX_NAME) :: var_name ! for informational log messages
+logical  :: allow_missing
+real(r8) :: mymissing
 
 ! if neither bound is set, return early
 minclamp = get_io_clamping_minval(dom_id, var_index)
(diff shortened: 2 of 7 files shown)


assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -7,7 +7,7 @@ module filter_mod
 !------------------------------------------------------------------------------
 use types_mod,             only : r8, i8, missing_r8, metadatalength, MAX_NUM_DOMS, MAX_FILES
 
-use options_mod,           only : get_missing_ok_status, set_missing_ok_status
+use options_mod,           only : get_missing_ok_status
```
timhoar, line 10: set_missing_ok_status() is not called by filter --  The model_mod must call it.
If the model does not call it, the default is for get_missing_ok_status to return FALSE, meaning
missing values are not allowed.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -833,20 +848,27 @@ subroutine read_variables(ncfile_in, var_block, start_var, end_var, domain)
...
+character(len=NF90_MAX_NAME) :: varname
+
+logical :: allow_missing
+logical :: failure
 
-logical :: missing_possible
+failure = .false.
 
-missing_possible = get_missing_ok_status()
+allow_missing = get_missing_ok_status()
```
timhoar, line 865: I think this variable name reads better.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -833,20 +848,27 @@ subroutine read_variables(ncfile_in, var_block, start_var, end_var, domain)
...
 
 istart = 1
 
 do i = start_var, end_var
 
+   varname  = get_variable_name(domain, i)
```
timhoar, line 871: The varname is needed multiple times, seems expedient to declare a variable for it.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -1442,60 +1482,41 @@ subroutine clamp_variable(dom_id, var_index, variable)
...
-! two lines need to be:
 allow_missing = get_missing_ok_status()
 
 if (allow_missing) then
-   my_minmax(1) = minval(variable, mask=(variable /= missing_r8))
-   my_minmax(2) = maxval(variable, mask=(variable /= missing_r8))
+   ! by the 'native' missing value ("replace_missing()").
+   call get_missing_value(dom_id, var_index, mymissing)
+   my_minmax(1) = minval(variable, mask=(variable /= mymissing))
+   my_minmax(2) = maxval(variable, mask=(variable /= mymissing))
```
timhoar, line 1491: Each variable may have its own missing value.  Since we've already checked that if it has both a `_FillValue` and `missing_value` that they are equal, we can replace the DART missing with the single known value.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -1507,23 +1528,23 @@ end subroutine clamp_variable
...
-                           do_file_clamping, force_copy)
+                           do_variable_clamping, force_copy)
 
 integer,  intent(in)    :: ncid
 real(r8), intent(inout) :: var_block(:)
 integer,  intent(in)    :: start_var
 integer,  intent(in)    :: end_var
 integer,  intent(in)    :: domain
-logical,  intent(in)    :: do_file_clamping
+logical,  intent(in)    :: do_variable_clamping
```
timhoar, line 1538: This is used in a loop around variables, not files. The direct_netcdf_mod.f90 calls this function with a variable called 'clamp_vars' ....

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -3030,17 +3064,34 @@ subroutine set_dart_missing_value(array, domain, variable)
 
 endif
 
+if ( get_has_FillValue(domain, variable) ) then 
+
+   select case ( get_xtype(domain, variable) )
+      case ( NF90_INT )
+         call get_FillValue(domain, variable, model_missing_valueINT)
+         where(array == model_missing_valueINT) array = MISSING_R8
```
timhoar, line 3072: This logic of letting the compiler do the variable type conversion mimics how the variables are read. Since we are testing for direct equality, I wanted to make sure `array` and `model_missing_xxxx` are type-converted the same.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -568,9 +567,12 @@ subroutine check_correct_variables(netcdf_filename, dom)
...
    if (ndims /= get_io_num_dims(dom,i)) then
-      write(msgstring,*) 'ndims ', get_io_num_dims(dom,i), ' in state does not', &
-                         ' match ndims ', ndims, ' in ', trim(netcdf_filename)
-      call error_handler(E_ERR, 'check_correct_variables', msgstring, source)
+      write(string1,*)'domain ',dom, ' variable ',trim(get_variable_name(dom,i))
+      write(string2,*)'ndims ', get_io_num_dims(dom,i), ' in state does not', &
+                      ' match ndims ', ndims
+      write(string3,*) ' in "'//trim(netcdf_filename)//'"'
+      call error_handler(E_ERR, 'check_correct_variables', string1, &
+                         source, text2=string2, text3=string3)
```
timhoar, line 575: With multiple domains, it is very nice to know which domain (and variable) is inconsistent.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -654,6 +657,25 @@ subroutine check_attributes(ncFile, filename, ncVarId, domid, varid)
...
+
+      case (NF90_DOUBLE)
+         call get_FillValue(domid, varid, spvalR8)
+         call check_attribute_value_r8(ncFile, filename, ncVarID, 'FillValue', spvalR8)
+
+      case default
+         call error_handler(E_ERR, 'check_attributes', 'unknown xtype', source)
+   end select
+endif
+
```
timhoar, line 678: Now that the variable can have `missing_value` or `_FillValue` we should check them separately.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -665,21 +687,25 @@ end subroutine check_attributes
...
 
-subroutine check_attribute_value_int(ncFile, filename, ncVarID, att_string, spvalINT)
+subroutine check_attribute_value_int(ncFile, filename, ncVarID, att_string, att_valINT)
 
 integer,          intent(in) :: ncFile
 character(len=*), intent(in) :: filename
 integer,          intent(in) :: ncVarID
 character(len=*), intent(in) :: att_string
-integer,          intent(in) :: spvalINT
+integer,          intent(in) :: att_valINT
```
timhoar, line 696: Its an attribute value ... 'spval' ... generally means Special Value ( i.e. missing_value or ...)

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -665,21 +687,25 @@ end subroutine check_attributes
...
-      write(msgstring,*) ' variable attribute, ', trim(att_string), ' in state', spvalINT, &
-                         ' does not match ', trim(att_string), ' ', ret_spvalINT, ' in ', trim(filename)
-      call error_handler(E_ERR, 'check_attributes', msgstring, source)
+if ( nf90_get_att(ncFile, ncVarID, att_string, ret_att_valINT) == NF90_NOERR ) then
+   if (att_valINT /= ret_att_valINT) then
+      write(string1,*) ' variable attribute "'//trim(att_string)//'" in state is ', att_valINT
+      write(string2,*) ' does not match "'//trim(att_string)//'"', ret_att_valINT
+      write(string3,*) ' in "'//trim(filename)//'"'
+      call error_handler(E_ERR, 'check_attributes', string1, &
+                         source, text2=string2, text3=string3)
```
timhoar, line 708: I like putting double-quotes around strings that - when things go wrong - can frequently be empty.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -1141,111 +1187,12 @@ function get_copy_name(file_handle, cnum)
...
-!#!
-!#! end function nc_get_ncid
-
-
 !------------------------------------------------------------------
 !> return whether the init routines have been called?
 !> or whether it's been marked as using a single file?
 !>@todo fixme: document what this routine does.
 
+
```
timhoar, line 1195: The rest of the file is formatted with two blank lines before a routine declaration.

assimilation_code/modules/io/state_vector_io_mod.f90
```diff
@@ -37,8 +37,7 @@ module state_vector_io_mod
                                  read_single_file, write_augmented_state, &
                                  initialize_single_file_io, finalize_single_file_io
 
-
-use types_mod,            only : r8, i4, i8, MISSING_R8
+use types_mod,            only : r8, i4, i8
```
timhoar, line 40: missing_r8 is simply not used.

timhoar (commented):
My suggested review sequence  is 

1. state_vector_io_mod.f90
2. location_mod.f90
3. filter_mod.f90
4. sampling_error_correction_mod.f90
5. state_structure_mod.f90
6. io_filenames_mod.f90
7. direct_netcdf_mod.f90

timhoar:
Just realized I forgot to check (lately) with r8=r4 ... I feel like I did that a while ago, but I should confirm.

assimilation_code/modules/io/direct_netcdf_mod.f90
nancycollins: i think this code is doing more work than is needed and is confusing.  we have already read the netcdf data into an R8 (or R4) array and that has done whatever conversion/roundoff might be needed (e.g. int -> real_r8 or float (real_r4) -> real_r8).  why not read the fill value directly into an R8 variable which will do the same conversion as the array.  then you don't need the different variable types, nor the case statement and you can do a single where() statement.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -3058,6 +3109,22 @@ subroutine set_model_missing_value(array, domain, variable)
          where(array == MISSING_R8) array = model_missing_valueR4
       case ( NF90_DOUBLE )
          call get_missing_value(domain, variable, model_missing_valueR8)
+         where(array == MISSING_R8)       array = model_missing_valueR8
+   end select
+
+endif
+
+if ( get_has_FillValue(domain, variable) ) then
```
nancycollins, line 3117: same comment here.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -557,9 +556,9 @@ subroutine check_correct_variables(netcdf_filename, dom)
 
    ! get variable id from necfile
```
nancycollins, line 557: typo -
```suggestion
   ! get variable id from ncfile
```

assimilation_code/modules/io/io_filenames_mod.f90
nancycollins: ok, here's where i'd like some explanation of how we are handing having both and if they aren't the same value.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -893,7 +926,20 @@ subroutine file_info_dump(file_info,context)
          write(*,'(A, I4,A,I4)') 'file_info%stage_metadata%io_flag(         ',i,'     ) ', &
                                 file_info%stage_metadata%io_flag(             i)
          write(*,'(A, I4,2A)')   'file_info%stage_metadata%copy_name(       ',i,'     ) ', &
-                                  file_info%stage_metadata%copy_name(i)
+                                  trim(file_info%stage_metadata%copy_name(    i))
+         write(*,'(A, I4,2A)')   'file_info%stage_metadata%long_name(       ',i,'   ) ', &
+                                  trim(file_info%stage_metadata%long_name(    i))
+
+         write(*,'(A, I4,A,L)')   'file_info%stage_metadata%clamp_vars(       ',i,'  ) ', &
```
nancycollins, line 933: i haven't seen L in a format.  is it supported on all compilers we can test with?

assimilation_code/modules/io/state_structure_mod.f90
```diff
@@ -765,114 +769,157 @@ subroutine load_common_cf_conventions(domain)
...
 
    var_xtype = domain%variable(ivar)%io_info%xtype
    select case (var_xtype)
       case ( NF90_INT )
-          if (nf90_get_att(ncid, NF90_GLOBAL, '_FillValue', cf_spvalINT) == NF90_NOERR) then
+          ret = nf90_get_att(ncid, NF90_GLOBAL, '_FillValue',    cf_spvalINT)
+          if (ret == NF90_NOERR) then
              domain%variable(ivar)%io_info%spvalINT       = cf_spvalINT
-             domain%variable(ivar)%io_info%has_missing_value = .true.
+             domain%variable(ivar)%io_info%spvalR8             = real(cf_spvalINT,r8)
```
nancycollins, line 806: why is there a spvalR8 and a missing_value in the io_info struct?  when will they be different?  and again - i don't see why the code needs to have per-type missing values since the only use will be when they're converted to R8.

nancycollins (changes requested):
we need some good test cases that have netcdf variables of int, r4 and r8; have no missing values; have missing values, fill values, both;  and have the test code compiled with R4=R8.   i am not convinced that storing INT, R4 and R8 values and all the code to read/write/access it is needed and it just gets worse trying to save fill values as well.

nancycollins:
i see you said you ran some test cases by hand.  could they be set up so you could commit them to the developer's test directory?  you can't automate tests if they need to be run interactively with changes between cases.

timhoar:
I have CLM datasets in the 'standard' test tarball, they have variables of several types with missing_values, FillValues, etc.
I don't believe I have integer valued variables with those attributes.
I agree - I don't know FOR SURE ... that we need separate variables for the different type 'special' values. I have run several tests a long time ago that led me to believe we did. I am in the process of reviving those tests. However, if someone has already definitively shown we do not need them, I'd like to know.  Since much of the CLM 'missing' handling is done by testing for an equality ... I am quite concerned by any potential representation nuances.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -856,12 +878,28 @@ subroutine read_variables(ncfile_in, var_block, start_var, end_var, domain)
...
+
+   ret = nf90_inquire_attribute(ncfile_in, var_id, 'offset')
+   if (ret == NF90_NOERR) failure = .true.
+
+   if (failure) then
+      write(msgstring2,*)'the use of scale_factor and/or offset is not supported.'
+      call error_handler(E_ERR,routine,varname, source, text2=msgstring2)
+   endif
+
+   ! Finally read the variable.
```
timhoar, line 897: By not providing the 'start' optional variable to nf90_get_var, it defaults to an array of ones ... so it is always starting at timestep 1, for example ... and using the dimensions (dims) from 'get_io_dim_lengths()' ...

nancycollins: sorry, i don't understand this comment.  is it for pointing out where future work might need to happen? the nf90_get_var() call on line 899 hasn't changed from the original.

assimilation_code/modules/assimilation/filter_mod.f90
timhoar: You are right - removing `allow_missing_clm` is a non-backward compatible change. I searched the entire 'main' branch, and it is the only namelist it appears in is the example filter_mod.nml ... not in a single input.nml.  I understand that there may be one out there in the wild ...

assimilation_code/modules/io/direct_netcdf_mod.f90
nancycollins: this was originally 'allow_missing' in much of the code.  but i remember changing the name because 'allow_missing' sounds like a verb and that you might be able to set it to change the behavior.  'missing_possible' sounded more like the result, not the action.  not a bit deal, but i'm not unhappy with the original name here ('missing_possible').

nancycollins (commented):
this is just a comment.  i'm confused about adding random comments and doing another review on github.  someday i'll understand github, but it's not today.

assimilation_code/modules/assimilation/filter_mod.f90
timhoar: What is not part of this pull request is the change to the CLM input.nml and model_mod.f90.  The proposed model_mod.f90 DOES call the set_missing_ok_status. This PR is intended to parcel out the changes in smaller (easier to understand and test) portions.   By _**not**_ calling set_missing_ok_status, the options_mod has the traditional behavior in that missing values are not allowed.

assimilation_code/modules/io/io_filenames_mod.f90
timhoar: The variable can have one or the other, it doesn't have to have both.

assimilation_code/modules/assimilation/filter_mod.f90
nancycollins: i'm fine with the change, and i'm very fine with parceling out the CLM changes in bite-sized chunks.  i'm glad to know our repo doesn't have this item in any input.nmls.  it's unlikely, but users may have older input.nmls that include allow_missing_clm in their filter namelist, so my request here is just to be sure our changelog and release notes include something about this item being removed from the filter namelist.  after updating if they get an error starting filter they'll have an idea of what's wrong.  seems like a pretty low risk change and it's moving functionality into a more appropriate place.

assimilation_code/modules/io/io_filenames_mod.f90
nancycollins: sure.  technically these 2 attributes mean different things.  missing value is something that's written to say there is no value here.  fill value is supposed to be used for data that hasn't been written yet - not yet filled in.  but people use them interchangeably so it's confusing.  i know they don't have to have both.  i'm asking what happens if they do?  if they're the same value?  if they're different values?  what combination of user choices are we intending to support with this code?

assimilation_code/modules/io/direct_netcdf_mod.f90
timhoar: CLM has both NF90_DOUBLE and NF90_FLOAT  variables, both with attributes for missing_value and _FillValue.
If the variable is NF90_DOUBLE, the fill value reported by ncdump is 'missing_value = 1.e+36 ;`
if the variable is NF90_FLOAT, the fill value is `missing_value = 1.e+36f ;`
so there is some information that the attribute has a matching netCDF 'xtype' if you will. 

I have not convinced myself that the conversion of the attribute value is the same as the conversion of the variable values. The variable values have a netCDF 'xtype', it is not known to me if the attribute values are of the same netCDF 'xtype' and are converted to r8 (or r4) identically. Do you know?

When we go to replace the DART missing value in the state with the (one) CLM special value (we've checked that if both are specified that they are identical), I want to make sure we put back EXACTLY what we found. The existing netCDF files may have NF90_DOUBLE variables and we might be running in reduced precision ... so we have to convert to the precision of the existing output variable,
using that exact missing value. So - I was keeping both in the state structure, even if we run with r8=r4.

timhoar:
state_structure_mod.f90   line 889    ensures that if both are supplied,
they are identical.

timhoar:
The test dataset for a bgrid  model_mod_check is    `/glade/work/thoar/git/DART_public/models/bgrid_solo/work/perfect_input.cdl`
You can generate your own netCDF file and run model_mod_check ... be aware that the default namelist settings for this do not use `perfect_input.nc` ... so you have to change it in the model_nml and the model_mod_check_nml ...

timhoar (commented):
> i see you said you ran some test cases by hand. could they be set up so you could commit them to the developer's test directory? you can't automate tests if they need to be run interactively with changes between cases.

I mucked with the bgrid perfect_input.cdl ... perhaps one of us could do something similar with  the L96 cdl or ???
I've got other fish to fry right now ...

> The test dataset for a bgrid model_mod_check is /glade/work/thoar/git/DART_public/models/bgrid_solo/work/perfect_input.cdl
> You can generate your own netCDF file and run model_mod_check ... be aware that the default namelist settings for this do not use perfect_input.nc ... so you have to change it in the model_nml and the model_mod_check_nml ...

hkershaw-brown:
My understanding from the couple of meetings we have had about this is that the missing value and fill value code is not a prerequisite for bringing CLM (as used by Brett, Andy, Xueli).  I propose we table this pull request for now and concentrate on code that is required to bring in CLM.

nancycollins (commented):
let's get this code onto clm_main and move on to the converters and model_mod.

nancycollins (approved):

hkershaw-brown (approved):

braczka (approved):

PR #241, opened 2021-06-08
Add latest local PF updates

jpoterjoy:
## Description:

All code changes reflect major updates to the local PF (filter option 9). Most changes are self-contained in assim_tools_mod.pf.f90, with a subroutine has been added to mpi_utilities_mod.f90 and null_mpi_utilities_mod.f90.

### Types of changes

- [ ] Documentation update

### Tests

Changes were tested for applications of the Lorenz 96, B-grid solo, and WRF models.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [x] No dataset needed

### DIFF
--- a/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
@@ -1914,7 +1914,7 @@ if (filter_kind == 9 .and. pf_kddm > 0 ) then
 
 endif ! KDDM
 
-! Perform RTPS step for hybrid
+! Perform RTPS step for hybrid. For hybrid PF-EAKF the last iteration switches to EAKF (filter_kind=1)
 if (filter_kind == 1 .and. .not. local_varying_ss_inflate) then
 
    if (my_task_id() == 0) then
--- a/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
@@ -1721,6 +1721,12 @@ ITERATIONS: do iter = 1,maxiter
    OBS_UPDATE: do j = 1, num_close_obs
       obs_index = close_obs_ind(j)
 
+      ! The ocal PF now iterates multiple times over observations, so it still needs 
+      ! to update obs-space priors for measurements that are already assimilated.
+      ! For other filter_kinds, this line should be put back in to cycle for obs which 
+      ! have already been processed. 
+      ! if (my_obs_indx(obs_index) <= i) cycle OBS_UPDATE
+
       ! Skip if regularization reaches threashold value 
       if (filter_kind == 9) then
          if ( beta_y(obs_index) == beta_max ) cycle OBS_UPDATE
@@ -1742,9 +1748,8 @@ ITERATIONS: do iter = 1,maxiter
          impact_factor = obs_impact_table(base_obs_type, my_obs_kind(obs_index))
          cov_factor = cov_factor * impact_factor
       endif
-
-      ! 
-      if (filter_kind == 9) then ! note other filter kinds will need to cycle
+ 
+      if (filter_kind == 9) then ! note other filter kinds will also need to cycle
          if ( cov_factor == 0.0_r8 ) cycle OBS_UPDATE
       end if
 
--- a/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
@@ -1253,10 +1253,6 @@ ITERATIONS: do iter = 1,maxiter
       endif
       whichvert_obs_in_localization_coord = nint(whichvert_real)
 
-      if (is_doing_vertical_conversion) then
-         ! use converted vertical coordinate value and type from owner
-         call set_vertical(base_obs_loc, vertvalue_obs_in_localization_coord, whichvert_obs_in_localization_coord)
-      endif
    endif
    !-----------------------------------------------------------------------
 
@@ -1268,6 +1264,12 @@ ITERATIONS: do iter = 1,maxiter
       endif
       cycle SEQUENTIAL_OBS
    endif
+   
+   !> all tasks must set the converted vertical values into the 'base' version of this loc
+   !> because that's what we pass into the get_close_xxx() routines below.
+   if (is_doing_vertical_conversion) &
+      call set_vertical(base_obs_loc, vertvalue_obs_in_localization_coord, whichvert_obs_in_localization_coord)
+   
    ! Skip for PF
    if (filter_kind /= 9) then
 
--- a/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
@@ -1721,7 +1721,7 @@ ITERATIONS: do iter = 1,maxiter
    OBS_UPDATE: do j = 1, num_close_obs
       obs_index = close_obs_ind(j)
 
-      ! The ocal PF now iterates multiple times over observations, so it still needs 
+      ! The local PF now iterates multiple times over observations, so it still needs 
       ! to update obs-space priors for measurements that are already assimilated.
       ! For other filter_kinds, this line should be put back in to cycle for obs which 
       ! have already been processed. 
--- a/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
@@ -1511,11 +1511,6 @@ ITERATIONS: do iter = 1,maxiter
       cov_factor = comp_cov_factor(close_state_dist(j), cutoff_rev, &
          base_obs_loc, base_obs_type, my_state_loc(state_index), my_state_kind(state_index))
 
-      ! SEEMS TO BE NEEDED FOR VERTICAL LOCALIZATION IN BGRID
-      if (filter_kind == 9) then
-         if ( cov_factor == 0.0_r8 ) cycle STATE_UPDATE
-      end if
-
       ! if external impact factors supplied, factor them in here
       ! FIXME: this would execute faster for 0.0 impact factors if
       ! we check for that before calling comp_cov_factor.  but it makes
@@ -1524,6 +1519,11 @@ ITERATIONS: do iter = 1,maxiter
          impact_factor = obs_impact_table(base_obs_type, my_state_kind(state_index))
          cov_factor = cov_factor * impact_factor
       endif
+
+      ! If no weight is indicated, no more to do with this state variable
+      if (filter_kind == 9) then ! note other filter kinds will need to cycle
+         if ( cov_factor == 0.0_r8 ) cycle STATE_UPDATE
+      end if
 
       ! Update state for PF
       if (filter_kind == 9) then
--- a/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
@@ -1,19 +1,34 @@
 ! DART software - Copyright UCAR. This open source software is provided
 ! by UCAR, "as is", without charge, subject to all terms of use at
 ! http://www.image.ucar.edu/DAReS/DART/DART_download
-
+!
+! $Id: assim_tools_mod.pf.f90 12591 2018-05-21 20:49:26Z nancy@ucar.edu $
+!
+!
+!!!!!!!!!!!!!!!!!!!!!!  PLEASE READ BEFORE USING  !!!!!!!!!!!!!!!!!!!!!!
+!
+! This version of assim_tools_mod supports ongoing developments toward a
+! localized particle filter (Poterjoy 2016; Poterjoy and Anderson 2016;
+! Poterjoy et al. 2019; Poterjoy 2021). It should NOT be used with 'filter 
+! kind' other than 9 or with any sort of adaptive prior/posterior inflation.
+! A breakdown of namelist options specific to the local PF are described
+! below. They include options to use a hybrid PF-EAKF update, which performs
+! a partial assimilation of observations with the PF followed by a second
+! update using the EAKF (Poterjoy 2021). Please contact the developer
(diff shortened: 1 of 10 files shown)


jpoterjoy:
For testing with Lorenz 96, you can omit all PF-specific namelist entries and run with default options. For localization and inflation: use the same cutoff as EAKF and no prior or posterior inflation.

hkershaw-brown:
A note to people reviewing this request:
The PF filter is not being added to the documentation yet - so public but not publicized. 
At a later date the PF filter will be added to the dart documentation so it can be used by a wider audience.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
```diff
@@ -1,19 +1,34 @@
 ! DART software - Copyright UCAR. This open source software is provided
 ! by UCAR, "as is", without charge, subject to all terms of use at
 ! http://www.image.ucar.edu/DAReS/DART/DART_download
-
+!
+! $Id: assim_tools_mod.pf.f90 12591 2018-05-21 20:49:26Z nancy@ucar.edu $
```
hkershaw-brown, line 5:
```suggestion

```

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
```diff
@@ -108,7 +122,7 @@ module assim_tools_mod
 ! and fill this 2d impact table.
 real(r8), allocatable  :: obs_impact_table(:,:)
 
-character(len=*), parameter :: source = 'assim_tools_mod.pf.f90'
+character(len=*), parameter :: source = 'assim_tools_mod.f90'
```
hkershaw-brown, line 125:
```suggestion
character(len=*), parameter :: source = 'assim_tools_mod.pf.f90'
```

assimilation_code/modules/utilities/null_mpi_utilities_mod.f90
```diff
@@ -637,6 +637,27 @@ end subroutine get_from_fwd
...
+!-----------------------------------------------------------------------------
+
+!-----------------------------------------------------------------------------
+
+!> Collect global max values on each task.
+
+subroutine get_global_max(max)
+
+real(r8), intent(inout)  :: max       !> global max over tasks
+integer :: errcode
```
hkershaw-brown, line 650:
```suggestion

```

unused variable

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
```diff
@@ -1014,6 +1253,10 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
       endif
       whichvert_obs_in_localization_coord = nint(whichvert_real)
 
+      if (is_doing_vertical_conversion) then
+         ! use converted vertical coordinate value and type from owner
+         call set_vertical(base_obs_loc, vertvalue_obs_in_localization_coord, whichvert_obs_in_localization_coord)
+      endif
```
hkershaw-brown, line 1259: This is inside the SEQUENTIAL_OBS do loop - I'm not sure it should be.
Should it be outside since looking at assm_tools_mod.f90 :
```
   !> all tasks must set the converted vertical values into the 'base' version of this loc
   !> because that's what we pass into the get_close_xxx() routines below.
```

nancycollins: it does need to be inside the SEQUENTIAL_OBS do loop - it was before.  but it looks to me like these lines have moved to inside the 'not owner of this obs' section, instead of below the endif at line 1260 where the 'all tasks do this' section starts. (see comment at line 1263).  it looks like it should move down to after line 1270.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
```diff
@@ -1244,18 +1497,23 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
...
          varying_ss_inflate_sd = 0.0_r8
       endif
-
-      ! Compute the distance and covariance factor
+     
+      ! Compute the distance and covariance factor 
       cov_factor = comp_cov_factor(close_state_dist(j), cutoff_rev, &
          base_obs_loc, base_obs_type, my_state_loc(state_index), my_state_kind(state_index))
 
+      ! SEEMS TO BE NEEDED FOR VERTICAL LOCALIZATION IN BGRID
```
nancycollins, line 1512: i don't understand this comment.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
```diff
@@ -1244,18 +1497,23 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
...
+      if (filter_kind == 9) then
+         if ( cov_factor == 0.0_r8 ) cycle STATE_UPDATE
```
nancycollins, line 1514: this code has moved and is now inside a PF-filter only if-test but all filters need to cycle if the covariance value is 0.  also the test should be after the impact_factor multiplication.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
```diff
@@ -1447,109 +1719,131 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
...
-            cov_factor = cov_factor * impact_factor
-         endif
+      ! Compute the distance and the covar_factor
+      cov_factor = comp_cov_factor(close_obs_dist(j), cutoff_rev, &
+         base_obs_loc, base_obs_type, my_obs_loc(obs_index), my_obs_kind(obs_index))
 
-         if(cov_factor <= 0.0_r8) cycle OBS_UPDATE
+      ! SEEMS TO BE NEEDED FOR VERTICAL LOCALIZATION IN BGRID
+      if (filter_kind == 9) then
+         if ( cov_factor == 0.0_r8 ) cycle OBS_UPDATE
```
nancycollins, line 1737: same comment - filter_kinds other than 9 need to cycle if cov_factor is 0, also it should be moved so impact_factor is applied first before testing.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
```diff
@@ -1447,109 +1719,131 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
    OBS_UPDATE: do j = 1, num_close_obs
       obs_index = close_obs_ind(j)
 
-      if (my_obs_indx(obs_index) <= i) cycle OBS_UPDATE
```
nancycollins, line 1450: this line causes the loop to only process and update expected obs values which have not yet been assimilated.  it's missing from the updated code.  if the PF needs to loop over all obs every time then it needs an if (filter_kind /= 9) and this test to cycle for obs which have already been processed.  if it doesn't need to do anything with obs already assimilated, it's doing a lot more work than it needs to.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
```diff
@@ -1604,7 +1910,90 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
       call error_handler(E_MSG,'',msgstring)
    endif
 
-endif
+endif ! KDDM
+
+! Perform RTPS step for hybrid
```
nancycollins, line 1915: this new RTPS section is for filter_kind 1, not the PF - is it yet a different experimental part of the code?

hkershaw-brown: the code is a hybrid, so for the last iteration it switches to an EAKF 
```
1987 filter_kind = 1
```

nancycollins: but if filter_kind starts out as 1 it also goes into this code.  i did see the comment at the top of this file saying only run this version with filter_kind == 9 so i understand this isn't the integrated version.  for now it might be worth adding a line in the code after the comment at 1915 about the last iteration switching to EAKF.

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -1977,6 +1977,28 @@ end subroutine get_from_fwd
 !-----------------------------------------------------------------------------
 !-----------------------------------------------------------------------------
 
+!-----------------------------------------------------------------------------
+
+!> Collect global max values on each task.
+
+subroutine get_global_max(max)
```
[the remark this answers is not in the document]
nancycollins: you know, i'm rethinking the name suggestion here for two reasons. 

first, get_global_max() is a good name for this routine if you are reading the code that calls it, which is the point of giving them names different from "mpi_all_reduce()".  

second, broadcast_minmax() is really a different kind of routine and making the names the same hides the difference.  broadcast_minmax() takes in 2 arrays and returns the global min/max for each element in the arrays.  it should probably be renamed to get_global_minmax_per_element() or something.  there's also a broadcast_flag() routine that sends a flag from one task to all tasks.  using the same name for these diverse routines hides the details in usage that differ pretty significantly.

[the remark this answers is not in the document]
hkershaw-brown: I agree, I think 'get_global_max' is a good descriptive name.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
hkershaw-brown: yeah I meant in the 'not owner of this obs' section not SEQUENTIAL_OBS.  I'll move the lines to 1270. 
I wonder if the `! SEEMS TO BE NEEDED FOR VERTICAL LOCALIZATION IN BGRID` comment is related to not all the tasks setting the vertical values.

nancycollins: the option to do all the vertical conversions up front for either obs, state or both before going into the sequential loop, vs just calling get_close_xx() and having that code convert locations "on demand" makes the flow a bit confusing. 

that option is useful for performance if the obs are only going to impact a small part of the state, or if a lot of the obs aren't going to be assimilated because of namelist choices or failed forward operators.

here, i think set_vertical() only needs to be done if the owning task has already converted the vertical, which is usually true for obs so there's no extra if() test added.  but it does make it look like obs are always already converted which isn't true depending on the namelist settings.  the get_close_xx() routines still need to be able to convert and update the locations if the vertical isn't already in the right coordinates.

in retrospect it would have been simpler to require all the locations to be converted at the start and then the get_close routines wouldn't have to have optional code for doing conversions on demand.  not sure the possible better performance for some unusual cases is worth the complexity.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
nancycollins: since this version is only going to be run with filter_kind == 9, this could stay as-is but maybe a comment should be added that other filter kinds will also need to cycle.  this test should still move down below the impact_factor multiply code so the impact code can be tested and used with this filter_kind and work correctly.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
nancycollins: this one is a science question.  i couldn't tell from the code if this loop had to cycle over all obs each time or not.  in the other filter kinds that's wasted work to update forward operator values for obs that have already been assimilated and won't be used again.  if this test was removed on purpose then a comment about adding the cycle code back in if the filter_kinds are integrated might be a good idea.

assimilation_code/modules/utilities/null_mpi_utilities_mod.f90
```diff
@@ -637,6 +637,26 @@ end subroutine get_from_fwd
...
+!> Collect global max values on each task.
+
+subroutine get_global_max(max)
+
+real(r8), intent(inout)  :: max       !> global max over tasks
+
+if ( .not. module_initialized ) then
+   write(errstring, *) 'initialize_mpi_utilities() must be called first'
+   call error_handler(E_ERR,'get_global_max', errstring, source)
+endif
```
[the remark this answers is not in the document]
hkershaw-brown: removed

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
jpoterjoy: Because the local PF now iterates multiple times over observations, it still needs to update obs-space priors for measurements that are already assimilated. We can, however, add this line back in for the last iteration.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
jpoterjoy: This part can definitely be cleaned up. I am currently using RTPS on the EAKF portion of the hybrid PF-EAKF update, but have it working on the EAKF alone for testing purposes. I advise users to not use this module for filter_kind other than 9 and with inflation turned off.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
hkershaw-brown: moved the `set vertical` to match assim_tools_mod.f90

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
hkershaw-brown: removed comment

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
hkershaw-brown: added comment about hybrid using filter_kind=1 for last iteration.

assimilation_code/modules/assimilation/assim_tools_mod.pf.f90
hkershaw-brown: added a note about this line

hkershaw-brown (approved):

PR #250, opened 2021-06-25
Clm update 3

timhoar:
## Description:
Update scripting to provide capability to test end-to-end.

- _simple.csh_  provides ability to run a single instance of CLM without any pieces from DART
- _CLM5_startup_freerun_  uses the multi-instance capability of CESM and the DS345.0 forcing from a single CLM 'startup' dataset.
- _CLM5_hybrid_freerun_ uses the multi-instance capability of CESM and the DS345.0 forcing from an existing ensemble.
- _CLM5_setup_pmo_ uses the DS345.0 forcing and a single existing instance of CLM
- _CLM5_setup_assimilation_ performs an assimilation.

### Fixes issue
#224 

### Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [X] Documentation update

- [X] My change requires a change to the documentation.
  - [X] I have updated the documentation accordingly.

### Tests
I ran experiments with each of the scripts.
They are configured to run on cheyenne and use datasets in public directories.

## Checklist for merging

- [ ] Updated changelog entry
- [X] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [X] No dataset needed
Although the observations for testing are all over the place.


braczka:
The clm_update_3 looks really good and testing scripts right now.   More of a comment for clm_update_2:   when performing diagnostic matlab scripts in ~/DART/models/clm/matlab, either 'clm_get_var.m' or 'clm_plot_var.m' I do not have access to function libraries to execute 'nc_isvar' (for example).   These libraries could be deprecated -- not sure, but not part of main DART matlab libraries as far as I could tell.

models/clm/shell_scripts/cesm2_2/assimilate.csh
```diff
@@ -0,0 +1,387 @@
+#!/bin/csh
+#
+# DART software - Copyright UCAR. This open source software is provided
+# by UCAR, "as is", without charge, subject to all terms of use at
+# http://www.image.ucar.edu/DAReS/DART/DART_download
+#
+# This script performs an assimilation by directly reading and writing to
+# the CLM restart file. There is no post-processing step 'dart_to_clm',
+# consequently, snow DA is not supported in this framework.
+
```
braczka, line 10: This comment is a bit confusing, because the main purpose of this update is to support snow DA.  I suppose it means if this script is run alone -- without support of clm_to_dart and dart_to_clm etc. snow DA is not possible?

models/clm/shell_scripts/cesm2_2/simple.csh
```diff
@@ -0,0 +1,114 @@
+#!/bin/csh
+#
+# DART software - Copyright UCAR. This open source software is provided
+# by UCAR, "as is", without charge, subject to all terms of use at
+# http://www.image.ucar.edu/DAReS/DART/DART_download
+
+# This file is just to see if CESM can be bullt on whatever architecture.
+# Single instance ... nothing fancy ... no DART ... just something simple.
+
+/glade/work/thoar/CESM/ctsm/cime/scripts/query_config  --compsets clm
```
nancycollins, line 10: i'm confused about the intent for this script.  if for a user, is it worth trying to use $USER or setting some HOMEDIR/WORKDIR variables?  this has "thoar" in several pathnames and TJH prints other places so it sounds like your own script, but it's in the pull request.  i guess my request here is to either 1) add something to the comments saying this would need to be updated to be used by someone else or 2) make it a bit more portable if that's feasible.

nancycollins (approved):
i'm leaving the detailed scripting review to brett.  i just skimmed them but assume they will be run by brett and updated where needed. 

i did have a couple optional comment on the "simple.csh" script - since it's part of the pull request and seems useful, maybe just explain in the comments what you need in order to run it - change pathnames, check out version XX of CLM, have datasets ??? for testing, etc.

timhoar:
I am still tweaking the perfect_model.csh and trying to document or encourage people to explore the DATA_ASSIMILATION_CYCLES mechanism and its interplay with the short-term archiver. The support for SourceMods is finished - the SourceMods themselves are being checked now.

models/clm/shell_scripts/cesm2_2/CLM5_hybrid_freerun
```diff
@@ -0,0 +1,532 @@
...
+if ( ${use_SourceMods} == TRUE ) then
+
+   if (    -d    ${SourceModDir} ) then
+      ${COPY} -r ${SourceModDir}/* ${caseroot}/SourceMods/ || exit 7
+
+      if ( -e    SourceMods/src.clm ) then
+         cd      SourceMods/src.clm
+         ${LINK} `find . -name "*F90"` . || exit 7
+         cd      ../..
+   else
```
braczka, line 491:
```suggestion
        endif
   else
```

assimilation_code/modules/observations/default_quantities_mod.f90
```diff
@@ -297,6 +297,7 @@
 !   QTY_SOIL_NITROGEN
 !   QTY_SOIL_RESPIRATION_FLUX
 !   QTY_SOIL_TEMPERATURE
+!   QTY_SOIL_MINERAL_NITROGEN
```
timhoar, line 300: See Issue #255

models/clm/shell_scripts/cesm2_2/assimilate.csh
timhoar: dart_to_clm  is the logical choice to implement the snow_update() mechanism, but it also replaces the CLM indeterminate values in the unused snow layers (and maybe elsewhere).  So right now - updating the prognostic variables based on increments to SWE is not possible. 
Updating the prognostic snow variables based on their ensemble correlation with (any) close observation is what happens currently (and may be the source of the 'soil balance errors').

models/clm/shell_scripts/cesm2_2/simple.csh
timhoar: I could definitely put in a comment indicating the the 'query_config' command has to come from their CESM clone.
I'll get rid of the TJH bits as well.

braczka:
Continuing to troubleshoot `CLM5_setup_assimilation` script and the error messages received when running this assimilation, similar to issue #254.   This script will run successfully (5 days) with most clm variables (20 out of 23) as defined in `&model_nml` **_except_** those related to snow layers including `frac_sno`, `SNOW_DEPTH`, and `H2OSOI_ICE`.  Does not matter if they are `updated` or `NO_COPY_BACK`. If any of these variables are included in &model_nml I receive a NaN error within the CESM coupler similar to:

` # of NaNs =            1`
`11: Which are NaNs =  F F F F F F F F F F F F T F F F F F F F F F F F F F F F F F F`
`11: F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F F`
`11: Sl_tsrf01` can also be `SI_t` `Fall_lat` `Fall_sen` `Fall_lwup` `Fall_evap`
`11: gridcell index =         3475`
`11: ENDRUN:`
`11: ERROR:`
`11: lnd_export                       ERROR: One or more of the output from CLM to the coupler are NaN`

I also tested `CLM5_setup_assimilation` with the `ignore_snow_update` branch which is intended to skip the posterior updates to the snow layer variables (`dart_to_clm`).  Doing this replicates the same error as reported in #254 as follows:

`37: ERROR: ERROR in UrbanRadiationMod.F90 at line 674`
`10: urban net longwave radiation error: no convergence`
`10: clm model is stopping`

Depending upon the clm variables included in the assimilation (e.g. `T_SOISNO`) sometimes the `ignore_snow_update` will produce this urban radiation error whereas the `clm_update_3` branch will not.  Including the `ignore_snow_update`  could be contributing to the error.  

I plan to examine the outgoing restart file created by dart_to_clm for `frac_sno`, `SNOW_DEPTH`, and `H2OSOI_ICE` more closely.   Prior checks indicated this step was behaving normally within DART.   I also want to check to what extent this error is specific to the release of CESM2.2 I am using here.   Will also try on other CLM tags including 5.0.34 to see if error is pervasive.

braczka:
Can confirm when executing `CLM5_setup_assimilation` that similar energy balance error with NaNs sent to coupler, occurs even when all observations are only 'evaluated'.  This indicates issue with `clm_to_dart` or `dart_to_clm`.   Currently rechecking input/output for these scripts.

braczka:
The `CM5_setup_assimilation` when `evaluating` all observations produces expected behavior for ensemble members 2-5: a posterior is calculated and no update is applied to the restart files.   Ensemble 1 produces unexpected behavior, in that the restart file for ensemble 1 is updated during the assimilation.   In addition the updated restart value for ensemble 1, is slightly numerically different than the posterior calculated for ensemble 1.  Ensemble 1 is likely creating the energy balance error.   Currently troubleshooting `assimilate.csh` and `dart_to_clm` scripts to identify source of unexpected behavior.

models/clm/shell_scripts/cesm2_2/assimilate.csh
```diff
@@ -0,0 +1,401 @@
...
+# Block 7: Put the DART posterior into the CLM restart file. The CLM
+# restart file is also the prior for the next forecast.
+#=========================================================================
+
+foreach RESTART ( ${CASE}.clm2_*.r.${LND_DATE_EXT}.nc )
+
+   set POSTERIOR = `echo $RESTART | sed -e "s/${CASE}.//"`
+
+   ${LINK} $POSTERIOR dart_posterior.nc
+   ${LINK} $RESTART   clm_prior.nc
```
braczka, line 339: The posterior updates will not be passed onto the restart files unless
`${LINK} $RESTART clm_prior.nc` is changed to
`${LINK} $RESTART clm_restart.nc`

braczka:
Tested the clm_update_3 code branch against the published assimilation (state-13) within Raczka et al., 2021; JAMES.   This manuscript used an older version of the private DART_CLM branch for the Western US assimilation using CAM4 reanalysis.  I merged in the clm_update_3 onto the Utah HPC (Kingspeak) and repeated both the free run and assimilation (state-13), being sure to match both the CLM and DART settings throughout.   Expected outcome was to demonstrate identical or nearly identical assimilation behavior with clm_update_3 branch vs. previous DART_CLM version used in Raczka et al..  The actual outcome was assimilation behavior that was nearly identical, but with noticeable differences in the evolution of important CLM states (AGB, LAI).  See figures below.   Part of this difference likely arose from differences in compiled version of CLM5.0.  Even the 'free' simulations demonstrated slight differences between manuscript version and current version.   These differences likely a result of updated compiler versions on kingspeak for CLM.   Performing some additional checks to verify this is the case.  

`&filter_nml`
	`Inf_flavor = 2`
	`Inf_initial_from_restart = .true.`
	`Inf_sd_initial_from_restart = .true.`
	`Inf_deterministic = .true.`
	`Inf_initial = 1.2`
	`Inf_lower_bound = 1.0`
	`Inf_upper_bound = 100.0`
	`Inf_damping = 0.9`
	`Inf_sd_initial = 0.6`
	`Inf_sd_lower_bound = 0.6`
        `Inf_sd_max_change = 1.05`

`&assim_tools_nml`
   `filter_kind                     = 1`
   `cutoff                          = 0.015`
   `sort_obs_inc                    = .false.`
   `spread_restoration              = .false.`
   `sampling_error_correction       = .false.`
   `adaptive_localization_threshold = -1`
   `output_localization_diagnostics = .false.`
   `localization_diagnostics_file   = 'localization_diagnostics'`
   `print_every_nth_obs             = 1000`

`&obs_kind_nml`
   `assimilate_these_obs_types = 'BIOMASS',`
                                `'LEAF_AREA_INDEX',`
   `evaluate_these_obs_types   = 'null'`

`&quality_control_nml`
   `input_qc_threshold = 1.0`
   `outlier_threshold  = 3.0`

`&model_nml`
   `clm_variables  = 'leafc',       'QTY_LEAF_CARBON',            '0.0', '500',     'restart', 'UPDATE',`
                    `'livestemc',   'QTY_LIVE_STEM_CARBON',       '0.0', '250.0',   'restart', 'UPDATE',`
                    `'deadstemc',   'QTY_DEAD_STEM_CARBON',       '0.0', '15000.0', 'restart', 'UPDATE',`
                    `'TLAI',        'QTY_LEAF_AREA_INDEX',        '0.0', '10.0',    'history', 'UPDATE',`
                    `'frootc',      'QTY_ROOT_CARBON',            '0.0', '800.0',   'restart', 'UPDATE',`
                    `'livecrootc',  'QTY_ROOT_CARBON',            '0.0', '20.0',    'restart', 'UPDATE',`
                    `'deadcrootc',  'QTY_ROOT_CARBON',            '0.0', '6000.0',  'restart', 'UPDATE',`
                    `'leafn',       'QTY_LEAF_NITROGEN',          '0.0', '10.0',    'restart', 'UPDATE',`
                    `'frootn',      'QTY_ROOT_NITROGEN',          '0.0', '20.0',    'restart', 'UPDATE',`
                    `'livecrootn',  'QTY_ROOT_NITROGEN',          '0.0', '0.80',    'restart', 'UPDATE',`
                    `'deadcrootn',  'QTY_ROOT_NITROGEN',          '0.0', '20.0',    'restart', 'UPDATE',`
                    `'livestemn',   'QTY_STEM_NITROGEN',          '0.0', '3.0',     'restart', 'UPDATE',`
                    `'deadstemn',   'QTY_STEM_NITROGEN',          '0.0', '70.0',    'restart', 'UPDATE',`

`&location_nml`
   `horiz_dist_only             = .true.`
   `vert_normalization_pressure = 100000.0`
   `vert_normalization_height   = 10000.0`
   `vert_normalization_level    = 20.0`
   `approximate_distance        = .false.`
   `nlon                        = 71`
   `nlat                        = 36`
   `output_box_info             = .true.`

Comparison of both 'free' and 'assimilation' (state-13) runs between original DART_CLM version and updated branch (clm_update_3):


image



image

braczka:
Update on benchmark testing:   I identified two reasons for differences between the updated (clm_update_3) assimilation Western US case vs. the Raczka et al., 2021 benchmark case.   See difference between state-13 (original) and state-13 (update) in previous comment.   To correct this, I re-ran the benchmark assimilation case using the current compiler versions (UofU's Kingspeak), and also corrected the adaptive inflation setting (Inf_initial =1.2) within the fill_inflation_restart namelist.  Previously, I had this value incorrectly set to inf_initial=1.0, which did not match the settings for the benchmark case.    These changes greatly improved the similarity between the assimilation runs (see differences in state-13 (original) and state-13 (update) below) -- basically scientifically indistinguishable.  However, they are still not bit-by-bit identical.   Currently running one more test with fill_inflation_restart disabled and inflation settings taken directly for &filter_nml for first time step, which should remove remainder of difference.

  image

braczka:
Update to the coupler NaN error:   This error occurs when updating the `H20SOI_ICE` clm variable.  The other related variables like `H20SOI_LIQ` and `T_SOISNO` are no problem when being updated through DART.   Importantly the error when `H20SOI_ICE` is updated is not associated with the snow layers as previously thought (layers 1-12), but the soil layers (layers 13-37).   The current setup will work as long as we don't update the surface and subsurface layers in `H2OSOI_ICE`.   Currently I am in contact with CLM development team to identify a work-around for this error, either through further customizing the update to `H20SOI_ICE` or through SourceMod changes to CLM code.

mgharamti:
> Update on benchmark testing: I identified two reasons for differences between the updated (clm_update_3) assimilation Western US case vs. the Raczka et al., 2021 benchmark case. See difference between state-13 (original) and state-13 (update) in previous comment. To correct this, I re-ran the benchmark assimilation case using the current compiler versions (UofU's Kingspeak), and also corrected the adaptive inflation setting (Inf_initial =1.2) within the fill_inflation_restart namelist. Previously, I had this value incorrectly set to inf_initial=1.0, which did not match the settings for the benchmark case. These changes greatly improved the similarity between the assimilation runs (see differences in state-13 (original) and state-13 (update) below) -- basically scientifically indistinguishable. However, they are still not bit-by-bit identical. Currently running one more test with fill_inflation_restart disabled and inflation settings taken directly for &filter_nml for first time step, which should remove remainder of difference.
> 
> image

Brett, this is great progress. Glad to see you are able to match the previous results.

braczka:
Update on benchmark testing to CLM5-DART Western US assimilation: The slight remaining differences between benchmark assimilation and assimilation using `clm_update_3` were isolated to slight differences in the CLM5 simulation, and not related to DART.   These differences are negligible and I consider the benchmark testing completed.

Update on `H2OSOI_ICE` NaN coupler error:   The soil columns within CLM that led to the NaN coupler error were located in a region remote from the observations (Northern Canada).   These soil columns *did not* receive an update through the assimilation (all increments were zero).   These soil columns, however, did receive ensemble wise perturbation  from the adaptive inflation setting during the first time step:

`&fill_inflation_restart_nml`
   `write_prior_inf   = .true.`
   `prior_inf_mean    = 1.01`
   `prior_inf_sd      = 0.6`

Apparently the inflation led to the error.   I turned off inflation for the first assimilation time step by setting `prior_inf_mean = 1.00`.   This eliminated inflation for the 1st assimilation time step, and greatly reduced the inflation to only spatial regions close to the observations for each assimilation time step thereafter (days 1-5).   This switch in the assimilation setting eliminated the NaN coupler error for all assimilation time steps and allowed the assimilation to run successfully (5 days with daily assimilation).     I reverted the `clm_to_dart.f90` to remove the hard-wired vertical localization such that the subsurface layers (layers 13-37 were allowed to vary again).  The assimilation also ran successfully.   

It is not exactly clear why a relatively large inflation setting (1.01) caused the assimilation to trigger the NaN coupler error.  In general, the adaptive inflation algorithm calculated inflation values much smaller than 1.01 for H2OSOI_ICE (e.g. 1.005 was the largest I could find across entire globe).     The spatial location of the error (Northern Canada) was strongly snow-covered (probably glaciated), and contained high amounts of surface ice (>100 kg /m2).   When inflation was turned on for the first assimilation time step, this did not create unrealistic `H2OSOI_ICE` values.    Nevertheless, turning the inflation off for the first assimilation time step and allowing the adaptive inflation (flavor=5) to calculate the inflation values has resolved the error.   I am performing some final checks, but plan to  permanently implement `prior_inf_mean = 1.00`, and allow for entire snow/soil column to vary (1-37), yet still maintain the mechanism for DART to preserve the original column value if it is an 'indeterminate' value (i.e. no snow exists in layer).

mgharamti:
> Update on benchmark testing to CLM5-DART Western US assimilation: The slight remaining differences between benchmark assimilation and assimilation using `clm_update_3` were isolated to slight differences in the CLM5 simulation, and not related to DART. These differences are negligible and I consider the benchmark testing completed.
> 
> Update on `H2OSOI_ICE` NaN coupler error: The soil columns within CLM that led to the NaN coupler error were located in a region remote from the observations (Northern Canada). These soil columns _did not_ receive an update through the assimilation (all increments were zero). These soil columns, however, did receive ensemble wise perturbation from the adaptive inflation setting during the first time step:
> 
> `&fill_inflation_restart_nml`
> `write_prior_inf = .true.`
> `prior_inf_mean = 1.01`
> `prior_inf_sd = 0.6`
> 
> Apparently the inflation led to the error. I turned off inflation for the first assimilation time step by setting `prior_inf_mean = 1.00`. This eliminated inflation for the 1st assimilation time step, and greatly reduced the inflation to only spatial regions close to the observations for each assimilation time step thereafter (days 1-5). This switch in the assimilation setting eliminated the NaN coupler error for all assimilation time steps and allowed the assimilation to run successfully (5 days with daily assimilation). I reverted the `clm_to_dart.f90` to remove the hard-wired vertical localization such that the subsurface layers (layers 13-37 were allowed to vary again). The assimilation also ran successfully.
> 
> It is not exactly clear why a relatively large inflation setting (1.01) caused the assimilation to trigger the NaN coupler error. In general, the adaptive inflation algorithm calculated inflation values much smaller than 1.01 for H2OSOI_ICE (e.g. 1.005 was the largest I could find across entire globe). The spatial location of the error (Northern Canada) was strongly snow-covered (probably glaciated), and contained high amounts of surface ice (>100 kg /m2). When inflation was turned on for the first assimilation time step, this did not create unrealistic `H2OSOI_ICE` values. Nevertheless, turning the inflation off for the first assimilation time step and allowing the adaptive inflation (flavor=5) to calculate the inflation values has resolved the error. I am performing some final checks, but plan to permanently implement `prior_inf_mean = 1.00`, and allow for entire snow/soil column to vary (1-37), yet still maintain the mechanism for DART to preserve the original column value if it is an 'indeterminate' value (i.e. no snow exists in layer).

Brett, again great progress! 
The problem I believe is not the size of the inflation but rather the way the first priors are getting inflated. When you specify in the namelist the input prior inflation to be 1.01, that means a spatially constant inflation of 1.01 will be applied to all variables and across the entire domain. However, when you start with 1.00 the first priors will not be inflated. The updated inflation values will be applied in the next cycle. This actually confirms the idea that inflating the domain with a single inflation value is not a good strategy (and generally causes model instabilities) and thus a spatially-varying  inflation algorithm is needed hence we use flavors 2 or 5 (but quite rarely 3). So, I agree that we need to start from 1 and let the algorithm itself adapts both in space and in time.

braczka:
"The problem I believe is not the size of the inflation but rather the way the first priors are getting inflated."

Yes, I agree with that, it seems applying a spatially uniform inflation value across a global domain is heavy-handed.   I will say, I have gotten away with that approach in previous CLM assimilations where I applied a spatially uniform inflation value of 1.2 for the first assimilation time step.  By comparison, an inflation setting of 1.01 seemed tiny to me.  * However*, it is very clear to me now that inflation settings are highly dependent upon the spatial domain of the model, and also the set of variables that are adjusted.   

Another surprise for me was that the problem was occurring in a region where no increments were applied, only inflation.   I had convinced myself the observations (and ultimately the increments) were the root cause of the issue.   Definitely a learning experience.

braczka:
After discussions with Moha, I tested a slightly different assimilation configuration to `CLM5_setup_assimilation` by changing `inf_lower_bound=1.0` which prevents 'deflation', to `inf_lower_bound=0.0`, which allows for deflation.   This is the preferred lower bound if using `inf_flavor =5` because the inverse gamma distribution prevents negative inflation values.   The assimilation ran successfully with similar/identical increments applied during the 6 days of the assimilation testing.   Just for reference I have included a comparison of the inflation values of a sample column of `H2OSOI_ICE` using the different `inf_lower_bound` to demonstrate the behavior was generally as expected:

Day 3 inflation:

`inf_lower_bound = 0.0`
   `0.996696363434313,   // H2OSOI_ICE(10,98531)`
    `0.997161730727702,   // H2OSOI_ICE(11,98531)`
    `0.994528841365125,   // H2OSOI_ICE(12,98531)`
    `0.997700280408085,   // H2OSOI_ICE(13,98531)`
    `0.999502019847758,   // H2OSOI_ICE(14,98531)`
    `0.995918563436969,   // H2OSOI_ICE(15,98531)`
    `0.999167447695364,   // H2OSOI_ICE(16,98531)`

`inf_lower_bound = 1.0,`    
    `1,   // H2OSOI_ICE(10,98531)`
    `1,   // H2OSOI_ICE(11,98531)`
    `1,   // H2OSOI_ICE(12,98531)`
    `1,   // H2OSOI_ICE(13,98531)`
    `1,   // H2OSOI_ICE(14,98531)`
    `1,   // H2OSOI_ICE(15,98531)`
    `1,   // H2OSOI_ICE(16,98531)` 

Day 6 inflation:
`inf_lower_bound = 0.0`
   `1.04219257526185,   // H2OSOI_ICE(10,98531)`
    `1.04361324091681,   // H2OSOI_ICE(11,98531)`
    `1.01854066800181,   // H2OSOI_ICE(12,98531)`
    `1.02898740477667,   // H2OSOI_ICE(13,98531)`
    `1.04737603172158,   // H2OSOI_ICE(14,98531)`
    `1.01522728086136,   // H2OSOI_ICE(15,98531)`
    `1.03652259186461,   // H2OSOI_ICE(16,98531)`

`inf_lower_bound = 1.0`
    `1.05383015743004,   // H2OSOI_ICE(10,98531)`
    `1.05675980951699,   // H2OSOI_ICE(11,98531)`
    `1.0237013854683,   // H2OSOI_ICE(12,98531)`
    `1.03904786386798,   // H2OSOI_ICE(13,98531)`
    `1.05786245590889,   // H2OSOI_ICE(14,98531)`
    `1.02006023203328,   // H2OSOI_ICE(15,98531)`
    `1.04153882084117,   // H2OSOI_ICE(16,98531)`

A reminder that this 'fixes' the Nan coupler error, in that, inflation values other than =1 are localized to where they are needed (i.e. where the observations are).   An interesting follow-up test would be to place an observation close to the location where the Nan coupler error had occurred to force a perturbation.   However, I don't believe that is necessary to push this PR forward.   Currently, I am revisiting the assimilate/evaluate 'obs' issue I brought up earlier in this thread.

PR #259, opened 2021-07-23
bug-fix: initialize uniquek to an invalid level.

### LINKED ISSUE #258: bug: uninitialized unique levels in wrf can lead to different results on different core counts.

### Describe the bug

Originally reported by a DART-WRF-chem user

WRF case provided by user
10 ensemble members
no adaptive inflation
no inflation
horizontal location only

Note the original report had inflation, vertical localization, etc. For reproducing for debugging switched these off. 

available at /glade/work/hkershaw/DART/cheyenne_run288

Same number of observations assimilated no matter what the number of processors. 
3. What actually happened?  
Different forward operators fail at various core counts. The example given in  /glade/work/hkershaw/DART/cheyenne_run288
produces bitwise identical results (with debugging flags) for 36, 72, 144, 576 cores, but has on 288 cores 1 observation fails the prior forward operator. 
Note you can output the posterior forward operator to see other differences.  

Screen Shot 2021-07-22 at 4 05 56 PM

WRF 
and originally reported with WRF-chem.  I don't have the WRF-chem code at the moment, so there could be other problems.  

### Screenshots   

Here is an uninitialized uniquek k that happens to have a reasonable level value in (on the right).  This trips up the all(uniquek /= k(e)) so you end up with level 29 and level 106766972 (=forward op fail on right) rather than level 28 and 29(=forward op success pn left). 

Screen Shot 2021-07-22 at 10 50 15 AM

### Version of DART      
9.11.3
but I think this has been around for several years. 

Yes  
I added wrf-chem quantities to run with the users obs_sequence file. But the actual case uses regular DART obs. 

### Build information
Cheyenne intel compiler. 
I think the uninitialized uniquek could affect any compiler/machine combo.

hkershaw-brown:
Fixes issue #258

## Description:

Initialize uniqiek in wrf model_mod to have an invalid level

### Fixes issue

Fixes #258 

### Types of changes

- [ ] Documentation update

### Tests
Bitwise with main branch (caveat the 288 processor case which showed the bug). 
Cheyenne, intel & gnu compilers.

## Checklist for merging

- [x] Updated changelog entry
- [ ] Documentation updated
- [x] Version tag 

## Testing Datasets

- [ ] No dataset needed

Test case here: 
/glade/work/hkershaw/DART/cheyenne_run288

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,11 @@ individual files.
 
 The changes are now listed with the most recent at the top.
 
+**July 23 2021 :: bug fix for wrf non-initialized unique levels. Tag: v9.11.4**
+
+- The array uniquek is now initialized to an invalid level to prevent random
+  reasonable level values in the array.
+
 **June 24 2021 :: bug fix for cam-fv model_interpolate. Tag: v.9.11.3**
 
 - cam-fv model_interpolate now passes the correct array slice of quad_vals
--- a/conf.py
+++ b/conf.py
@@ -21,7 +21,7 @@ copyright = '2021, University Corporation for Atmospheric Research'
 author = 'Data Assimilation Research Section'
 
 # The full version, including alpha/beta/rc tags
-release = '9.11.3'
+release = '9.11.4'
 master_doc = 'README'
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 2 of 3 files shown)


hkershaw-brown:
I've assumed here that -1 is an invalid level for wrf.

nancycollins (approved):
-1 is indeed an illegal value for the Z level.  this fix should prevent random levels from being skipped if by chance one of the valid level numbers is already in the uninitialized array.

Issue #261, opened 2021-07-30
mpich mpif08


hkershaw-brown:
I think the mpi_utilities errors are mpich using f77 bindings when you `use mpi`. Hello world mpi (no dart code) will not compile if you have a mpi_send(integer) mpi_send(real) in the same file.

[hello.f90.txt](https://github.com/NCAR/DART/files/6525415/hello.f90.txt)
[hello_f08.f90.txt](https://github.com/NCAR/DART/files/6525419/hello_f08.f90.txt)

The f08 mpi module `use mpi_f08` follows the fortran standard. 
There is a branch mpi_f08 which has a version of mpi_utilites_mod.f90 with mpi_f08 bindings. This will work with mpich and gfortran 10+ (only done a quick test).
The cray window module (cray pointers) could get replaced with iso_c_bindings. 

or short term solution: `port select --set mpi openmpi-mp-fortran`

_Originally posted by @hkershaw-brown in https://github.com/NCAR/DART/issues/158#issuecomment-846307106_

hkershaw-brown:
I love these foundation shaking questions from Nancy.   Have you played with autoconf to configure/make DART vs. mkmf?

hkershaw-brown:
other notes from #288
review MPI_INTEGER_KIND, COUNT MPI3 as part of this update.

hkershaw-brown:
up-to-date now.  I think this should go in a f2008(3) update (see https://github.com/NCAR/DART/pull/288).  Hanging out on the mpi_f08 branch now

hkershaw-brown:
moved mpi_f08 https://github.com/hkershaw-brown/DART/tree/mpi_f08 
someone can grab it or not as needed if they are updating to f08.

hkershaw-brown:
Gust gfortran

hkershaw@gust01:~/DART/models/lorenz_96/work.gfortran> module -t list
ncarenv/23.04
craype/2.7.20
gcc/12.2.0
ncarcompilers/0.8.0
hdf5/1.12.2
netcdf/4.9.1
cray-mpich/8.1.25

<img width="1270" alt="Screen Shot 2023-05-01 at 6 16 12 PM" src="https://user-images.githubusercontent.com/20047007/235541023-212ad613-7b3a-4ffc-8457-a55447b078a6.png">

hkershaw-brown:
To reproduce this on gust you can use the following modules:

```
hkershaw@gust01:~/DART/models/lorenz_96/work> module load gcc
[...]

Issue #158, opened 2021-02-12
NCEP libraries: gfortran 10.2.0 and the argument-mismatch


timhoar:
GNU Fortran (MacPorts gcc10 10.2.0_5) 10.2.0  is very strict about rank mismatches and such. Putting `-fallow-argument-mismatch` in the FFLAGS allows the compiler to proceed - but it still throws a ton of warnings, which is a bit of unnecessary chatter in the output log. 

Worse ... the NCEP libraries do not build without manual intervention, because they do not use the mkmf.template - they use the `NCEP/prep_bufr/install.sh` script to build - which does not dynamically use the FFLAGS from the `mkmf.template`.

The `obs_converters/run_tests.csh` script queries the `mkmf.template` for the compiler and then sets some variables that are used by `install.sh` ... perhaps this can be extended to use the `FFLAGS` as well.

nancycollins:
i'll give you some history on why it is how it is.  the prepbufr libraries are downloaded from NCEP.  they have been updated over the years but the original code is very old.  we decided to make as few changes as possible so when a new version was downloaded it wouldn't require redoing a lot of work that would again get lost in the next update.  install.sh is what comes with the libs, and the strange directory structure and filename convention (bob.x for executables) is all theirs.   the prepbufr obs converter itself is very much based on a dump program that is distributed with the library.

having said that, yes, i snuck some code into install.sh to try to use the "current" compiler if running in a higher level script.  i'm not sure if any of the other FFLAGS we set in mkmf.template will cause the libs to not compile.  if not, then sure, the same process can be extended, i'd think.  users may still want to build just the ncep libs, so install.sh needs to be able to work standalone and not only in a testing environment.

i'll caution people not to try to "fix" the lib code itself unless it won't build.  as long as it's working, no one cares and anything done there will be overwritten if a new version is downloaded.  if you don't look, you won't be horrified.

timhoar:
So I am 'discovering' this problem again.  Just to recap and to help aid in any possible searches, here is an example of the error:
```
Error: Rank mismatch between actual argument at (1) and actual argument at (2) (rank-1 and scalar)
../../../assimilation_code/modules/utilities/mpi_utilities_mod.f90:1394:19:
```

Fixing `install.sh` to define

```
elif [ $FCOMP = gnu ] ; then
   ff='gfortran -fallow-argument-mismatch'
```
[...]

PR #264, opened 2021-08-04
New test for verifying one-sided MPI communication 

nancycollins:
## Description:

Add a new MPI one-sided test to verify MPI is working correctly.   Update other tests to use the standard 'system' block formatting, add a version of fixsystem here.  Add new test to Makefile and update it.  Update the README to fix directory names, contact info, add ftest_onesided info.

i also have some minor non-code changing updates to other files in this dir.  Will commit them separately later.

### Fixes issue

Currently there is a user report of problems with one-sided communication.  This standalone test should verify that the one-sided MPI functions work.  (Even if MPI works, sometimes there are ways for one-sided to break.)

### Types of changes

- [ x] New feature (non-breaking change which adds functionality)
- [ x] Documentation update

- [ x] My change requires a change to the documentation.
  - [x ] I have updated the documentation accordingly.

### Tests
Built and ran in this directory on cheyenne compute node (needed for MPI0 with intel and gfortran.

## Checklist for merging

- [ ] Updated changelog entry
- [ x] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [ x] No dataset needed

### DIFF
--- a/developer_tests/mpi_utilities/tests/Makefile
+++ b/developer_tests/mpi_utilities/tests/Makefile
@@ -2,16 +2,16 @@
 # by UCAR, "as is", without charge, subject to all terms of use at
 # http://www.image.ucar.edu/DAReS/DART/DART_download
 #
-# DART $Id$
 
 # Makefile for various mpi test and timing programs
 #
-# This Makefile directly 'include's the mkmf.template file used by the 
+# This Makefile directly includes the mkmf.template file used by the 
 #  mkmf system.  If the mkmf.template is set wrong for this system, 
-#  these builds will fail.  See ../../mkmf/mkmf.template.* for various
-#  compiler and platform settings.
+#  these builds will fail.  See ../../../build_templates directory
+#  for various different compiler and platform files..
 
 include ../../../build_templates/mkmf.template
+
 
 # the include defines the following makefile variables:
 # MPIFC =    # mpi fortran wrapper
@@ -24,25 +24,37 @@ include ../../../build_templates/mkmf.template
 # LIBS =     # any additional libraries needed
 # LDFLAGS += $(FFLAGS) -L$(NETCDF)/lib -lnetcdf  # loader flags
 
+# if you want to test the C programs, you will need to
+# define CC and MPICC here because they are not defined in
+# the standard DART templates
+CC    = gcc    # c compiler (e.g. gcc, icc, cc)
+MPICC = mpicc  # mpi wrapper for cc
+
+# one test uses the DART utilities
+UTILS_DIR = ../../../assimilation_code/modules/utilities
+
 # ----------------------------------------------------------------------------
 
 SRC = ftest_f90.f90 ftest_mpi.f90 ftest_sendrecv.f90 ftest_nc.f90 \
-      ftest_stop.f90 ftest_go.f90 ctest.c ctest_mpi.c ctest_nc.c
+      ftest_onesided.f90 ftest_stop.f90 ftest_go.f90 \
+      ctest.c ctest_mpi.c ctest_nc.c
 
 ALLSRC = $(SRC) Makefile README runme_*
 
-EXE = ftest_f90 ftest_mpi ftest_nc ftest_sendrecv
+EXE = ftest_f90 ftest_mpi ftest_nc ftest_sendrecv ftest_onesided
 ALLEXE = $(EXE) ftest_stop ftest_go ctest ctest_mpi ctest_nc
 
 # ----------------------------------------------------------------------------
 
 # default target - build and test only what is necessary for dart
-all:  $(EXE)
-
+all:  fixsys $(EXE)
 
 # all executables - if you are having problems, this might help diagnose
 # what parts of the install are giving problems
-everything: $(ALLEXE)
+everything: fixsys $(ALLEXE)
+
+fixsys:
+	./fixsystem $(FC)
 
 
 # run the basic executables interactively
@@ -50,6 +62,7 @@ check:  all
 	./ftest_f90                   # this should run
 	./ftest_nc                    # this should run
 	mpirun -np 2 ./ftest_mpi      # this may not be allowed on a login node
+	mpirun -np 2 ./ftest_onesided # this may not be allowed on a login node
 
 # submit the jobs to a batch queue
 batch: ftest_mpi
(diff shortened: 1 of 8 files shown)


nancycollins:
i have updated versions of the remaining tests in this dir that simply remove the older SVN lines.  they aren't worth a separate release, but i would like to update them at some point.  can they be part of this pull request or should they be a separate pull request?  sorry, i am unclear on how 'non-code-change-cleanup' things are being handled.

hkershaw-brown:
quick question on the fixsystem.  Can you use the fixsytem script already in DART?

nancycollins:
> quick question on the fixsystem. Can you use the fixsytem script already in DART?

great question.  i looked at it and the differences with in the one in the utilities dir are:

1) it has the mpi_utilities_mod.f90 and null_mpi_utilities_mod.f90 filenames hardcoded in it.
2) it fails if it can't find the key strings in the files.

i looked at trying to make it take filenames as an argument but it already takes the compiler name as an optional arg.  the compiler name could be changed to be required and not optional, and then followed by the filenames to process, with the mpi modules as a default.  but those 2 files must have the strings in them and i'd still want it to fail if they aren't found.  (that might mean they've been corrupted, or edited by hand and no longer have the key strings that sed needs to work.)  only some of the files in the test directory have a call to system() and so some of the test files don't need to be changed.  i could list the exact test files that do need editing, i suppose, in the call in the Makefile to fixsystem if i could work out how to supply optional filenames in addition to a compiler type.

i don't like replicating the functionality but wasn't sure i wanted to mess with the utilities directory version.  that will need a lot more testing.  you can convince me that's a lesser evil compared to replicating the file.

nancycollins:
i've updated the utilities version of fixsystem to work with the test files, and removed the copy of fixsystem.  this still needs to be tested with NAG because i think the original fixsystem logic was wrong.  waiting to see if i can get access to one before it's ready to merge.

nancycollins:
ok.  i've tested this with the NAG compiler on hobart.cgd.ucar.edu.   i've had to update the mkmf.template file to get DART to build ok.  i've removed the go/stop mpi tests because no one uses async4 anymore.  they're in the history if they need to be revived.  i've updated the ftest_mpi test to have the same conditional blocks as the current mpi modules, and it's now using the utilities directory version of fixsystem, so i've removed the copy here.

nancycollins:
i haven't done anything to the version or changelog files.

hkershaw-brown:
Hey Nancy, 

Can you add what changes are in this pull request to description 
I think it is this:
* new mpi one-sided test
* bug-fix(?) for fixsystem for the nag compiler
* removed async 4 tests
*  added flag to mkmf.template.nag.linux (is this just for the tests or for general DART compiling?)
* portability-fix(?) for strings.h (-> string.h)
* bug-fix: use `stop` instead of `call exit`

A couple of questions/paranoia:
1. . There are still references to async and asyn4 so I am a bit wary of removing the tests, e.g.
`models/lorenz_96/shell_scripts/run_filter_async4.csh`
https://docs.dart.ucar.edu/en/latest/guide/filter_async_modes.html?highlight=async
What do you think?
2.  Using -mismatch with mpi is the f77 vs mpi08 problem #158 #261.  It seems a bit strange to use the nag compiler, but downgrade the consistency checking of procedure argument lists.  Do you think nag compile updates should be with an update to use mpif08?

nancycollins:
i updated the description, but the changelog and version still need work.   to answer your questions:

1) i think we should eventually remove all references to async4.  while it worked, it was fragile and overly complex.  i think if we need to revive that behavior there may be newer MPI facilities that allow one MPI program to spawn another, or there may be other solutions that aren't so cryptic.  i vote async4 off the island.

2) i looked that those issues with MPI and NAG and yes, it's exactly the same problem.  we have a subroutine in the mpi_utilities_mod.f90 file that first calls MPI_Ssend() with a real array, and then calls MPI_Ssend() with an integer array.  that's perfectly fine for MPI - there's an argument to tell MPI what the array type is.  but the compiler freaks out if they're both used in the same subroutine.  i guess for now i'd leave -mismatch in the nag mkmf.template, because it still prints out warnings if it sees mismatching args but it doesn't error out.    this is needed for building dart, not just the tests.  if derecho is going to have NAG installed then we probably will have to address this more completely in the future, but for now i think it's an ok intermediate step.  i will add a comment in the mkmf.template.nag.linux file about this so there's a reminder of why that option is used.  if/when 'use mpi_f08' is widely available, we can start using that in our code instead of just 'use mpi'.

hkershaw-brown:
> 
> 1. i think we should eventually remove all references to async4.  while it worked, it was fragile and overly complex.  i think if we need to revive that behavior there may be newer MPI facilities that allow one MPI program to spawn another, or there may be other solutions that aren't so cryptic.  i vote async4 off the island.

So,  leave the tests in and remove asyn4 at a later date?

nancycollins:
the async4 tests, if they stay in, will need to be updated with the same conditional code that i added to ftest_mpi.f90.  it's a copy of what's in the real mpi_utilities_mod.f90.  these tests are supposed to be standalone (although ftest_sendrecv isn't - i'm not sure why and i don't remember the history there) so i can't re-use the utilities module code.  i took them out rather than update them.  if you really think they need to stay, let me know and i'll go back and add the conditional sections to them.

originally we didn't have the 'fixsystem' script, so people had to edit mpi_utilities_mod by hand to make it work.  when i added support for the NAG compiler there were a couple other conditional sections added that the script took care of.  but the tests were never updated to use the script, so to make them work someone would have had to edit them by hand if they used anything other than the gfortran compiler.  i've made them look more like how the dart code works so they should be easier for someone to run, but are still standalone in case they need to run a small program, or send their system support a small program as an example of something not working.

hkershaw-brown:
oh ok, fair enough, let's get rid of the async4 tests.

nancycollins:
the regression test for fixsystem is to build L96 with various compilers.  it should build.  

the regression test for the programs in the developer_tests/mpi_utilities/tests directory is to 'make everything' with various compilers.  they should build.

nancycollins:
i've finished messing around with the new ftest_onesided.f90 source.  i think i've cleaned it up and that the logic is simpler to follow, the subroutines are cleaner, and the comments are less wrong.   i think this code is ready for review.  i'll be gone for the next 4 days starting tomorrow, so no rush on this.

PR #285, opened 2021-09-09
Updates to DART for CM1

jonlabriola:
## Description:

I am trying to run DA experiments with the CM1 model and decided to update DART to better work with my experiment design.  I have made the following updates:

1.) Updated the interpolation scheme in ./cm1/model_mod.f90 so that the forward operator for the cm1 model can now perform three-dimensional interpolation.  Originally the model_mod.f90 for cm1 only interpolated fields on a two-dimensional surface, such as wind speed at 10 m AGL.  Now you can pass DART a three-dimensional field such as reflectivity and interpolate across the full volume.

2.) Corrected a bug in ./threed_cartesian/location_mod.f90:  The original code did not properly offset distances across periodic boundaries, and consequently the covariance information could not wrap-around the northern and eastern boundaries of the experiment domain.   I updated the get_dist function in location_mod.f90 to correct this bug and better handle loopy%offset.

3.) Updated ./cm1/model_mod.f90 and ./threed_cartesian/location_mod.f90 so that users can runDA experiments with mixed-case boundary conditions.  For example users can run experiments with boundary conditions that are periodic in the x-direction but non-periodic in the y-direction.  

4.) Added capability to change localization radius for each assimilated observation type for the CM1 model.  To do this I updated the get_close functions in threed_cartesian/location_mod.f90 so that the get_close data type contained a box data type that could be used to calculate localization information for each specified cutoff radius type. This is similar to what is done in the threed_sphere code.

### Types of changes

- [ ] Documentation update

The DART documentation online mentions special_localization_obs_types "is only implemented for the threed_sphere location module".  If you decide to include these updates, CM1 (which assumes threed-cartesian coordinates) would also be able to use special_localization_obs_types.  No other NWP models that assume threed-cartesian coordinates will be able to use this option.

### Tests

Change 1.) To ensure three-dimensional interpolation works I compared DART interpolated u-wind velocity values against my own interpolation estimates using python code.  Results were identical.

Changes 2 and 3) To ensure boundary condition updates work correctly I ran several experiments where I assimilated a single wind speed observation near the domain boundary (x=185 km y = 185 km). Please see attached powerpoint to see how an observation updates surrounding grid points assuming x-periodic BC’s, y-periodic BC’s and xy-periodic BC’s.

Change 4) I assimilated two seperate observation types (OBS 1 at x = 50 y = 100, OBS 2 at x = 150 y = 100) with very different covariance localizations to ensure the localization radius changes with observation type.  Results (please see attached images) show that the code appears to work correctly.

A demonstration of the updates can be found at the attached link:
[DART_Presentation.pdf](https://github.com/NCAR/DART/files/7137687/DART_Presentation.pdf)
## Checklist for merging

- [x] Updated changelog entry
- [x] Documentation updated
- [x] Version tag 

## Testing Datasets

- [ ] No dataset needed


hkershaw-brown:
Hi Jon, thanks for this!
Do you have a CM1 test case (restart files, obs_seq, input.nml) that we can use to test on our machines.
Cheers,
Helen

jonlabriola:
Yes, I have a test case uploaded to google drive that I can share. The tar file includes all the restart files, obs_seq.prior, input.nml, and the batch submission file used to assimilate a single V-wind observation located in the northeast corner of the experiment domain.  The tar file also contains obs_seq.prior_twoobs and input.nml_twoobs, which can be used to assimilate two separate observations and see if the localization radius changes with observation type.

models/cm1/model_mod.f90
```diff
@@ -1147,7 +1180,7 @@ function observation_on_grid(obs_location, ndims)
 
    endif
 else
-   write(string1,*) 'only grids with periodic x and y grids, or non-periodic '
+   write(string1,*) ' This option is not supoorted '
    write(string2,*) 'boundary conditions supported'
```
nancycollins, line 1184: is the original string1 line wrong now?  the new string1 message (besides the typo) doesn't say *which* option is not supported.  also if that line stays as-is, string2 is now a non-sequitur and should be removed.

jonlabriola: Right now, no code in the "else" statement should be called because I account for instances where x and y are periodic, x alone is periodic, y alone is periodic, and neither x nor y is periodic.  I included that section as a sanity check just in case a user put in an odd combination of boundary condition values that could not be recognized by DART.

nancycollins: makes sense, so how about:

```suggestion
   write(string1,*) ' Unsupported combination of periodic boundary conditions.'
   write(string2,*) 'Contact DART support for more help.'
```

jonlabriola: That looks great

nancycollins:
i see in the test input.nml file that radar obs are listed as possible assimilation types.

be aware that observation sequence files need to use the same location type as the filter and other executables.  whatever program is going to create the obs will need to be compiled with threed_cartesian.

also, for filter the radar forward operator has code that assumes threed_sphere (lat/lon/vert/vert type) locations - e.g. the metadata includes the location of the radar unit along with az/el of the beam.  the obs_seq that is created for CM1 runs has to use x/y/z cartesian coords in order to be read by filter or any of the other executables.

i haven't though enough about what would have to change to support radar obs that live in an x/y/z world.  if they're synthetic it might be easier than if they're real obs.  the programs i'm aware of to declutter buildings, birds, etc from real radar obs are all going to be assuming threed_sphere locations.

jonlabriola:
I have only been working with synthetic radar observations and have not dealt with real-case radar observations. Right now I assign each reflectivity/Vr observation an x, y, and z location and calculate the azimuth/elevation angle based upon the location of the assumed radar site.   I've compared radar plots calculated via my python scripts and the DART forward operator and the results look quite similar. Minor discrepancies occur in part because the python and DART synthetic radar observations are calculated at slightly different grid points (a few 100 m difference in location).  DART radial velocity estimates are also slightly different in part because DART calculates Vr where observed precipitation is present while my scripts calculate Vr where precipitation is present in the prior model output (i.e., my scripts treat the prior like observations). 
[DART_Radar_Operator.pdf](https://github.com/NCAR/DART/files/7179565/DART_Radar_Operator.pdf)

I don't have any experience working with real radar observations in DART, but could we use a function to convert observations from threed_sphere to threed_cartesian before outputting them into an obs_seq file? I noticed the NCOMMAS model in DART has a subroutine that allows you to convert from lon/lat to x/y coordinates (ll_to_xy in model_mod.f90).

nancycollins:
> I don't have any experience working with real radar observations in DART, but could we use a function to convert observations from threed_sphere to threed_cartesian before outputting them into an obs_seq file? I noticed the NCOMMAS model in DART has a subroutine that allows you to convert from lon/lat to x/y coordinates (ll_to_xy in model_mod.f90).

very good question.  i think converting the locations to cartesian is a good idea for the obs_seq file.  there will also probably be some additional work needed in the forward operator code. (the code that computes the 'expected value' at filter run time for each obs.)

i haven't looked at it for a while but i remember the radar FO code is assuming locations are lat/lon and we'll have to see what calculations get affected by that.

models/cm1/model_mod.f90
```diff
@@ -1127,6 +1129,37 @@ function observation_on_grid(obs_location, ndims)
       return ! exit early
 
    endif
+elseif ( periodic_x) then
+    ! require that the point is contained within the staggered grid for the 
+    ! y - direction since you cannot wrap-arround
+    if ( (obs_location(1) < xf(1)) .or. (obs_location(1) > xf(nip1)) .or. &
+        (obs_location(2) < yh(1)) .or. (obs_location(2) > yh(njp1)) ) then
```
hkershaw-brown, line 1136: The yh grid is nj long, so go off the end of the array with njp1.

```suggestion
    if ( (obs_location(1) < xf(1)) .or. (obs_location(1) > xf(nip1)) .or. &
        (obs_location(2) < yh(1)) .or. (obs_location(2) > yh(nj)) ) then
```

Maybe I am mis-understanding the comment, but the code seems opposite to the comment here:

non-staggered x would be xh
staggered in y would be yf

models/cm1/model_mod.f90
```diff
@@ -1127,6 +1129,37 @@ function observation_on_grid(obs_location, ndims)
...
+         print *, 'x_min, x_max ', xf(1), xf(nip1)
+         print *, 'y_min, y_max ', yh(1), yh(nj)
+      endif
+
+      return ! exit early
+    endif
+elseif ( periodic_y) then
+    ! require that the point is contained within the staggered grid for the 
+    ! x - direction since you cannot wrap-arround
+    if ( (obs_location(1) < xh(1)) .or. (obs_location(1) > xh(nip1)) .or. &
```
hkershaw-brown, line 1150: Same as above comment: xh is lenght ni:

```suggestion
    if ( (obs_location(1) < xh(1)) .or. (obs_location(1) > xh(ni)) .or. &
```

or swap xh -> xf, yf -> yh

models/cm1/model_mod.f90
jonlabriola: You are correct, the updates look good.

models/cm1/model_mod.f90
jonlabriola: Same here, the updates look good.

jonlabriola:
Hello, I'm touching base to see this code is able to be merged into the main branch, or if I need to do more work to prepare it for merger.  In addition to the updates in this pull request I would also like to push forward an update I made that allows users to modify the vertical localization radius for the CM1 DART code.  Some of the preliminary radar DA experiments I have been running look promising.  Thank you again for all the time and effort this group puts in to maintaining the large code package!

- Jon

hkershaw-brown:
Hi Jon,
This ended up on the back burner due to some other pull requests, and some initial debate about good tests for the location_mod. But we'll bump this up priority and get it reviewed and merged.

There are a couple of suggestions to be committed (in the future you can do this in the Github interface by clicking on "commit suggestion"). The second thing to do is to bring this branch up to date with the current main branch for DART.  I will go ahead and make these changes this afternoon. 

Thanks again for you contributions.  
Cheers,
Helen

assimilation_code/location/threed_cartesian/location_mod.f90
```diff
@@ -67,32 +68,56 @@ module location_mod
...
+!   type(box_type)    :: box
+!end type get_close_type
+
 type(random_seq_type) :: ran_seq
 logical               :: ran_seq_init = .false.
 logical, save         :: module_initialized = .false.
 
 character(len = 512) :: errstring
-
+character(len = 512) :: msgstring, msgstring1, msgstring2
```
nancycollins, line 121: sorry - i can't figure out how to leave a comment in the unchanged part of the code.  the comment block that starts below on lines 154 - 157  should be removed, right?  because that's what this pull request is adding - per-type localization distances as an option.

assimilation_code/location/threed_cartesian/location_mod.f90
```diff
@@ -272,7 +297,6 @@ function get_dist(loc1, loc2, type1, kind2)
 !> for all combinations of location 1 and location 2 where
```
nancycollins, line 297: again, a comment for code in the unchanged section.  this isn't your change, but i think the control flow would be clearer if line 285 was an 'endif', not an else.  if no periodic conditions, get distance and return.  period.

then lines 286-288 are no longer inside a conditional section, and the rest of the subroutine deals with the periodic issues.  i'd eventually like to see if we can deal with X, Y, and Z once but for now i believe the code will work with all 8 cases separate and that's fine.

assimilation_code/location/threed_cartesian/location_mod.f90
```diff
@@ -1020,30 +1178,72 @@ subroutine get_close(gc, base_loc, base_type, locs, loc_qtys, &
...
-if (gc%num == 0) return
+if (gc%box(bt)%num == 0) return
 
-this_maxdist = gc%maxdist
+this_maxdist = gc%box(bt)%maxdist
 
 !> @todo this is doing an exhaustive search each time.  expensive
 !> but should give the right answer.
-
 if(.true.) then
```
nancycollins, line 1240: is this really intended to be .true. ??  this code is going to call exhaustive_collect() every time and not use any of the code in the rest of this routine.  should this be .false. when not debugging?  or true when one or more of the dimensions are periodic?  i haven't timed this yet but i'm sure it is going to slow down other models which use the cartesian get_close() routine.

assimilation_code/location/threed_cartesian/location_mod.f90
```diff
@@ -1222,8 +1422,8 @@ end subroutine find_box_ranges
 
 !----------------------------------------------------------------------------
 
-subroutine find_nearest(gc, base_loc, loc_list, nearest, rc)
- type(get_close_type), intent(in), target  :: gc
+subroutine find_nearest(box, base_loc, loc_list, nearest, rc)
```
nancycollins, line 1425: find_nearest() is a public routine so it can't have a box_type as the first arg.  it needs to be passed a get_close_type.   this routine doesn't use maxdist at all - it returns the single nearest location to the target.  it should be ok to simply use bt=1 to get the first box_type from the gc.

nancycollins (commented):
i've added some comments with some suggested changes.  i'm still testing some things but i will come back and re-review once these changes are addressed.

models/cm1/work/input.nml
```diff
@@ -16,14 +16,15 @@
...
+  num_output_obs_members      = 40
+  num_output_state_members    = 40
   output_interval             = 1
   output_members              = .true.
   output_mean                 = .true.
   output_sd                   = .true.
-  stages_to_write             = 'preassim', 'analysis', 'output'
+  !stages_to_write             = 'preassim', 'analysis', 'output'
+  stages_to_write             = 'forecast', 'preassim', 'postassim', 'analysis', 'output'
   write_all_stages_at_end     = .true.
```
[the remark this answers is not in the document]
jonlabriola: All the updates to the namelist look reasonable.  Since the time I originally submitted this pull request I made additional updates that tend to optimize the performance of the DA experiments I am conducting.  Should I submit this updated namelist, or is this template good enough?

assimilation_code/location/threed_cartesian/location_mod.f90
jonlabriola: I believe we want to keep these lines of code in tact.  That's because the term lines 286 - 288 define the term square_dist, which later referenced in the function when working with the periodic boundary conditions.

assimilation_code/location/threed_cartesian/location_mod.f90
jonlabriola: I remember first seeing this section of code when I started editing location_mod.f90.  Since I believe it was in the original code, I did not run any tests to see what the impact is on accuracy/time if I block the option.   Also I believe CM1 is the only model that uses the file 'assimilation_code/location/threed_cartesian/location_mod.f90'.  mpas is on a cartesian grid but refers to 'xyz_location_mod.f90'.

assimilation_code/location/threed_cartesian/location_mod.f90
jonlabriola: This all makes sense to me.  After looking through the DART code, I realized this subroutine is not called anywhere.  I bet that is why I never had any errors come up previously when running the code.

Also sorry, but this is my first time making any type of contribution to a public code package. To correct these bugs, do you want me to fix my own local copy and push it into the pull request?

jonlabriola: Also I have a few more questions:  
1.) What is the best strategy to make sure this branch is up to date with the main code?  Should I update my main branch and use that to rebase this current branch?

2.) Since opening this pull request I created a new branch that is capable of making horizontal/vertical localization radii different for CM1.  Should I complete this pull request first, and once complete open a new one for those updates?

assimilation_code/location/threed_cartesian/location_mod.f90
nancycollins: you're absolutely right those lines have to stay.  it's just now i realize the non-periodic code is about 6 lines and the periodic section is about 100.  i'd like to visually make the two cases more distinct.  i was thinking of something like

```
! the simple non-periodic case
if (.not. any_periodic) then
   if (debug > 0) write(0,*)  'non-periodic distance: '
   get_dist = dist_3d(diff)
   return
endif

! everything below deals with the periodic options
if (debug > 0) write(0,*)  'periodic distance: '
square_dist = dist_3d_sq(diff)
if (debug > 0) write(0,*)  'non-periodic 0 sq_dist, dist: ', square_dist, sqrt(square_dist)

```

nancycollins:
hi jon,  thanks for addressing my comments.  your changes look good and i'm sure we'll incorporate them.  this comment is not really directed at you; i'm just going to put a bit of history of the threed_cartesian module here for future reference.

the original need for a threed_cartesian locations module was indeed the CM1 model.  we implemented it, but i'm not sure any of our original users had any periodic cases so the periodic code wasn't tested much.  at the time we didn't have support for per-type localization distances in any location module, so that code wasn't added either.

when an mpas interface to dart was written we discussed how to handle cases at the poles where a cell of the mpas grid would enclose the actual pole.  the lat/lon grid math was frightening.  we decided we could convert obs locations and grid locations into cartesian coords on demand when needed and there would be fewer problems.  i started by working on the actual threed_cartesian code, adding code for find_nearest and optimizing how to speed up get_close, since mpas has huge grids.

it took me quite a while to realize that i couldn't include both the threed_sphere location module (lat, lon, vert for both obs and the native mpas grid locations) and the threed_cartesian location module in the same executable.  we could have used fortran 'use' lines to rename the subroutines so there weren't name collisions (use bob, only : routine_a => routine_b).   but you can't include two modules with the same module name.  all location mods have to have the same module name to work interchangeably with the rest of the dart system, so that was the sticking point.

eventually i copied the threed_cartisian module into xyz_locations_mod.f90, cut out what wasn't needed, and worked on optimizing that code more.  it's now used to find the closest grid point to an obs location for interpolation in mpas, and it was used in the cubed-sphere grid code for that version of cam (cam-se).

fully supporting all the possible periodic cases, and adding support for per-type localization distances is a good addition to this code.

assimilation_code/location/threed_cartesian/location_mod.f90
nancycollins: hi jon.  yes, please fix your local copy and push the changes.  we'll see them here and can review what's been addressed.

i'll let helen reply to your git questions.  (i'm a novice at git but i can answer any subversion questions you might have!)

i would definitely finish this pull request and make the horizontal/vertical localization changes a different pull request.  multiple smaller pull requests are easier to review and test, and help to ensure lurking bugs don't creep in with huge changes that don't get caught for a long time.

assimilation_code/location/threed_cartesian/location_mod.f90
nancycollins: it's possible that this was unintentionally left .true., so for now i'd leave it.  that probably means the "optimized" search code below this case hasn't been fully tested.  we'd like you to make progress on your project so leave this alone and we can test this later.  it's going to take some complicated tests to ensure that the code below (which tries to search only boxes around the base location) actually searches enough of the space to get all locations which are within the given radius.

models/cm1/work/input.nml
```diff
@@ -16,14 +16,15 @@
...
   distributed_state           = .true.
```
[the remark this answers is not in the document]
nancycollins: (rant - i find git reviews confusing sometimes to know exactly what lines are being referenced).

i meant 'distributed state' is in the namelist twice separated by quite a few other items.

assimilation_code/location/threed_cartesian/location_mod.f90
jonlabriola: That is a smart idea, this update looks good to me.  I'll incorporate it within my code.

jonlabriola:
> Hi Jon, This ended up on the back burner due to some other pull requests, and some initial debate about good tests for the location_mod. But we'll bump this up priority and get it reviewed and merged.
> 
> There are a couple of suggestions to be committed (in the future you can do this in the Github interface by clicking on "commit suggestion"). The second thing to do is to bring this branch up to date with the current main branch for DART. I will go ahead and make these changes this afternoon.
> 
> Thanks again for you contributions. Cheers, Helen

Hi Helen,  sorry for being slow to get to the proposed updates.  I noticed you mentioned I need to make sure my local git repository is up to date with the master version of DART.  How do I do this?  Should I do a git pull from main your copy of DART?

hkershaw-brown:
Hi Jon, 

Yup, correct, you pull from https://github.com/NCAR/DART main into your branch. 

I did update your branch `jonlabriola:updt_bcs` with the latest changes in commit 5eb1d8bc6f062319ee89b4830a73e3d8e447bb06  (main hasn't change since then) so you are up to date at the moment. 

Cheers,
Helen

jonlabriola:
I see that now, thank you for the quick response! I'll try to get through all the extra change requests in the next few days.

hkershaw-brown:
Perfect!

jonlabriola:
Hi All, 

I have a few questions and comments about how I should configure the DART CM1 input file.  I pushed forward an updated input.nml file to highlight some of my concerns.

1.) I updated my CM1 code to output air_temp, spec_hum, and fall_spd (now located in the under “model_variables” in the namelist.  This was done to assimilate RADIOSONDE_TEMPERATURE, RADIOSONDE_SPECIFIC_HUMIDITY, and RADAR_RADIAL_VELOCITY observations, respectively.  To make the CM1 DART more user friendly (i.e., avoid making every user update there own personal CM1 code), should I remove these model variables and observations from the input file?  I can mention in the cm1 readme file that temperature and specific humidity fields can be asismilated if CM1 is updated to output the fields. I should be able to still assimilate radial velocity observations because the DART forward operator includes calculations for fall speed.

2.) In a similar vein, I deleted several model variables from the old CM1 namelist (u10,v10,t2,tsk,q2, psfc).  These fields are required to assimilate several observations (TEMPERATURE_2M, U_WIND_10M, V_WIND_10M, SURFACE_PRESSURE, SPECIFIC_HUMIDITY_2M); however, CM1 only outputs them if you are running simulations with a PBL scheme. Even though CM1 simulations are often run without a PBL scheme, should I put these model variables back in the CM1 namelist because they were originally there?

hkershaw-brown:
Hi Jon, 

I think the notes you have in the input.nml are great. 

For this:
> To make the CM1 DART more user friendly (i.e., avoid making every user update there own personal CM1 code), should I  remove these model variables and observations from the input file? I can mention in the cm1 readme file that temperature and specific humidity fields can be asismilated if CM1 is updated to output the fields. 

Keep the variables the input.nml but commented out with a note that CM1 needs to be modified to use these model variables.  I think it is good to add to the readme & input.nml because who reads instructions, right? The input.nml they will definitely read. 

The cam-fv input.nml is an example where there are comments about various options, state variables and choices for CAM vs. WACCM. 
https://github.com/NCAR/DART/blob/main/models/cam-fv/work/input.nml
So your input.nml notes will be helpful for someone else running a DART-CM1 experiment. 

Cheers,
Helen

jonlabriola:
Thank you for the advice Helen, I made some updates to the input.nml documentation and added to the readme file.  I also corrected any concerns brought up about input.nml and the location_mod.f90.  Please let me know what else I can do to help.

Jon

hkershaw-brown:
Excellent, thanks Jon.

hkershaw-brown (approved):

Issue #278, opened 2021-08-27
bug: model size is sometimes i8 sometimes i4


hkershaw-brown:
:bug: Your bug may already be reported!
Please search on the [issue tracker](https://github.com/NCAR/DART/issues) before creating a new issue.

Originally reported by a user trying to run a very wrf large model. 

### Describe the bug

1. List the steps someone needs to take to reproduce the bug.  
use wrf model mod with model size declared as i8
run perfect model_obs

2. What was the expected outcome?
I think a user should be able to have a model size i8

3. What actually happened?  
bomb in direct_netcdf_mod.f90.  For example `subroutine read_transpose`
`169 integer,                   intent(inout) :: dart_index `

There may be other locations where integer is being used instead of integer(i8)

### Which model(s) are you working with?
WRF

### Version of DART      
v9.11.8

### Have you modified the DART code?   
Yes  to add i8 to wrf model_mod.f90

hkershaw-brown:
big pop run with i8?

nancycollins:
testing with 1/10th degree pop should smoke out any i8/i4 issues introduced into the main dart code.  but if the WRF model_mod code has i8/i4 issues with the state vector size or indices into the state vector, you'll have to find/construct a large wrf state to test that.

hkershaw-brown:
do you know where 1/10th degree pop is? Is this it: https://github.com/NCAR/DART_development/tree/static_data

nancycollins:
the code?  it's just the standard models/POP 

the data?  good question.  possibly on image.ucar.edu in the data area, or on campaign storage, or lost to the sands of time.

hkershaw-brown:
The big pop test case as is, is too small to catch  `direct_netcdf_mod.f90` i4 vs i8

big pop test case:
`/glade/p/cisl/dares/DART_test_cases/pop/pop_tx0.1v2/`

|                  | # elements       |
|-----------|----------------|
| model size | 2,151,360,000 |
| 3D variable size | 535,680,000 |
| 2D variable size | 8,640,000 |
| i4 | 2,147,483,647 |

````
 PE 0: verify_state_variables variable  1 is SALT_CUR, QTY_SALINITY, UPDATE
 PE 0: verify_state_variables variable  2 is TEMP_CUR, QTY_POTENTIAL_TEMPERATURE
 , UPDATE
 PE 0: verify_state_variables variable  3 is UVEL_CUR, QTY_U_CURRENT_COMPONENT, 
 UPDATE
[...]

PR #288, opened 2021-09-16
Model size i8 fix for state vector io and mpi_utilities

hkershaw-brown:
## Description:

Model size in DART can be an i8 number. However, some of the indexing into the state vector was not set up to use i8.
This problem was discovered by a user was running a large wrf perfect_model_obs (3796738754 elements in the state vector).

This pull request contains:

* wrf model_mod.f90 fixes to use i8  for model size. This includes removing the function `get_wrf_index` which is not used (and not i8)
* direct_netcdf_mod.f90, state_structure_mod.f90, state_vector_io_mod.f90 have i8 integers for variables that deal with indices into the state vector. 
* mpi_utitliites_mod.f90  `itemcount` is now i8 so the `else` clause of `if (itemcount <= SNDRCV_MAXSIZE)` is entered.  Bug fix 
  in the receive section for number of elements sent (was itemsize, now next size)
  There is a test added in developer_tests/mpi_utilities/tests   `ftest_sendrecv_big` to test sending large messages.
* check and bail if my_num_vars is >= i4
* comment correction in perfect_model_obs. removes outdate info about perfect_model_obs being not-mpi.

### Fixes issue

fixes: #278

### Types of changes

- [x] Documentation update

I have changed the description for convert_state_vertical which did not have loc_indx(:) (i8 index into the state vector)

### Tests

* perfect_model_obs with 3796738754 element wrf to test the state io.  Checked output state is the same as the input state (not advancing the model). Note I ran this on big memory nodes on Casper.
* added a ftest_sendrecv_big for large message.
* bitwise small wrf filter (intel and gfortan) 

## Checklist for merging

- [x] Updated changelog entry
- [x] Documentation updated
- [x] Version tag 

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/assimilation_code/modules/utilities/ensemble_manager_mod.f90
+++ b/assimilation_code/modules/utilities/ensemble_manager_mod.f90
@@ -712,6 +712,17 @@ subroutine set_up_ens_distribution(ens_handle)
 type (ensemble_type),  intent(inout)  :: ens_handle
 
 integer :: num_per_pe_below, num_left_over, i
+integer(i8) :: per_pe, suggest_pes
+
+! Check that there are enough pes for the state
+per_pe = ens_handle%num_vars / num_pes
+if (per_pe >= (huge(i)-100) ) then
+   suggest_pes = ( ens_handle%num_vars / (huge(i)) ) * 2
+   write(msgstring, '(A,I5,X,A)') &
+   'not enough MPI tasks for the model size, suggest at least ' , &
+   suggest_pes, 'tasks'
+   call error_handler(E_ERR, 'set_up_ens_distribution', msgstring, source)
+endif
 
 ! Option 1: Maximum separation for both vars and copies
 ! Compute the total number of copies I'll get for var complete
--- a/assimilation_code/modules/utilities/ensemble_manager_mod.f90
+++ b/assimilation_code/modules/utilities/ensemble_manager_mod.f90
@@ -57,7 +57,7 @@ type ensemble_type
    !!!private
    integer(i8)                  :: num_vars
    integer                      :: num_copies, my_num_copies
-   integer(i8)                  :: my_num_vars
+   integer                      :: my_num_vars
    integer,        allocatable  :: my_copies(:)
    integer(i8),    allocatable  :: my_vars(:)
    ! Storage in next line is to be used when each pe has all copies of subset of vars
@@ -672,7 +672,7 @@ function get_my_num_vars(ens_handle)
 ! Returns the number of vars stored on my processor when copy complete.
 ! Same as num_vars if single process is in use.
 
-integer(i8)                      :: get_my_num_vars
+integer                          :: get_my_num_vars
 type (ensemble_type), intent(in) :: ens_handle
 
 get_my_num_vars = ens_handle%my_num_vars
@@ -712,7 +712,7 @@ subroutine set_up_ens_distribution(ens_handle)
 
 type (ensemble_type),  intent(inout)  :: ens_handle
 
-integer(i8) :: num_per_pe_below, num_left_over
+integer :: num_per_pe_below, num_left_over
 integer :: i
 
 ! Option 1: Maximum separation for both vars and copies
@@ -825,7 +825,7 @@ function get_max_num_vars(ens_handle, num_vars)
 
 type (ensemble_type), intent(in)  :: ens_handle
 integer(i8), intent(in) :: num_vars
-integer(i8)             :: get_max_num_vars
+integer                 :: get_max_num_vars
 !!!integer, intent(in) :: distribution_type
 
 !could this be instead:
--- a/assimilation_code/modules/utilities/ensemble_manager_mod.f90
+++ b/assimilation_code/modules/utilities/ensemble_manager_mod.f90
@@ -56,7 +56,8 @@ type ensemble_type
    !DIRECT ACCESS INTO STORAGE IS USED TO REDUCE COPYING: BE CAREFUL
    !!!private
    integer(i8)                  :: num_vars
-   integer                      :: num_copies, my_num_copies, my_num_vars
+   integer                      :: num_copies, my_num_copies
+   integer(i8)                  :: my_num_vars
    integer,        allocatable  :: my_copies(:)
    integer(i8),    allocatable  :: my_vars(:)
    ! Storage in next line is to be used when each pe has all copies of subset of vars
--- a/assimilation_code/modules/utilities/ensemble_manager_mod.f90
+++ b/assimilation_code/modules/utilities/ensemble_manager_mod.f90
@@ -672,7 +672,7 @@ function get_my_num_vars(ens_handle)
 ! Returns the number of vars stored on my processor when copy complete.
 ! Same as num_vars if single process is in use.
 
-integer                          :: get_my_num_vars
+integer(i8)                      :: get_my_num_vars
 type (ensemble_type), intent(in) :: ens_handle
 
 get_my_num_vars = ens_handle%my_num_vars
@@ -712,7 +712,8 @@ subroutine set_up_ens_distribution(ens_handle)
 
 type (ensemble_type),  intent(inout)  :: ens_handle
 
-integer :: num_per_pe_below, num_left_over, i
+integer(i8) :: num_per_pe_below, num_left_over
+integer :: i
 
 ! Option 1: Maximum separation for both vars and copies
 ! Compute the total number of copies I'll get for var complete
@@ -824,7 +825,7 @@ function get_max_num_vars(ens_handle, num_vars)
 
 type (ensemble_type), intent(in)  :: ens_handle
 integer(i8), intent(in) :: num_vars
-integer                 :: get_max_num_vars
+integer(i8)             :: get_max_num_vars
 !!!integer, intent(in) :: distribution_type
 
 !could this be instead:
--- a/assimilation_code/modules/utilities/ensemble_manager_mod.f90
+++ b/assimilation_code/modules/utilities/ensemble_manager_mod.f90
@@ -56,8 +56,7 @@ type ensemble_type
    !DIRECT ACCESS INTO STORAGE IS USED TO REDUCE COPYING: BE CAREFUL
    !!!private
    integer(i8)                  :: num_vars
-   integer                      :: num_copies, my_num_copies
-   integer                      :: my_num_vars
+   integer                      :: num_copies, my_num_copies, my_num_vars
    integer,        allocatable  :: my_copies(:)
    integer(i8),    allocatable  :: my_vars(:)
    ! Storage in next line is to be used when each pe has all copies of subset of vars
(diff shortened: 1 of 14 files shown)


hkershaw-brown:
hold the phone on this.

I'd convinced myself that ensemble manager would not need my_num_vars(i8) b2f0e7bc40afbdc8caa561a1389e5c58b5a5c2ca
but, after chatting with Moha, it is totally reasonable that you might have a massive state vector and want to run single core.
f2263ae692d170b24c6f97b32c087f6d8f2cdd6a because people run things other than filter.

nancycollins:
maybe i'll try L96 with a >32bit state size and run it single task to see where it breaks.  (it may run into system limitations first.)

both model_mod_check and PMO can be built and run with mpi, just fyi.  i'm not sure if we have to support >32bit states in a single task or not.  we could say you have to run multi-core in that case, and convert any single task programs to use a distributed state if they don't already.  that might be a quicker route to success than following where i8 my_num_vars in the ensemble manager takes you.

i don't know about now, but in the past some compilers have required you build with additional flags if you're going to use >32bit data segments (e.g. --big-model or something).  when the compiler generates references to data items in the execution image, they can use relative offsets which are often only 4 byte ints.

nancycollins:
there is a comment in perfect_model_obs.f90 near mpi_initialize_utilities() that says we only run pmo single task but that's obsolete.  it has built and run with mpi for a long time.  i'd like to remove that comment on this branch if that's ok - it is tangentially related to i8 issues and not worth a pull request of its own.

nancycollins:
testing hint:  if on cheyenne, get a compute node first.  trying to allocate huge arrays on login nodes gets you nuked.

nancycollins:
testing L96 with a state size > 32bits mysteriously dies in the ensemble manager at transpose time.  if we decide that we are going require that users run multi-task if their state > 32bits we should add a size test and helpful error message before getting to this point.

hkershaw-brown:
> 
> i don't know about now, but in the past some compilers have required you build with additional flags if you're going to use >32bit data segments (e.g. --big-model or something). when the compiler generates references to data items in the execution image, they can use relative offsets which are often only 4 byte ints.

So this is stack variables that you have to worry about, dynamically allocated variables not a problem, right?

The flag is -mcmodel=medium(or large). Same for gfortran or intel

hkershaw-brown:
> testing L96 with a state size > 32bits mysteriously dies in the ensemble manager at transpose time. if we decide that we are going require that users run multi-task if their state > 32bits we should add a size test and helpful error message before getting to this point.

The transpose will be doing something uncool like 

```
ens_handle%copies(:,1:-22343253) = buffer
```

nancycollins:
there are usually 5 separate segments loaded when a program runs: text (code), initialized data, uninitialized data, stack and heap (run-time allocated data).  the loader tells the compiled program the full memory address of where the segments start in memory and then the executable adds offsets to that base to generate addresses of code and data.

those offsets are relative to the start of the segment.  base memory addresses these days are 64 bits, but the offsets are still typically 32 bits (4 bytes).  building with a large memory model forces them to be 64 bit, too.  it makes the code segment larger because offsets are twice as big now.  i don't know what kind of integers the stack and heap use for indexing.  

for those who think visually, start here:  https://en.wikipedia.org/wiki/Data_segment

i've got the dragon book, too (aho and ullman - the canonical compiler book) if anyone wants to go down that rabbit hole.

mgharamti:
Nancy and Helen, since you both are testing this and I'm working on a similar problem I thought I'd ask here. I'm trying to run model_mod_check in parallel on Cheyenne using an interactive job but I'm getting ``forrtl: severe (41): insufficient virtual memory`` errors. Any ideas on how to fix that?

nancycollins:
can you point me to the directory where you're running from, and the directory where you've built the executables if that's a different place?  there may be some array in model_mod_check that's being allocated state vector size instead of the number of items that are "owned" by this task.

mgharamti:
Nancy, everything is in one place: /glade/work/gharamti/DART/MITgcm-NBLING-DART/models/MITgcm_ocean/work

nancycollins:
cool, thanks.  i'll take a look.

nancycollins:
when i set this in the input.nml and ran model_mod_check with 36 tasks:

```
&state_vector_io_nml
   buffer_state_io         = .true.
```

i was able to run tests 1, 2 and 3 successfully.  (i didn't try going further.)  so there's hope.

nancycollins:
test 7 cannot be run on large state vector sizes.  model_mod_check::find_closest_gridpoint() has an allocate of get_model_size().   i think the right fix is for it to allocate only the number of state vector points on its own task and do the exhaustive search only on its own points.  each task would find a 'closest' point and some code would need to be added to send all the distances to task 0 and have it print out the state vector point with the smallest distance.

nancycollins:
test 4 is failing at the call which creates the mpi window.  am checking to see what size window it is requesting and what the limits might be on sizes.

nancycollins:
i was able to get moha's test case to run test 4 by doubling the number of tasks from 36 to 72.  the mpi one-sided communication creates a copy of the state data to share with other tasks, effectively doubling the memory needed for the state on each task.  

i also think that test 4 should only call interpolate on task 0.  some model_mod_check tests have to be called by all tasks (e.g. read_state and write_state) and some are doing duplicate work that never prints out (interpolation of a single point - right now it happens on all tasks but only prints from task 0.)   that will complicate debugging if this program finds errors in the model_mod routines (which is its purpose).

tests 4 and 5 should only call interpolate from task 0.  tests 6 and 7 need more work because one writes to an output file - that should only happen from a single task or each task should write to a different file - and test 7 allocates a get_state_size() array instead of ensemble_handle%my_num_vars.  it also needs some additional mpi calls to consolidate distances from each task onto task 0 and then have it select the shortest distance.

i have a copy here:  

/glade/work/nancy/git/main/assimilation_code/programs/model_mod_check/model_mod_check.f90

which only runs the interpolation tests 4 and 5 on task 0.  i have not yet tried to update tests 6 and 7 yet.

hkershaw-brown:
the test for 0 files perfect_model_obs fix is already on main - this is why this branch is showing up as having a conflict.  Nancy I'm going to revert your commit, then bring this branch up to date with main.

nancycollins:
i thought i pulled the main branch down before i did my update but that caused a conflict?  sorry - i'm still baffled by git sometimes.

hkershaw-brown:
> i thought i pulled the main branch down before i did my update but that caused a conflict? sorry - i'm still baffled by git sometimes.

no worries it is only git.

hkershaw-brown:
notes from chat with Nancy:

The model_mod_check improvements are going into a different pull request.

my_num_vars is staying as i4.  The reasons for this are:
* You run out of memory even if you can keep count of the indices. e.g.:
20 member ensemble, r8, 4,294,967,294 state is 687GB.  
* my_num_vars touches a lot of the code (particularly get_close).

This pull request adds a check in setup_ens_distribution to bail if my_num_vars is too large.

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -576,7 +576,7 @@ subroutine send_to(dest_id, srcarray, time, label)
    if (verbose) write(*,*) "PE", myrank, ": send_to ", itemcount, " dest ", dest_id
 
    if (.not. make_copy_before_sendrecv) then
-      call MPI_Ssend(srcarray, itemcount, datasize, dest_id, tag, &
+      call MPI_Ssend(srcarray, int(itemcount,i4), datasize, dest_id, tag, &
```
nancycollins, line 579: this looks ok, but there must be an mpi kind defined for the number of items.  at some point they may change it from a 4 byte int to an 8 byte int.   does:  int(itemcount, MPI_INTEGER) work? or whatever the prototype for MPI_Ssend gives for this argument?   MPI does have a lot of kinds defined for different uses.

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -586,7 +586,7 @@ subroutine send_to(dest_id, srcarray, time, label)
       ! have been needed, and is a performance/memory sink.
 
       tmpdata = srcarray
-      call MPI_Ssend(tmpdata, itemcount, datasize, dest_id, tag, &
+      call MPI_Ssend(tmpdata, int(itemcount,i4), datasize, dest_id, tag, &
```
nancycollins, line 589: same comment here.

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -606,11 +606,11 @@ subroutine send_to(dest_id, srcarray, time, label)
       if (verbose) write(*,*) 'sending array items ', offset, ' thru ' , offset + nextsize - 1
 
       if (.not. make_copy_before_sendrecv) then
-         call MPI_Ssend(srcarray(offset:offset+nextsize-1), nextsize, datasize, dest_id, tag, &
+         call MPI_Ssend(srcarray(offset:offset+nextsize-1), int(nextsize,i4), datasize, dest_id, tag, &
```
nancycollins, line 609: ditto

assimilation_code/modules/utilities/ensemble_manager_mod.f90
```diff
@@ -712,6 +712,17 @@ subroutine set_up_ens_distribution(ens_handle)
 type (ensemble_type),  intent(inout)  :: ens_handle
 
 integer :: num_per_pe_below, num_left_over, i
+integer(i8) :: per_pe, suggest_pes
+
+! Check that there are enough pes for the state
+per_pe = ens_handle%num_vars / num_pes
```
nancycollins, line 718: this looks good.  great idea to suggest a bottom threshold for how many PEs/tasks are needed to make each task's share of the data small enough to fit in an i4.   they'll probably have to add more tasks to fit on a node's memory but this gives them at least a minimum.

developer_tests/mpi_utilities/tests/ftest_sendrecv_big.f90
```diff
@@ -0,0 +1,60 @@
+! DART software - Copyright UCAR. This open source software is provided
+! by UCAR, "as is", without charge, subject to all terms of use at
+! http://www.image.ucar.edu/DAReS/DART/DART_download
+
+program ftest_sendrecv
```
nancycollins, line 5: do you want this to be ftest_sendrecv_big ?

nancycollins (approved):
looks good - optionally look at using a MPI_xxx kind for the MPI_Ssend and MPI_Recv calls.  (maybe MPI_INTEGER_KIND?)   otherwise all changes fine.

developer_tests/mpi_utilities/tests/ftest_sendrecv_big.f90
hkershaw-brown: yes yes yes

assimilation_code/modules/utilities/mpi_utilities_mod.f90
hkershaw-brown: There is MPI_COUNT_KIND but it is MPI3.  I want to do this, but as part of updating mpi_utilities_mod to f08 and MPI3.

https://github.com/NCAR/DART/issues/261

window f08 update branch:
https://github.com/NCAR/DART/tree/mpi_f08

Also, I think we can remove the cray pointer window. 
And also, maybe (_maybe_) use fortran intrinsics for executing other programs so we don't have to use fixsystem.  Note I don't know if we can do this, especially since we do things asynchronously.

Issue #298, opened 2021-10-11
bug: Definition and use of "unnormalized" scale height in cam-fv is inconsistent


kdraeder:
### Describe the bug
There are several, related bugs, which I've tried to separate into managable units.
This is the first I'd like to resolve, even though I filed one (#296) before this one.

The vertical coordinate "scale height" is defined as log(ps/p), ps = surface pressure (or some other reference)
and log = natural log ln. This can be written as  log(ps) - log(p).
When the difference of 2 scale heights is calculated, the 2 log(ps) terms cancel,
so we could save some computation by defining an "unnormalized" scale height = log(1/p) = - log(p)
and calculating SH1 - SH2 = log(p2) - log(p1)  (note the numbers).
But when one of those unnormalized SH is compared against some threshhold height, 
like ramping_end, that threshhold must be calculated the same way - unnormalized - which it was not.

This topic was further confused by the misdefinition of unnormalized scale height as "log(p)" instead of "- log(p)"
and the resulting, unnecessary complexity in defining higher_is_smaller, and unintended consequences
in routines that use higher_is_smaller.

1. List the steps someone needs to take to reproduce the bug. 
Set up a cam-fv assimilation using  

- vertical_localization_coord = SCALEHEIGHT
- model_damping_ends_at_level = 5
- an obs_seq.out file which has obs at heights above model level 6

2. What was the expected outcome?
There should be no increments at and above layer 5.

3. What actually happened?  
There are non-0 increments in those layers.  
The unnormalized scale heights of the state variables have values ranging from 
5.88 (log(360 Pa)) to 11.5 (log(99200 Pa)),  while the ramp_start (normalized scale height) 
had  a value of ~ 3.0 (log(10^5/3500) - .3 ramp_depth).
It would appear that all of those state variables are in the damped layer,
but the choice of using unnormalized SH caused higher_is_smaller to be .true. (which it should not have),
and that caused v_above to test whether the state variables have values *less* than ramp_start.
They do not, so above_ramp_start decided that *none* of them were in the damped layer.
and so their increments were not damped.

### Which model(s) are you working with?
Cam-fv only (all versions of CAM and WACCM(-X))

### Version of DART      
Which version of DART are you using?   Manhattan, all of the versions

### Have you modified the DART code?   
Only cam-fv/model_mod.f90
I'll issue a pull request shortly on a branch of Manhattan.

### Build information
Please describe:  
[...]

PR #299, opened 2021-10-12
Scaleh

kdraeder:
## Description:
Define scale height with the straightforward, conventional definition.
Define "unnormalized" scale height consistent with that.
Define vertical threshholds using no_normalization_of_scale_heights, instead of hard coding .false.

Renamed variable higher_is_smaller to higher_is_decreasing.
This is (still) ambiguous, but keeps part of the old name and is less cumbersome than something unambigous
like gradient_with_height_is_negative.

Added debugging code (default is off) to above_ramp_start to help with verifying that it's working.

Applied nint to the reals returned by query_location.  This may not have been broken, but it seemed worth doing
while I was debugging.

~~Fix the order of dimensions in model_interpolate: call quad_lat_lon_evaluate(...quad_vals) to match the definition.  This isn't related to the vertical coordinate issues, but is a trivial fix of something 
that keeps surprising me.~~ HK edit: this was already fixed on main.

### Fixes issue
#298 

### Types of changes

- [x] Documentation update

### Tests
I ran an assimilation using:

- just radiosonde U winds, 
- no_normalization_of_scale_heights = .true.
- model_damping_ends_at_level = 7 (a level below the highest U obs)

The program entered the appropriate sections of above_ramp_start for the heights of the observations.
There were no increments at the levels above model_damping_ends_at_level.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [ ] No dataset needed


hkershaw-brown:
Hello, where did you branch this from and where do you want it to go? main and main?
At the moment, it is against Manhattan, so there are 1182 files changed, which seems a little excessive.

hkershaw-brown:
You can edit the pull request to change the base branch to main.
Screen Shot 2021-10-12 at 9 32 03 AM

Your scaleh is a couple of months behind main. 
Screen Shot 2021-10-12 at 9 31 05 AM

You can make your branch update with main:

```
git checkout main
git pull
git checkout scaleh
git merge main scaleh
```

nancycollins:
we should have a consistent dart-wide definition of scale height, not only in the cam interface code.  being able to specify the vertical localization in units of scale height is an option for any model that uses the threed_sphere locations code.  specifically i know the wrf model_mod has lots of code that deals with scale height.  if these changes are consistent with how wrf or any other model_mod that includes scale height conversion code, then i'm all for it.  otherwise we need to be careful to not make inconsistent definitions or non-backwards compatible changes.  for example - do we know what vertical localization choices the waccm folks use?  if scale height, does this change their results?

kdraeder:
I changed the base and updated my local main and scaleh branch.

Nancy, good point.  I'll try to dig that stuff up.  I've never looked in those models,
so I don't know how long that will take.

kdraeder:
> kevin - when you tested this new code, what did you use for the scale height normalization factor in the locations module? would it be around the log of whatever the pressure normalization factor is or different? it would be interesting to see the differences between setting the vertical localization to pressure and then to scale height. do you have any plots of those?
I used 
vert_normalization_scale_height = 1.5
We can't compare the 2 using a simple log(p) because there's not a linear relationship between them.
They're used to decide whether a vertical distance between 2 points is large or small.
They're both distances, but they represent space differently and have different units.
Near the surface points that are 20000 Pa apart are much closer (geometrically) 
than if 1 of the points were in the thermosphere and separated by 20000 Pa.  
Scaled height preserves the measure of closeness with altitude, which is why we use it
for high topped models.
So 1.5 scaled heights covers a lot of pressure difference near the surface,
but only a little pressure difference in the upper atmosphere.

vert_normalization_pressure = 20000  
Calculating log(20000) is using 1 Pa as a reference pressure and yields the scaled height difference 
between 20000 and 1 hPa = 9.9 scaled heights.
The same 20000 Pa distance _below_ 20000 Pa translates into log(40000/20000) = 0.69 scaled heights.

We did comparisons of using the 2 different vertical coordinates (pressure and scaled height) years ago.
They perform similarly in the lower atmosphere, but scaled height is much better in the upper atmosphere,
which was the motivation for implementing it.
> 
> i didn't see where you added a global variable for the scale height cutoff and removed the local variable in the init routine, but you'd catch that with testing because it wouldn't work otherwise.
I didn't add it to this commit because it's more closely related to another issue that I'm planning to open.

models/cam-fv/model_mod.f90
```diff
@@ -3584,7 +3584,8 @@ subroutine init_discard_high_obs()
 ! incoming observations which have vertical units of scale height.
 ! so far we have localized in scale height but never had obs
```
nancycollins, line 3585: this comment is misleading and should be updated at some point, since you're moving the call to 'obs_too_high()' to the get_close routines.  the call in model_interpolate() hasn't seen incoming obs with scale height as vertical, but that's getting removed.  at get_close() time the obs are converted to the vertical localization type, so definitely yes there will be obs that have this type.

models/cam-fv/model_mod.f90
```diff
@@ -3691,6 +3699,9 @@ function above_ramp_start(test_value, gc, obs_type, ramp_end, total_dist, extra_
 ! FIXME: test this!!!
 ! is it above the ramp end? set damp dist to something
 ! large enough to turn off all impacts.  is vert_localize_dist enough?
+! KDR it's 0.3 in my tests on 2021-9-5.
+!     cutoff*vert_norm_scaleheight = 0.15*1.5 = .225, so "yes" in this case.
+! ? Are those values used to set the maxdist in the gc?
 vert_localize_dist = get_maxdist(gc, obs_type)
 if (.false. .and. onetime) then
    print *, 'vert_localize_dist = ', vert_localize_dist
```
nancycollins, line 3707: this debugging code shouldn't have been in the committed version either.  that's my fault.

hkershaw-brown:
I'm closing this since I think we've ended up reviewing code at bit too early. 

Lets step back from reviewing the code, and start with:
1. Documentation of what the code is supposed to do.
2. A link to restart files + obs_seq that lets someone reproduce the issue (on Cheyenne or wherever).  
3. If this is going to change people's results, who does it affect. e.g does this affect everyone who localizes in SCALEHEIGHT? All WACCM users? Is this a non-bitwise change for CAM users?

hkershaw-brown:
moved scaleh branch to https://github.com/hkershaw-brown/DART/tree/scaleh

PR #312, opened 2021-11-18
Removed the outlier threshold test for posterior adaptive inflation s…

### LINKED ISSUE #310: bug:
There is an inappropriate check for an outlier when updating posterior inflation with groups in filter_assim. This problem is only possible when running posterior spatially-varying adpative inflation with more than one group. In that case, an outlier threshold is computed with a hard-coded threshold of 3 for each group. However, an outlier threshold with the namelist value has already been computed when the posterior forward operator is computed. A fix for this will simply comment out the current check on an okay value for the outlier_ratio with a statement that is always true. This problem would impact any model. 

            if (inflate_only) then
               diff_sd = sqrt(obs_err_var + r_var)
               if (diff_sd > 0.0_r8) then
                  outlier_ratio = abs(obs(1) - r_mean) / diff_sd
*                 ! Outlier thresholds already enforced by posterior forward operators
*                 !do_adapt_inf_update = (outlier_ratio <= 3.0_r8)
*                 ! Always update the posterior, no matter what the group outlier_ratio
*                 do_adapt_inf_update = .true.
               endif

The proposed fixed code is above. The four lines with a preceding asterisk indicate the change. Obviously, more code could be removed here but I want the footprint to be absolutely minimum.

### Version of DART     
v9.11.12-1-gabff04a93 

Yes but haven't yet made a pull request.

jlaucar:
…ince

the outlier threshold has already been checked when computing forward
operators. Note that the removed test could have an impact when using
groups with posterior inflation since the test was for the group subset
of the ensemble here.

## Description:
When using groups with posterior spatially-varying adaptive inflation, an outlier threshold test was done on each group during the adaptive inflation computation. This test had a hard-coded value of 3. An outlier test is already done for the entire ensemble when the posterior forward operators are computed, using the namelist outlier threshold. When the test for the whole ensemble was passed, but the additional test for the group failed, the information from the failed group was still used for subsequent groups which was clearly inconsistent with the original intent. This fix removes the additional outlier threshold check on the groups. This can change answers when using posterior inflation with more than one group. 

A single line of executable code is modified:
                  do_adapt_inf_update = (outlier_ratio <= 3.0_r8).   Old code
                  do_adapt_inf_update = .true.                                       New code

### Fixes issue
fixes #310

### Types of changes

- [ ] Documentation update

### Tests
The update was run with a suite of options with Lorenz-96 to check that made no changes to results unless run with adaptive inflation with two or more groups. Results did change when using two groups and posterior adaptive inflation, but with a namelist outlier threshold of 3, only three out of 41000 observations assimilated resulted in a change. Results were qualitatively unchanged even with posterior adaptive inflation and two groups.

## Checklist for merging

- [x] Updated changelog entry
- [ ] Documentation updated
- [x] Version tag 

## Testing Datasets

- [x] No dataset needed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,8 +24,8 @@ The changes are now listed with the most recent at the top.
 
 **November 22 2021 :: Bug fix for groups with posterior spatially-varying adaptive inflation. Tag: v9.12.13**
 
-- Removed the additional outlier threshold check for each group when using posterior 
-  spatially-varying adaptive inflation. The outlier threshold is done for the entire ensemble
+- Removed the additional outlier threshold test for each group when using posterior 
+  spatially-varying adaptive inflation. The outlier test is done for the entire ensemble
   when the posterior forward operators are computed.
 
 **October 27 2021 :: Observation converter documentation update. Tag: v9.11.12**
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,12 @@ individual files.
 
 The changes are now listed with the most recent at the top.
 
+**November 22 2021 :: Bug fix for groups with posterior spatially-varying adaptive inflation. Tag: v9.12.13**
+
+- Removed the additional outlier threshold check for each group when using posterior 
+  spatially-varying adaptive inflation. The outlier threshold is done for the entire ensemble
+  when the posterior forward operators are computed.
+
 **October 27 2021 :: Observation converter documentation update. Tag: v9.11.12**
 
 - Improved documentation for radar observation converters
--- a/assimilation_code/modules/assimilation/assim_tools_mod.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.f90
@@ -1133,7 +1133,10 @@ SEQUENTIAL_OBS: do i = 1, obs_ens_handle%num_vars
                diff_sd = sqrt(obs_err_var + r_var)
                if (diff_sd > 0.0_r8) then
                   outlier_ratio = abs(obs(1) - r_mean) / diff_sd
-                  do_adapt_inf_update = (outlier_ratio <= 3.0_r8)
+                  ! Outlier thresholds already enforced by posterior forward operators
+                  !do_adapt_inf_update = (outlier_ratio <= 3.0_r8)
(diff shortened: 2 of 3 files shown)


hkershaw-brown (commented):
Can you strip out do_adapt_inf_update completely? It is always true. 

```diff
-! Update adaptive values if posterior outlier_ratio test doesn't fail.
-! Match code in obs_space_diags() in filter.f90
-do_adapt_inf_update = .true.
-if (inflate_only) then
-   diff_sd = sqrt(obs_err_var + r_var)
-   if (diff_sd > 0.0_r8) then
-      outlier_ratio = abs(obs(1) - r_mean) / diff_sd
-      ! Outlier thresholds already enforced by posterior forward operators
-      !do_adapt_inf_update = (outlier_ratio <= 3.0_r8)
-      ! Always update the posterior, no matter what the group outlier_ratio
-      do_adapt_inf_update = .true.
-   endif
-endif
-if (do_adapt_inf_update) then
-   ens_handle%copies(ENS_INF_COPY, state_index) = varying_ss_inflate
-   ens_handle%copies(ENS_INF_SD_COPY, state_index) = varying_ss_inflate_sd
-endif
+ens_handle%copies(ENS_INF_COPY, state_index) = varying_ss_inflate
+ens_handle%copies(ENS_INF_SD_COPY, state_index) = varying_ss_inflate_sd
```

similarly `outlier ratio` and `diff_sd` are calculated but not used.

mgharamti (commented):
Jeff, I'm happy with the change. I see Helen's comment and I vote to keep this chunk of the code although it's not very useful in this current form. My idea is to revisit our approach to posterior inflation in the future and this outlier check is important to keep in mind. I just would like to keep the necessary pieces in place by the time I take another stab at the algorithm.

mgharamti (approved):

hkershaw-brown:
> Jeff, I'm happy with the change. I see Helen's comment and I vote to keep this chunk of the code although it's not very useful in this current form. My idea is to revisit our approach to posterior inflation in the future and this outlier check is important to keep in mind. I just would like to keep the necessary pieces in place by the time I take another stab at the algorithm.

works for me

hkershaw-brown (approved):

Issue #253, opened 2021-06-29
Updating snow in CLM


timhoar:
### Use case  
Assimilation experiments that update the snow in CLM require a partitioning function that redistributes the
posterior (diagnostic) SWE into the (prognostic) snow variables - which have an arbitrary number of layers.
Most snow observations are of snowcover fraction or snow water equivalent (SWE). It is useful to provide these
diagnostic quantities (frac_sno and H2OSNO, respectively) from the CLM history files as part of the DART vector.
However, updating each layer of the prognostic variables (H2OSOI_LIQ, H2OSOI_ICE) directly through the ensemble
correlation may not result in a posterior SWE that matches the posterior SWE in H2OSNO. 

Note that we have never attempted to create snow from scratch. Nor have we ever attempted to change things like grain size or the amount of black carbon. Also, and perhaps most importantly, if one ensemble member did not have (unobserved) snow-related quantities, the (unobserved) snow-related values were left unchanged in ALL ensemble members. This is different than a failed forward operator. The forward operator can calculate a meaningful value if there is snow in the gridcell, but when the state is being updated - anything in the cutoff radius is considered.   (It may be clearer to state that if there were only 79 values on the unobserved axis (think snow ice in layer 5) and 80 estimates of the observed SWE, the regression is ill-defined.)

### Is your feature request related to a problem?  

This is really a feature request, but if there is a problem, it is that updating the prognostic variables directly does not result in a SWE consistent with the observations.

### Describe the your prefered solution 

Maintain the original ratios of snow in each active layer, but modify the amount of snow in each layer such that the
resulting SWE matches the desired posterior SWE. The most logical place to do this  is in ``dart_to_clm`` since it
reads both the posterior variables from DART and the original variables from CLM (which are needed to calculate the original ratios of snow in the layers). This removes the consideration that each ensemble member may have a different number of active snow layers in each column.

Eroding the snow is easy. Increasing snow is harder. Creating snow from scratch is impossible.

### Describe any alternatives you have considered  
Ideally, I suspect it is most appropriate to modify the atmospheric forcing to provide new snow.

timhoar:
This is awful, but I found a prototype update_snow() routine in an ancient version of the CLM model_mod (circa 2015).  This was committed to the classic_Tb branch (brightness temperature) and has extra routines in it for that. The git revision is commit 78007e2b86f25fc1d3bf423d46fc4abcca5cb94d

[...]

PR #313, opened 2021-11-22
Add SWE repartitioning for CLM

braczka:
## Description:

This adds the capability to repartition snow layer mass and dimensional variables.  The repartitioning assures that the adjustments applied to the snow layers are consistent (mass and dimensions are conserved) with the adjustment of the total snow water equivalent.  This also resolves the weakness where if a snow layer does not exist for all ensemble members, it is not adjusted for all ensemble members.  We allow the user to choose the SWE repartitioning approach within the ``dart_to_clm`` namelist within ``SWE_repartition``.  A value of '0' provides the previous, default behavior of CLM-DART, a value of '1' repartitions snow to all active snow layers, and a value of '2' repartitions snow to the bottom snow layer only.  

This ``SWE_repartition_fix`` branch began from ``clm_main`` at ``commit 337686f ``(8/27/21).  Concurrent changes of the ``clm_main`` branch have occurred during this time by @hkershaw-brown.  I made an attempt to merge the ``SWE_repartition_fix`` with the current ``clm_main`` at ``commit a741218`` (10/19).    This led to expected behavior for all variables except for the clm variable ``T_SOISNO`` (snow/soil temperature) which was updated to unrealistic large values for certain snow layers. This led to mass balance errors in CLM.  At the suggestion of Helen I reverted the changes to ``SWE_repartition_fix`` to the last working version before that attempted merge with ``clm_main``.  File changes directly related to adding new SWE repartitioning capability as part of this PR include:

`~/models/clm/dart_to_clm.f90`
`~/models/clm/dart_to_clm.rst`
`~/models/clm/readme.rst`
`~/models/clm/dart_to_clm.nml`
`~/models/clm/work/input.nml`
`~/models/clm/shell_scripts/cesm2.2/assimilate.csh`

### Fixes issue

This directly addresses github issue #253 , updating snow in CLM.

### Types of changes

- [x] Documentation update

### Tests

Performed perfect data experiment of CLM5-DART across global domain with daily assimilation across 7 day period.  See case ``/glade/u/home/bmraczka/cases/cesm2.2.0/clm5_SWE1_fix`` as an example of setup including setup scripts of ``CLM5_setup_assimilation.original`` and ``input.nml``.  Performed this test for all 3 SWE repartitioning options (`clm5_SWE0_fix`, `clm5_SWE1_fix`, `clm5_SWE2_fix`) that demonstrated: 1) the default implementation (no repartitioning) provides identical bitwise behavior to previous working `clm_main` version, 2) implementation of repartitioning schemes were mathematically sound (mass/dimensions of snow layers conserved), and 3) Update of CLM restart file is compatible with all internal CLM checks.

I intend to perform further scientific testing of repartitioning script across Western United States domain for Utah
NASA CMS collaboration -- mainly intended to explore impact of SWE repartition options upon soil moisture and surface albedo impacts.

## Checklist for merging

- [ ] Updated changelog entry
- [x] Documentation updated
- [ ] Version tag 

Additional changes have to occur before this PR is ready to merge.

## Testing Datasets

- [ ] No dataset needed

For an example set of CLM files  for testing dart_to_clm functionality see:  `/glade/scratch/bmraczka/cesm2.2.0/clm5_SWE_PR/run`

   `dart_to_clm_input_file  = 'clm2_0001.r.2011-01-02-00000.nc.original'`
   `dart_to_clm_output_file = 'clm5_SWE_PR.clm2_0001.r.2011-01-02-00000.nc.original'`
   `repartition_vhist_file = 'clm5_SWE_PR.clm2_0001.h2.2011-01-02-00000.nc'`
   `repartition_analysis_file = 'clm_analysis_member_0001_d03.2011-01-02-00000.nc'`


assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -308,8 +305,7 @@ module filter_mod
    output_mean,                  &
    output_sd,                    &
    write_all_stages_at_end,      &
-   write_obs_every_cycle,        & 
-   allow_missing_clm
+   write_obs_every_cycle
```
nancycollins, line 308: personally i love this change.  i'm just checking that this has been discussed in a software or group meeting?  in the past we've worked really hard to avoid non-backwards compatible namelist changes.   honestly, it probably complicates the code more than is worthwhile.  

but we have left in obsolete items for a few releases before removing them.  we add code to check to see if someone is setting the namelist value and print a 'this item is obsolete, we are ignoring it' message.   

if someone has this in their filter namelist, the current namelist read code will print out a fatal error saying this item is unrecognized.  if the release notes say this is being removed and is a non-backwards compatible change, i'm good with it.  i'm just checking what the current rules are here.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -916,8 +954,7 @@ subroutine read_transpose_single_task(state_ens_handle, name_handle, domain, dar
    iend = istart + block_size -1
 
    if (query_read_copy(name_handle, copy)) then
-      call read_variables(ncfile, vector, 1, get_num_variables(domain), domain)
-      ! close netcdf file
+      call read_variables(ncfile, vector, start_var, get_num_variables(domain), domain)
```
nancycollins, line 957: why was the code starting at 1 before and now respects start_var?  is this going to break something?  it is counting on all the places this is called from to have the right start_var value, which it apparently didn't matter before.  

anything in this file is more than just CLM changes - this is read/write for all of dart, every model.  i'd like to understand this change better before agreeing to it.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -833,20 +848,27 @@ subroutine read_variables(ncfile_in, var_block, start_var, end_var, domain)
...
+character(len=NF90_MAX_NAME) :: varname
+
+logical :: allow_missing
+logical :: failure
 
-logical :: missing_possible
+failure = .false.
 
-missing_possible = get_missing_ok_status()
+allow_missing = get_missing_ok_status()
```
nancycollins, line 865: ok, i'm confused.  are missing values allowed or no?

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -1556,12 +1572,14 @@ subroutine write_variables(ncid, var_block, start_var, end_var, domain, &
       dims = get_io_dim_lengths(domain, i)
 !>@todo FIXME, the first variable in the second domain is not found when using coamps_nest.
       ret = nf90_inq_varid(ncid, trim(get_variable_name(domain, i)), var_id)
-      call nc_check(ret, 'write_variables:', 'nf90_inq_varid "'//trim(get_variable_name(domain,i))//'"')
+      write(msgstring,*) 'nf90_inq_varid "'//trim(get_variable_name(domain,i))//'"'
+      call nc_check(ret, 'write_variables:', msgstring)
 
-      if (missing_possible) call set_model_missing_value(var_block(istart:iend), domain, i)
+      if (allow_missing) call set_model_missing_value(domain, i, var_block(istart:iend))
```
nancycollins, line 1578: this is only changing the order of the arguments to this routine?  that is a very error-prone change and unless there is a good reason i'm not a fan.  this is an internal-only, non-public routine so you can should be able to verify that all the calls have been correctly changed, but in most situations i'd downvote an argument-order shuffle.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -1619,7 +1637,7 @@ function create_and_open_state_output(name_handle, dom_id, copy_number, &
 call nc_check(ret, routine, 'nf90_create "'//trim(filename)//'"')
 
 ret = nf90_enddef(ncfile_out)
-call nc_check(ret, routine, 'end define mode')
+call nc_check(ret, routine, 'end define mode before writing grid info')
```
nancycollins, line 1640: some models don't have a 'grid' per-se.   nc_write_model_atts() calls model_mod code and it can write any number of variables into the netcdf file, including elevation, masks, etc - any non-state fields that it needs.   i don't think this error message change is accurate.  i'd also update the comment on line 1642 to say something like 'write additional model-specific information' because it's typically more than grid info.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -3003,17 +3024,16 @@ end function find_start_point
 !> replace the netCDF missing_value or _FillValue with
 !> the DART missing value.
 
-subroutine set_dart_missing_value(array, domain, variable)
+subroutine set_dart_missing_value(domain, variable, array)
```
nancycollins, line 3027: i'm going to object to this change as risky for limited benefit.  (but feel free to downvote me here.)

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -3030,17 +3050,33 @@ subroutine set_dart_missing_value(array, domain, variable)
...
+endif
+
 end subroutine set_dart_missing_value
 
 !--------------------------------------------------------
 !> replace the DART missing value code with the 
 !> original netCDF missing_value (or _FillValue) value.
 
-subroutine set_model_missing_value(array, domain, variable)
+subroutine set_model_missing_value(domain, variable, array)
```
nancycollins, line 3075: ditto here.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -187,7 +188,7 @@ module io_filenames_mod
 
 end type
 
-character(len=512) :: msgstring
+character(len=512) :: string1, string2, string3
```
nancycollins, line 191: i'd prefer to leave msgstring alone and add string2 and string 3, to be honest.  trying to look at differences which are just variable renames makes it easy to overlook errors in editing, or masks real code changes.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -632,24 +640,39 @@ subroutine check_attributes(ncFile, filename, ncVarId, domid, varid)
...
          call get_missing_value(domid, varid, spvalR8)
          call check_attribute_value_r8(ncFile, filename, ncVarID, 'missing_value', spvalR8)
 
       case default
+         !>@todo FIXME report the variable with the unsupported xtype
+         call error_handler(E_ERR, 'check_attributes', 'unknown xtype', source)
+   end select
+endif
+
+if ( get_has_FillValue(   domid, varid) ) then
```
nancycollins, line 660: are these routines used in model_to_dart and dart_to_model?   because i thought the idea was to get away from having to support either missing values or fill values.  this looks like it's adding more code to support them.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -690,21 +717,23 @@ end subroutine check_attribute_value_int
 !> check r4 values are the same
 
 
-subroutine check_attribute_value_r4(ncFile, filename, ncVarID, att_string, spvalR4)
+subroutine check_attribute_value_r4(ncFile, filename, ncVarID, att_string, att_valR4)
```
nancycollins, line 720: i honestly don't believe that we have to have int and r4 and r8 versions of all these routines.  i was hoping they'd all go away so i haven't said much before.  but we use the default fortran conversion routines to go from what's in the netcdf files to our real(r8) arrays, so that should do the same for comparisons/tests of missing values or fill values using the fortran intrinsics.

assimilation_code/modules/io/state_structure_mod.f90
```diff
@@ -765,114 +769,170 @@ subroutine load_common_cf_conventions(domain)
...
       case ( NF90_INT )
-          if (nf90_get_att(ncid, NF90_GLOBAL, '_FillValue', cf_spvalINT) == NF90_NOERR) then
-             domain%variable(ivar)%io_info%spvalINT       = cf_spvalINT
-             domain%variable(ivar)%io_info%has_missing_value = .true.
+          ! Sometimes the attributes are specified as GLOBAL attributes
+          ret = nf90_get_att(ncid, NF90_GLOBAL, '_FillValue',    cf_spvalINT)
+          if (ret == NF90_NOERR) then
+             domain%variable(ivar)%io_info%spvalINT            = cf_spvalINT
+             domain%variable(ivar)%io_info%spvalR8             = real(cf_spvalINT,r8)
+             domain%variable(ivar)%io_info%has_FillValue       = .true.
```
nancycollins, line 808: can we take care of this in the converter programs and not add more code to the overall dart i/o system?  just asking.

models/clm/work/input.nml
```diff
@@ -306,21 +368,21 @@
...
+   quantity_of_interest  = 'QTY_SOIL_TEMPERATURE'
    interp_test_vertcoord = 'VERTISHEIGHT'
    interp_test_lonrange  = 320.0, 90.0
    interp_test_dlon      = 1.0
    interp_test_latrange  = 0.0, 75.0
    interp_test_dlat      = 1.0
    interp_test_vertrange = 1.0, 2.0
    interp_test_dvert     = 1.0
-   verbose               = .false.
+   verbose               = .true.
```
nancycollins, line 385: i suspect you don't want to leave verbose as true.  i think it prints out the entire state location info if this is true?  or maybe it does that anyway for some tests.  if this is helpful, leave it in.

models/clm/shell_scripts/cesm2_2/datm.streams.txt.CPLHISTForcing.Solar_complete
```diff
@@ -0,0 +1,76 @@
+<streamstemplate>
```
nancycollins, line 1: to be honest, i'm trusting you on the forcing files and scripts and nml changes here.

nancycollins (commented):
i've reviewed everything *EXCEPT* clm_to_dart, dart_to_clm and the model_mod.  the changes there are just so extensive i'm at a bit of a loss to know if there's anything useful i can contribute there.  if the code works and is tested i can skim them over and look for anything that seems unwise but i won't be catching any logic errors.

i did a much more through review of the changes that are outside of the clm-only code.   i hope my tone isn't too cranky - i know some of these changes were inherited from before.  i'm dubious some of them are really needed.  code that is called for every model we support is higher risk to change since introducing bugs there (or non-backwards-compatible changes) could break other models in subtle ways.  it's a bigger risk and therefore should require a better justification (and/or extensive testing with as many models as possible) before changing it. 

i'm open to discussion either here or a video call if that's better.  and apologies if some of these changes have been discussed already in group meetings and i'm out of the loop.

hkershaw-brown:
Note, Brett is preparing a CLM tutorial so we tagged a pre-release for CLM with SWE: https://github.com/NCAR/DART/releases/tag/clm-swe_pre-release  that can be used for the tutorial.   `git checkout clm-swe_pre-release`

The goal on this pull request is to get SWE_repartition_fix upto date with main, SWE correct, and merged into main. 

Note on clm_main:  I had a go at bringing clm_main up to date (it was last updated from main in June) and removing the non-essential changes.  To get a flavor of my commits to clm_main, (getting clm_main upto date with main and reverting any changes not related to clm) see below.  They are along the lines of Nancy's commits in the review of this pull request.   Second time lucky, we do this clean up on SWE_repartition_fix without breaking the new code. 

```
229c3771b clm_main@{1}: commit: remove swe from namelist
bea62478a clm_main@{2}: commit: remove repartition_swe because the repartitioning is not implemented
bc85f0ae1 clm_main@{3}: commit: add QTY_SOIL_MATRIC_POTENTIAL to default quantities
fe41246c2 clm_main@{4}: commit: move get_clos_obs declation to "in other modules" section
8d4a0a3aa clm_main@{5}: commit: missed routine in last commit
118e41aaf clm_main@{6}: commit: remove non-fill value changes
cbcb31bd7 clm_main@{7}: commit: mmc_clm_specific.f90 is not in any path_names_files
6a54f7041 clm_main@{8}: commit: get_close_obs pass-through
64204b955 clm_main@{9}: commit: allow_missing_clm controlled through filter_mod namelist
d24924cc6 clm_main@{10}: commit: revert non-fill value changes
be550f3a5 clm_main@{11}: commit: keep allow_missing_clm as namelist option in filter_mod
263d394ce clm_main@{12}: merge main: Merge made by the 'recursive' strategy.
b1248218c clm_main@{13}: commit: revert changes that are not for fillValue
99c22e417 clm_main@{14}: commit: first batch of reverting non-essential changes
28762930c clm_main@{15}: branch: Created from refs/remotes/origin/clm_main
```

assimilation_code/modules/assimilation/filter_mod.f90
hkershaw-brown: reverted filter_mod.f90  match main.

hkershaw-brown:
I don't think the fill value code in this pull request works in general.  

from netcdf to state vector:

  netcdf fill_value -> MISSING_R8
  netcdf missing_value -> MISSING_R8

from state vector to netcdf:

   MISSING_R8 ->  ?
   how can you tell whether this is fill or missing?

If a variable has a missing value _and_ a fill value, the code as it is in this pull request will replace all missing_r8 with the missing value.  So you have converted all fill values to missing values.See below code in set_model_missing_value. 

```diff
--------------------------------------------------------
!> replace the DART missing value code with the 
!> original netCDF missing_value (or _FillValue) value.
subroutine set_model_missing_value(domain, variable, array)

integer,  intent(in)    :: domain
integer,  intent(in)    :: variable
real(r8), intent(inout) :: array(:)

integer        :: model_missing_valueINT
real(r4)       :: model_missing_valueR4
real(digits12) :: model_missing_valueR8

! check to see if variable has missing value attributes
if ( get_has_missing_value(domain, variable) ) then

   select case ( get_xtype(domain, variable) )
      case ( NF90_INT )
         call get_missing_value(domain, variable, model_missing_valueINT)
         where(array == MISSING_R8)       array = model_missing_valueINT
      case ( NF90_FLOAT )
         call get_missing_value(domain, variable, model_missing_valueR4)
         where(array == MISSING_R8)       array = model_missing_valueR4
      case ( NF90_DOUBLE )
         call get_missing_value(domain, variable, model_missing_valueR8)
         where(array == MISSING_R8)       array = model_missing_valueR8
   end select

endif

if ( get_has_FillValue(domain, variable) ) then
+ If (get_has_missing_value(domain, variable) == .true. ) then all the missing_r8 in the state have been replaced already. 
+ so you've converted any fill values to missing values.
   select case ( get_xtype(domain, variable) )
      case ( NF90_INT )
         call get_FillValue(domain, variable, model_missing_valueINT)
         where(array == MISSING_R8)   array = model_missing_valueINT
      case ( NF90_FLOAT )
         call get_FillValue(domain, variable, model_missing_valueR4)
         where(array == MISSING_R8)   array = model_missing_valueR4
      case ( NF90_DOUBLE )
         call get_FillValue(domain, variable, model_missing_valueR8)
         where(array == MISSING_R8)   array = model_missing_valueR8
   end select

endif
```

clm_to_dart is replacing indeterminate values with FillValue, why not replace indeterminate values with missing values?  Then you don't need to add all the FillValue code to direct_netcdf_mod.f90.

```diff
155          ! Prevent unused snow layers from being updated by the assimilation.
                ! Unused snow layers have indeterminate values. The indeterminate values are replaced with 
                ! FillValue which prevents filter from updating the unused snow layers during the assimilation.
                do i = 1, nlevsno - numsnowlevels  ! loop over unused layers

                    
                     ! trace amounts of snow are in the level closest to the ground
                     ! frac_sno(j) seems to be a reliable indicator of a trace of snow
                     if (frac_sno(j) > 0.0_r8 .and. i == nlevsno) cycle
                       
-                     variable(i,j) = FillValue
+                     variable(i,j) = missingValue
                enddo

         enddo
```

assimilation_code/modules/io/io_filenames_mod.f90
hkershaw-brown: yeah I'm reverting the string changes.  I think there are very few actual changes in this file once you take out the variable name changes.

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -632,24 +640,39 @@ subroutine check_attributes(ncFile, filename, ncVarId, domid, varid)
...
+         !>@todo FIXME report the variable with the unsupported xtype
+         call error_handler(E_ERR, 'check_attributes', 'unknown xtype', source)
+   end select
+endif
+
+if ( get_has_FillValue(   domid, varid) ) then
+   select case (get_xtype(domid, varid))
+      case (NF90_INT)
+         call get_FillValue(domid, varid, spvalINT)
+         call check_attribute_value_int(ncFile, filename, ncVarID, 'FillValue', spvalINT)
```
hkershaw-brown, line 664: why is this 'FillValue' and not '_FillValue'?

hkershaw-brown: There is no error check in check_attribute_value if nf90_get_att returns NF90_NOERR, so I think this code has merrily continues because there is no attribute called 'FillValue'

hkershaw-brown: fixed this in 72c1d2de08fc16f70fcfc816fd6f19aa6f05fc8c

assimilation_code/modules/io/state_structure_mod.f90
```diff
@@ -765,114 +769,170 @@ subroutine load_common_cf_conventions(domain)
...
           endif
 
       case ( NF90_FLOAT )
-          if (nf90_get_att(ncid, NF90_GLOBAL, '_FillValue', cf_spvalR4) == NF90_NOERR) then
-             domain%variable(ivar)%io_info%spvalR4        = cf_spvalR4
-             domain%variable(ivar)%io_info%has_missing_value = .true.
+          ret = nf90_get_att(ncid, NF90_GLOBAL, '_FillValue',    cf_spvalR4)
+          if (ret == NF90_NOERR) then
+             domain%variable(ivar)%io_info%spvalR4             = cf_spvalR4
+             domain%variable(ivar)%io_info%spvalR8             = real(cf_spvalR4,r8)
```
hkershaw-brown, line 834: why does float need r8 set?

assimilation_code/modules/io/state_structure_mod.f90
```diff
@@ -765,114 +769,170 @@ subroutine load_common_cf_conventions(domain)
...
-         call error_handler(E_ERR, 'load_common_cf_conventions',string1,source)
+         call error_handler(E_ERR,routine,string1,source)
    end select
 
-   !>@todo FIXME : Not using scale factor or offset at the moment. Need to
+   ! If the variable has one or the other, no problem.
+   ! If the variable has both _FillValue and missing_value attributes, the
+   ! values must be the same or we are lost. DART only supports one missing
+   ! value code. When we go to write, we have no way of knowing which value
+   ! to use as a replacement for the DART missing code.
```
hkershaw-brown, line 897: .

braczka:
_"clm_to_dart is replacing indeterminate values with FillValue, why not replace indeterminate values with missing values? Then you don't need to add all the FillValue code to direct_netcdf_mod.f90."_

That's a good point Helen-- its not clear to me what benefit there is by assigning an indeterminate value as a `FillValue` vs. a `missingValue`, just as long as the original indeterminate value is replaced within the model state during  `dart_to_clm`.  I am swamped by tutorial preparation right now, but will take a closer look and maybe test this later this week.

braczka:
Tested my local branch with replacement of indeterminate variable with missingValue instead of FillValue as proposed by Helen (as shown below).   This reproduces the same functionality as before, therefore I included this change in latest commit: `Using missing instead of FillValue for indeterminate 3ad1273`        

```
155          ! Prevent unused snow layers from being updated by the assimilation.
                ! Unused snow layers have indeterminate values. The indeterminate values are replaced with 
                ! FillValue which prevents filter from updating the unused snow layers during the assimilation.
                do i = 1, nlevsno - numsnowlevels  ! loop over unused layers

                    
                     ! trace amounts of snow are in the level closest to the ground
                     ! frac_sno(j) seems to be a reliable indicator of a trace of snow
                     if (frac_sno(j) > 0.0_r8 .and. i == nlevsno) cycle
                       
-                     variable(i,j) = FillValue
+                     variable(i,j) = missingValue
                enddo

         enddo
                                                                                                                                       `

**Comment by hkershaw-brown** [comment]
> Tested my local branch with replacement of indeterminate variable with missingValue instead of FillValue as proposed by Helen (as shown below). This reproduces the same functionality as before, therefore I included this change in latest commit: `Using missing instead of FillValue for indeterminate 3ad1273`
> 
> ```
> 155          ! Prevent unused snow layers from being updated by the assimilation.
>                 ! Unused snow layers have indeterminate values. The indeterminate values are replaced with 
>                 ! FillValue which prevents filter from updating the unused snow layers during the assimilation.
>                 do i = 1, nlevsno - numsnowlevels  ! loop over unused layers
> 
>                     
>                      ! trace amounts of snow are in the level closest to the ground
>                      ! frac_sno(j) seems to be a reliable indicator of a trace of snow
>                      if (frac_sno(j) > 0.0_r8 .and. i == nlevsno) cycle
>                        
> -                     variable(i,j) = FillValue
> +                     variable(i,j) = missingValue
>                 enddo
> 
>          enddo
>                                                                                                                                        `     
> ```

very nice

**Comment by braczka** [comment]
I should clarify I tested only the small incremental change within commit `3ad1273`, not the`SWE_repartition_fix` branch in its current form.  Figured we were not quite ready for that yet....

**Comment by hkershaw-brown** [comment]
side question about naming conventions. Why do we use `spvalINT` rather than calling the variable `fillINT`?
Is spval a Netcdf term?

```
state_structure_mod.f90:
   integer  :: missingINT   = MISSING_I   ! missing values
   integer  :: spvalINT     = MISSING_I   ! fill values
```

**Comment by hkershaw-brown** [comment]
Note on testing this pull request against [clm-swe_pre-release](https://github.com/NCAR/DART/releases/tag/clm-swe_pre-release)


This pull request (SWE_repartition_fix branch) uses the main branch "allow_missing" logic so requires that you use the following option in filter_nml:

```
&filter_nml
allow_missing_clm = .true.
```

File: assimilation_code/location/threed_sphere/location_mod.f90 [hunk 17]
```diff
@@ -797,19 +803,23 @@ subroutine write_location(locfile, loc, fform, charstring)
...
+
+!> some land models has vertical levels specified in cm ... 
+!> they need many digits when printing (in km) 
+
 select case  (loc%which_vert)
    case (VERTISUNDEF)
       write(charstring, '(A,A)')       trim(string1), '              Undefined'
    case (VERTISSURFACE)
-      write(charstring, '(A,F13.5,A)') trim(string1), loc%vloc, ' surface (m)'
+      write(charstring, '(A,F18.10,A)') trim(string1), loc%vloc, ' surface (m)'
```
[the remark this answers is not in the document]
hkershaw-brown: @nancycollins do you think it is best to just revert location_mod.f90 to main?

models/clm/dart_to_clm.f90
```diff
@@ -1,112 +1,919 @@
...
+   else
+      write(string1, *) 'no support for data array of dimension ', rank
+      call error_handler(E_ERR, source, string1, source)
+   endif
+
+enddo UPDATE
+
+ ! Manually repartition snow layer variables with H2OSNO variable
+
+  if (repartition_swe > 0.0_r8) then
```
hkershaw-brown, line 190: repartition_swe is an integer, so you can use 0. 

```suggestion
  if (repartition_swe > 0) then
```

models/clm/dart_to_clm.f90
```diff
@@ -1,112 +1,919 @@
...
+
+
+dom_restart = 1
+
+UPDATE : do ivar=1, get_num_variables(dom_restart)
+  
+  ! If repartitioning SWE is enabled, then skip applying snow variable updates
+  ! within UPDATE loop which applies un-repartitioned DART posterior
+  ! The repartitioned snow variable updates will be performed in 'update_snow'
+  if (repartition_swe > 0.0_r8) then
```
hkershaw-brown, line 157:
```suggestion
  if (repartition_swe > 0) then
```

assimilation_code/location/threed_sphere/location_mod.f90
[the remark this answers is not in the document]
nancycollins: yes.  i'm concerned about unintentional code breakage in the calling code (passing in buffers that are now too short).

braczka:
My latest batch of commits are associated with the observation converters mostly used with clm, that never got migrated to the the DART repository during the prior clm_update PRs.  These commits include soil moisture and leaf area observations converters coming from NASA (NASA_Earthdata folder) and NSIDC (NSIDC folder).  I updated the links to which acquire the raw netcdf or hdf5 data files, tested functionality, and created documentation for each, which was either sparse or non-existent.    Changes to HDF5_utilities_mod.f90, obs_seq_utilities.f90, EASE_utilities_mod.f90 were needed to support the obs converters within the NASA_Earthdata and NSIDC folders.

observations/obs_converters/NASA_Earthdata/README.rst
```diff
@@ -0,0 +1,254 @@
+PROGRAMS ``LPRM_L3_to_obs.f90`` ``AMSR_E_L2_to_obs.f90`` ``SMAP_L2_to_obs``
```
hkershaw-brown, line 1:
```suggestion
PROGRAMS ``LPRM_L3_to_obs.f90`` ``AMSR_E_L2_to_obs.f90``
 ```

The SMAP_L2_to_obs is the NISDC directory

**Comment by braczka** [comment]
@hkershaw-brown  I re-tested this current version of the branch for all 3 snow re-partitioning schemes and for the most part it looks pretty good, after I added the `allow_missing_clm=.true.` option within `&filter_nml`.    I am not sure if it is a concern that it is not bit-by-bit identical to a previous version that I have run for the `CLM5_setup_assimilation` toy script caserun.  It starts out identical for the first couple assimilation time steps, then starts to deviate for all snow adjusted CLM variables.   Did we add/remove some precision in the DART scripting with these recent updates?  Not sure this is expected behavior, and realize sometimes getting bit-by-bit reproducibility can be messy.......

**Comment by hkershaw-brown** [comment]
> @hkershaw-brown I re-tested this current version of the branch for all 3 snow re-partitioning schemes and for the most part it looks pretty good, after I added the `allow_missing_clm=.true.` option within `&filter_nml`. I am not sure if it is a concern that it is not bit-by-bit identical to a previous version that I have run for the `CLM5_setup_assimilation` toy script caserun. It starts out identical for the first couple assimilation time steps, then starts to deviate for all snow adjusted CLM variables. Did we add/remove some precision in the DART scripting with these recent updates? Not sure this is expected behavior, and realize sometimes getting bit-by-bit reproducibility can be messy.......

Let's check this out.  

I'm not sure if CLM is always bit-for-bit idenical, but DART should be. 
There is a small bug fix that is not in the clm-swe_pre-release tag, which is  `io_filenames_mod.f90` should be checking `_FillValue` not `FillValue`.

```diff
diff --git a/assimilation_code/modules/io/io_filenames_mod.f90 b/assimilation_code/modules/io/io_filenames_mod.f90
index a3a777a15..632cbb3d6 100644
--- a/assimilation_code/modules/io/io_filenames_mod.f90
+++ b/assimilation_code/modules/io/io_filenames_mod.f90
@@ -661,15 +661,15 @@ if ( get_has_FillValue(   domid, varid) ) then
    select case (get_xtype(domid, varid))
       case (NF90_INT)
          call get_FillValue(domid, varid, spvalINT)
-         call check_attribute_value_int(ncFile, filename, ncVarID, 'FillValue', spvalINT)
+         call check_attribute_value_int(ncFile, filename, ncVarID, '_FillValue', spvalINT)
 
       case (NF90_FLOAT)
          call get_FillValue(domid, varid, spvalR4)
-         call check_attribute_value_r4(ncFile, filename, ncVarID, 'FillValue', spvalR4)
+         call check_attribute_value_r4(ncFile, filename, ncVarID, '_FillValue', spvalR4)
 
       case (NF90_DOUBLE)
          call get_FillValue(domid, varid, spvalR8)
-         call check_attribute_value_r8(ncFile, filename, ncVarID, 'FillValue', spvalR8)
+         call check_attribute_value_r8(ncFile, filename, ncVarID, '_FillValue', spvalR8)
```

I'm wondering if this is the difference between your old and new runs.  But it could be something else.
Do you mind if I grab the files you are running with?  I just need which directory they are in Cheyenne.

Cheers,
Helen

braczka:
> > @hkershaw-brown I re-tested this current version of the branch for all 3 snow re-partitioning schemes and for the most part it looks pretty good, after I added the `allow_missing_clm=.true.` option within `&filter_nml`. I am not sure if it is a concern that it is not bit-by-bit identical to a previous version that I have run for the `CLM5_setup_assimilation` toy script caserun. It starts out identical for the first couple assimilation time steps, then starts to deviate for all snow adjusted CLM variables. Did we add/remove some precision in the DART scripting with these recent updates? Not sure this is expected behavior, and realize sometimes getting bit-by-bit reproducibility can be messy.......
> 
> Let's check this out.
> 
> I'm not sure if CLM is always bit-for-bit idenical, but DART should be. There is a small bug fix that is not in the clm-swe_pre-release tag, which is `io_filenames_mod.f90` should be checking `_FillValue` not `FillValue`.
> 
> ```diff
> diff --git a/assimilation_code/modules/io/io_filenames_mod.f90 b/assimilation_code/modules/io/io_filenames_mod.f90
> index a3a777a15..632cbb3d6 100644
> --- a/assimilation_code/modules/io/io_filenames_mod.f90
> +++ b/assimilation_code/modules/io/io_filenames_mod.f90
> @@ -661,15 +661,15 @@ if ( get_has_FillValue(   domid, varid) ) then
>     select case (get_xtype(domid, varid))
>        case (NF90_INT)
>           call get_FillValue(domid, varid, spvalINT)
> -         call check_attribute_value_int(ncFile, filename, ncVarID, 'FillValue', spvalINT)
> +         call check_attribute_value_int(ncFile, filename, ncVarID, '_FillValue', spvalINT)
>  
>        case (NF90_FLOAT)
>           call get_FillValue(domid, varid, spvalR4)
> -         call check_attribute_value_r4(ncFile, filename, ncVarID, 'FillValue', spvalR4)
> +         call check_attribute_value_r4(ncFile, filename, ncVarID, '_FillValue', spvalR4)
>  
>        case (NF90_DOUBLE)
>           call get_FillValue(domid, varid, spvalR8)
> -         call check_attribute_value_r8(ncFile, filename, ncVarID, 'FillValue', spvalR8)
> +         call check_attribute_value_r8(ncFile, filename, ncVarID, '_FillValue', spvalR8)
> ```
> 
> I'm wondering if this is the difference between your old and new runs. But it could be something else. Do you mind if I grab the files you are running with? I just need which directory they are in Cheyenne.
> 
> Cheers, Helen

Sure -- I am currently running with the files within `/glade/work/bmraczka/DART`.   This should be identical, (nearly identical) to the upstream files in NCAR/DART branch SWE_repartition_fix.   The directories for the newer case runs are located at `/glade/u/home/bmraczka/cases/cesm2.2.0/clm5_SWE0_final`, `clm5_SWE1_final` and `clm5_SWE2_final`.  The run folders for each case respectively are located within `/glade/scratch/bmraczka/cesm2.2.0`.   I have been comparing the output files against the older cases: `clm5_SWE0_fix`, `clm5_SWE1_fix` and `clm5_SWE2_fix` for benchmarking.  These older test cases were generated with the `clm-swe_pre-release` tag.

I have diffused the netcdf restart files for each companion case (the new and older versions) for variable `H2OSOI_ICE`.  Haven't really checked the other variables, but assume the same loss/gain in precision occurs across all variables.   This issue is slightly different than previous benchmark testing with DART-CLM in which there were immediate differences with the first set (first time step) of CLM output files -- suggesting CLM was the culprit.   In this situation, the first couple of time steps (with assimilation) are identical (as far as I can tell), and differences start to creep in at the 2nd time step after the DART update.  This suggests perhaps DART is introducing the differences....?

braczka:
Odds and ends update:

The `156dca3` commit updates the syntax fix to replace 'nlon' with 'lon' with model_mod.f90.  This was not causing a problem with any of my recent assimilation run tests -- but this is indeed a typo as brought to our attention from the E3SM-DART user today.

In reference to the reproducibility of  `SWE_repartition_fix`, the slight differences in clm variables are pervasive -- not just the snow related variables that start to deviate from the tag `clm-swe_pre-release`, but all other variables that I tested as well (e.g. biomass).   The deviation in behavior always occurs for portions of the variable state that are updated by DART, suggesting a difference in the DART code, or the way it was compiled.   Below are my DART compiler options on Cheyenne:

`MPIFC = mpif90`
`MPILD = mpif90`
`FC = ifort`
`LD = ifort`

`INCS = -I$(NETCDF)/include`
`LIBS = -L$(NETCDF)/lib -lnetcdff -lnetcdf`
`FFLAGS  = -O -assume buffered_io $(INCS)`
`LDFLAGS = $(FFLAGS) $(LIBS)`

 I am going to re-run the test `clm5_setup_assimilation` with the tag `clm-swe_pre-release` one more time to see if I can match the behavior.  I also plan to re-run the CLM5-DART Western US benchmark simulation with `SWE_reparition_fix`.  This is probably overkill, but won't take too long to check.

braczka:
The CLM5-DART Western US benchmark simulation with `SWE_repartition_fix` has been completed and is nearly identical to the previous benchmark simulation using the tag `clm-swe_pre-release`.   This is ready to merge as far as I am concerned and am pulling it off draft.  

Out of curiosity, I performed some tests running CLM-DART assimilations, with CLM in 'branch' mode (bit by bit reproducible for the start of CLM runs) instead of 'hybrid'  mode which is not bit by bit.   This didn't explain some of the small differences I was seeing in reproducibility, but doesn't effect the readiness of this PR as far as I am concerned.

assimilation_code/modules/observations/obs_sequence_mod.f90
```diff
@@ -665,7 +665,7 @@ subroutine insert_obs_in_seq(seq, obs, prev_obs)
    ! Later do an increase of space and copy
    write(string1,*) 'ran out of room, num_obs (',seq%num_obs, &
                                ') > max_num_obs (',seq%max_num_obs,')'
-   call error_handler(E_ERR,'insert_obs_in_seq',string1, source)
+    call error_handler(E_ERR,'insert_obs_in_seq',string1, source)
```
hkershaw-brown, line 668:
```suggestion
   call error_handler(E_ERR,'insert_obs_in_seq',string1, source)
```

hkershaw-brown (approved):
Approved. 

I'll bring this up to date with main (it is just the inflation namelist bug-fix)and bump the version before merging.

PR #321, opened 2022-01-10
bug-fix character namelist values need to be the same type as variable

### LINKED ISSUE #245: pgi does not like the inflation flavors being integers.
pgi does not like the inflation flavors being integers. 

```
 ERROR FROM:
  source : utilities_mod.f90
  routine: check_namelist_read
  message:  INVALID NAMELIST ENTRY:    inf_flavor                  = 2,                       0, in namelist filter_nml
```

```diff
- inf_flavor                  = 2,                       0,
+ inf_flavor                 = '2',                     '0',
```

_Originally posted by @hkershaw-brown in https://github.com/NCAR/DART/issues/219#issuecomment-849177858_

hkershaw-brown:
## Description:
The nvidia compiler enforces the fortran standard that characters in namelists must have "" or ''. This pull request removes the 
"string or integer namelist" option added in pull request https://github.com/NCAR/DART/pull/154. It does not remove any of the integer parameters added to to adaptive_inflate_mod in  https://github.com/NCAR/DART/pull/154

The standard is:
Note 10.33, in 10.9.1.3

"A character sequence corresponding to a namelist input item of character data type shall be delimited either with apostrophes or quotes. The delimiter is required to avoid ambiguity between undelimited character sequences and object names. The value of the DELIM = specifier, if any, in the OPEN statement for an external file is ignored during namelist input "

(Note 10.36 in the f2008 standard).

### Fixes issue
Fixes #245 

### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
Using strings in the namelist to specify inflation flavor was added to main but not documented.  Anyone who is using strings will need to revert to using integers instead, e.g.
```diff
  &filter_mod
-  inf_flavor                  = VARYING_SS_INFLATION,                       NO_INFLATION ,
+  inf_flavor                  = 2,                       0,
/
```
- [ ] Documentation update

### Tests
* Lorenz_96 compile and run with nvfortran  (nvhpc module on Cheyenne)
* Bitwise test with intel on main. 

## Checklist for merging

- [x] Updated changelog entry
- [ ] Documentation updated
- [x] Version tag 

## Testing Datasets

- [X] No dataset needed. You can use lorenz_96

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,11 @@ individual files.
 
 The changes are now listed with the most recent at the top.
 
+**January 11 2022 :: Bug fix for inflation namelist options. Tag: v9.12.1**
+
+- reverted inf_flavor namelist option to be integer only to conform to Fortran
+  standards. 
+
 **December 7 2021 :: Refactored filter_assim. Tag: v9.12.0**
 
 - Filter_assim refactored so each process calcuates increments
--- a/assimilation_code/modules/assimilation/filter_mod.dopplerfold.f90
+++ b/assimilation_code/modules/assimilation/filter_mod.dopplerfold.f90
@@ -29,7 +29,7 @@ use time_manager_mod,      only : time_type, get_time, set_time, operator(/=), o
                                   operator(-), print_time
 
 use utilities_mod,         only : error_handler, E_ERR, E_MSG, E_DBG,                         &
-                                  logfileunit, nmlfileunit, timestamp, get_value_from_string, &
+                                  logfileunit, nmlfileunit, timestamp,                        &
                                   do_output, find_namelist_in_file, check_namelist_read,      &
                                   open_file, close_file, do_nml_file, do_nml_term, to_upper,  &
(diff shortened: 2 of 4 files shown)


nancycollins (approved):
this all looks good.  

i think eventually all inflation related namelist items merit their own namelist in the adaptive_inflate_mod module.  at that point we can use string names properly delimited with quotes to specify the various types of inflation.

jlaucar:
Just to note. There are going to be massive changes to how inflation is
controlled in conjunction with the new more powerful filter variants. It
will become necessary to control the inflation differently for different
state variable kinds (or whatever they are called) with possibly many
parameters. We have no idea how to implement this yet. As an example, I
just finished hard-coding three different methods of doing bounds
respecting inflation in adaptive_inflate_mod along with completely
hard-coded logic on how to select those given parameters of the inflation
details and the state kind.

We need to have a discussion of how we're going to cleanly access this new
info and allow much more flexible construction of new options.

PR #329, opened 2022-02-23
Updt localization

jonlabriola:
## Description:

This update adds the ability to vary the vertical localization radius for different observation types for the CM1 model. The code to do this was borrowed from ./location/threed_sphere/. I have updated the input.nml file as well to highlight that cm1 now has the ability to modify the vertical localization radius.  I have run tests with my existing datasets to ensure the updated code is fully functioning.   I added some comments to the code, readme.rst, and input.nml to highlight the updates I made.  Please let me know if I should add any additional comments to better highlight the changes made.

### Types of changes

- [ ] Documentation update

### Tests
I have run tests on my datasets to ensure the vertical localization updates do not break the code and perform as expected.  I am attaching a file that shows the difference between the old CM1 DART code (which defines horizontal and vertical localization radius as the same) and the updated CM1 DART code (which defines separate horizontal and vertical localization radii). If need be, I can provide some test cases to ensure the code works. 
[Updated_Vertical_Localization.pdf](https://github.com/NCAR/DART/files/8125069/Updated_Vertical_Localization.pdf)

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [ ] No dataset needed


hkershaw-brown:
Hi Jon, thanks for this. We'll get reviewed. 

Cheers,
Helen

hkershaw-brown (changes requested):
Hi Jon, thanks for being patient with this. 

The code to do the per type normalization looks good.  
There are a few comments I'm going to go ahead and clean up this afternoon.  This is just style that we generally remove authors initials except for explicit todos. Git blame takes care of showing who changed what. 

One thing I do want to change is to not _always_ use the vertical normalization.

```
real(r8) :: vert_normalization_height       = 10000.0_r8  ! namelist default value
```

 in `function dist_3d_sq` (and similarly `dist_3d`) you have:
```
if (allocated(per_type_vert_norm)) then
  vert_normal = separation(IZ) / per_type_vert_norm(VERTISHEIGHT, type1)
else
   vert_normal = separation(IZ)/vert_normalization(VERTISHEIGHT)
endif

val = sqrt(separation(IX)*separation(IX) + &
           separation(IY)*separation(IY) + &
           vert_normal*vert_normal )
```

whereas someone may want to have the original behavior:

```
val = sqrt(separation(IX)*separation(IX) + &
           separation(IY)*separation(IY) + &
           separation(IZ)*separation(IZ)
```

So I would add a namelist option to which switches on using the vertical_normalization. 

I thought maybe you could just make the default value for `vert_normalization_height=1.0_r8`.  However, since you can give per_type vertical normalization  (and you don't have to give every type), this won't work.  

Let me know if I'm missing something here.

Cheers,
Helen

jonlabriola:
Hi Helen, thanks for looking over everything!  It would make sense to set the default value for vert_normalization_height to 1, this would set the vertical and horizontal localization radius to the same value.   Further, this is still the case even if "per_type_vert_norm" is allocated but not all observation types are defined in the namelist.  This is because pert_type_vert_norm is initialized for all assimilated observation types with vert_normalization_height, and if a specific vertical localization value is specified, the array is then updated.

` ! Set the defaults for all specific types not listed in the special list 
     per_type_vert_norm(VERTISHEIGHT, :)      = vert_normalization_height`

hkershaw-brown:
ok great, I'll make that change and get this merged and released. 

Cheers,
Helen

hkershaw-brown (approved):

Issue #176, opened 2021-03-09
NaN processing in adaptive_inflate_mod.f90


timhoar:
We employ several different methods to test for NaNs, not sure which one is best, or if we need to be consistent.

In adaptive_inflate_mod.f90 there is code:

```
   ! Computational errors check (small numbers + NaNs)
   if (abs(density_1) <= TINY(0.0_r8) .OR. &
       abs(density_2) <= TINY(0.0_r8) .OR. &
       density_1 /= density_1 .OR. density_1 /= density_1 .OR. &
       density_2 /= density_2 .OR. density_2 /= density_2) then
      new_cov_inflate_sd = lambda_sd
      return
   endif
```
This code does the same test on density_1 - twice - and density_2 - twice - but before changing the code, I want to know what method to use. In other codes, we have employed the test
`if (var < 0.0_r8 .and. var > 0.0_r8)` to test for NaNs.

nancycollins:
i think (var /= var) is an ok test.  the duplicates aren't needed but they don't hurt anything.

that other line is not quite the right test in your note above. 
var, if NaN, always *fails* tests, so that line as written would evaluate as:
 
if (.false. .and. .false.) then nan  ! which will never be true.

it would have to be:

if ((.not. var <= 0.0_r8 ).and. ( .not. var > 0.0_r8)) then nan  
! note <= and > because otherwise var == 0 would wrongly indicate nan
 
which isn't the most direct thing to read but works.

hkershaw-brown:
@nancycollins  (sanity) check this out:  (NaN /= NaN) is false  for pgi/nvfortran unless you compile with the flag -Kieee 

```
hkershaw@cheyenne5:$ cat test_nan.f90 
program test_nan

implicit none

real :: a, x

x = -1
a = sqrt(x)

print*, a

print*, a /= a
print*, a == a

if (a /= a) then
  print*, "you have a NaN"
endif

end program test_nan
hkershaw@cheyenne5:$ gfortran test_nan.f90 
hkershaw@cheyenne5:$ ./a.out 
              NaN 
 T
 F
 you have a NaN 
hkershaw@cheyenne5:$ nvfortran test_nan.f90 
hkershaw@cheyenne5:$ ./a.out 
             NaN 
  F
  T
hkershaw@cheyenne5:$ nvfortran -Kieee test_nan.f90 
hkershaw@cheyenne5:$ ./a.out 
             NaN 
  T
  F
 you have a NaN 
```

[...]

PR #341, opened 2022-04-01
Bug fix - Fixes issue if FillValue is NaN

### LINKED ISSUE #323: bug: Checking equality of _FillValues that are NaNs in io_filenames_mod causes DART to exit.
:bug:  :bug:  :bug: 

### Describe the bug

Checking equality of _FillValues that are NaNs in io_filenames_mod causes DART to exit.

`check_attribute_value` is called to check `_FillValue` and `missing_value` with the following code (r4 and r8) versions:

```
if ( nf90_get_att(ncFile, ncVarID, att_string, ret_spvalR4) == NF90_NOERR ) then
   if (spvalR4 /= ret_spvalR4) then
      write(msgstring,*) ' variable attribute, ', trim(att_string), ' in state', spvalR4, &
                         ' does not match ', trim(att_string), ' ', ret_spvalR4, ' in ', trim(filename)
      call error_handler(E_ERR, 'check_attribute_value_r4', msgstring, source)
   endif
endif
```

If _FillValue is NaN then `(spvalR4 /= ret_spvalR4)` is always true. 

### Error Message  

```
 ERROR FROM:

  source : io_filenames_mod.f90
  routine: check_attribute_value_r4
  message:   variable attribute, _FillValue in state            NaN  does not match _FillValue             NaN  in /glade/work/gharamti/wrfhydro_dart/florence/initial_ens/hybrid/init_ens/member_000/HYDRO_RST.2018-08-01_00:00_DOMAIN1
```

Found with wrf-hydro, will effect any netcdf variable that has a missing or fill value attribute that is NaN.

### Version of DART      
9.12 (been there since at least 2016 though)

No

Note to myself:  You can test for NaNs before testing /=, but why do we even have to check the missing/fill value attribute is the same? State vs. netcdf file. Input vs. Output.   Query the missing/fill value before you write?

Benjamin-Gunn:
Fixes #323

## Description:

Bug fix for the Nan FillValue.

### Fixes issue
Fixes #323 

### Types of changes

- [ ] Documentation update

### Tests
Tested with wrf NaN case.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [ ] No dataset needed


assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -717,6 +717,9 @@ subroutine check_attribute_value_r4(ncFile, filename, ncVarID, att_string, spval
 real(r4) :: ret_spvalR4
 
 if ( nf90_get_att(ncFile, ncVarID, att_string, ret_spvalR4) == NF90_NOERR ) then
+   if (isnan(ret_spvalR4)) then
```
hkershaw-brown, line 720:
```suggestion
   if (ret_spvalR4 /= ret_spvalR4) then 
```

assimilation_code/modules/io/io_filenames_mod.f90
```diff
@@ -742,6 +745,9 @@ subroutine check_attribute_value_r8(ncFile, filename, ncVarID, att_string, spval
 real(r8) :: ret_spvalR8
 
 if ( nf90_get_att(ncFile, ncVarID, att_string, ret_spvalR8) == NF90_NOERR ) then
+   if (isnan(ret_spvalR8)) then
```
hkershaw-brown, line 748:
```suggestion
   if (ret_spvalR8 /= ret_spvalR8) then
```

hkershaw-brown (changes requested):
Great fix, thanks Benjamin.

I've been testing the fix on Cheyenne with several different compilers.  (ifort, gfortran, nvfortran). One of the compilers (nvfortran) that people use on Cheyenne does _not_ have an `isnan` function. 

So let's replace `isnan` with `ret_spvalR8 /= ret_spvalR8` to make the code work with all the Cheyenne compilers. 

If you are interested, we have has some discussion of NaNs here #176 
Once you have access to Cheyenne you will be able to try out the various compilers. 

Cheers,
Helen

hkershaw-brown (approved):

PR #355, opened 2022-05-23
bug-fix: rtps inflation 

### LINKED ISSUE #353: RTPS inflation flavor not behaving as it supposed to
## Describe the bug
Relaxation to Prior Spread (RTPS) is inflation flavor 4 in DART. This is only posterior inflation and so the user cannot do `inf_flavor` 4 in the first namelist column. For more information about RTPS, you can refer to the inflation section in the documentation [here](https://docs.dart.ucar.edu/en/latest/guide/inflation.html?highlight=RTPS#inflation). 

To activate RTPS, one can set the second column of `inf_flavor` to 4 and enter a value between 0 and 1 for `inf_initial`. It's important to mention that this `inf_initial` parameter is not the inflation value but rather a weighting factor the algorithm uses to find the effective posterior inflation. When the experiment is done, all posterior inflation values are stored in the netcdf variable `state_postinf_mean`. Setting `inf_inital` to 0 technically means that RTPS in inactive so filter should behave as if `inf_flavor` was set 0 rather than 4. However, this is not the case as I show in the L96 example below.

### Case 1: `inf_flavor = 0, 0` 
This is a scenario where no inflation is used. The ensemble size is set to 10 and I assimilate every other observation (a total of 20 out 40). I plot the prior diagnostics using MATLAB's `plot_total_error` below:
<p align="center">
  
</p>

### Case 2: `inf_flavor = 0, 4` and `inf_initial = 0, 0` 
Here, I am selecting RTPS however I'm also choosing 0 for `inf_inital` so no inflation should be applied. The resulting diagnostics for this case are shown in the figure below: 
<p align="center">
  
</p>
As can be seen these two cases do not match!

## Error Message  
No error messages. 

## Which model(s) are you working with?
Lorenz 96

## Version of DART      
v9.11.12-256-g2b92b63da

## Have you modified the DART code?   
Yes. I believe the issue lies in the `inflate_ens` routine within `adaptive_inflate_mod.f90` (lines 551-553). Here is the original section that does RTPS inflation: 
```
      ! only inflate if spreads are > 0
      if ( asprd .gt. 0.0_r8 .and. fsprd .gt. 0.0_r8) &
          inflate = 1.0_r8 + inflate * ((fsprd-asprd) / asprd) 
          ens     = mean + (ens-mean) * inflate 
```
My changes that fixe the problem are below: 

```
      ! only inflate if spreads are > 0
      if (asprd .gt. 0.0_r8 .and. fsprd .gt. 0.0_r8) then 
         inflate = 1.0_r8 + inflate * ((fsprd-asprd) / asprd)
      else 
         inflate = 1.0_r8
      endif 

      if (inflate .ne. 1.0_r8) ens = mean + (ens-mean) * inflate
```
This should be looked at in more detail. In the old code, it looks like even when `inflate` is 1, the ensemble is changing after this routine. There is some funny numerics action happening here! To fix this, I enforced inflation only when `inflate` is different than 1. Now, when I run the two cases again I get the same results. 

One last observation: It looks like this behavior starts fading away as I increase the ensemble size. It's definitely more pronounced when the ensemble size is tiny. 

## Build information
- Machine: Laptop
- Compiler: gfortran

hkershaw-brown:
Bug fix for rpts with inflate=1
Follows the non-rtps order of operations to avoid inflation of 1.0 changing ens.

## Description:

Please provide a summary of the change and include any relevant motivation and context. 

### Fixes issue
Fixes #353 

### Types of changes

- [ ] Documentation update

### Tests

```
$cat test_order.f90 
program order

implicit none

integer,parameter :: n = 3
real*8 :: ens(n), mean, inflate
real*8 :: ens2(n), ens3(n), ens4(n)
integer :: i

inflate = 1.d0

do i = 1, n
   ens(i) = real(i,8)/10000
   mean = mean + ens(i)
enddo
!mean = mean/real(n,8)
! This is not the mean, just a number to trigger differences in the 
! calculation for this test code
mean = 4.0d8


ens2 = mean + (ens-mean) * inflate 
ens3 = ens * inflate + mean * (1.d0 - inflate)
ens4 = ens*inflate + mean - mean*inflate

print*, 'mean', mean
print*, 'ens ', ens
print*, 'ens2 ', ens2
print*, 'ens3 ', ens3
print*, 'ens4 ', ens4

end program order
```

```
gfortran test_order.f90 
./a.out

mean   400000000.00000000     
 ens    1.0000000000000000E-004   2.0000000000000001E-004   2.9999999999999997E-004
 ens2    1.0001659393310547E-004   1.9997358322143555E-004   2.9999017715454102E-004
 ens3    1.0000000000000000E-004   2.0000000000000001E-004   2.9999999999999997E-004
 ens4    1.0001659393310547E-004   1.9997358322143555E-004   2.9999017715454102E-004
```

```
ifort test_order.f90 
./a.out 
mean   400000000.000000     
 ens   1.000000000000000E-004  2.000000000000000E-004  3.000000000000000E-004
 ens2   1.000165939331055E-004  1.999735832214355E-004  3.000000000000000E-004
 ens3   1.000000000000000E-004  2.000000000000000E-004  3.000000000000000E-004
 ens4   1.000165939331055E-004  1.999735832214355E-004  3.000000000000000E-004
```

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
+++ b/assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
@@ -550,7 +550,7 @@ if(inflate_handle%deterministic) then
       ! only inflate if spreads are > 0
       if ( asprd .gt. 0.0_r8 .and. fsprd .gt. 0.0_r8) &
           inflate = 1.0_r8 + inflate * ((fsprd-asprd) / asprd) 
-          ens     = mean + (ens-mean) * inflate 
+          ens = ens * inflate + mean * (1.d0 - inflate)
    else 
 
       ! Spread the ensemble out linearly for deterministic
--- a/assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
+++ b/assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
@@ -550,7 +550,7 @@ if(inflate_handle%deterministic) then
       ! only inflate if spreads are > 0
       if ( asprd .gt. 0.0_r8 .and. fsprd .gt. 0.0_r8) &
           inflate = 1.0_r8 + inflate * ((fsprd-asprd) / asprd) 
-          ens = ens * inflate + mean * (1.d0 - inflate)
+          ens = ens * inflate + mean * (1.0_r8 - inflate)
    else 
 
       ! Spread the ensemble out linearly for deterministic
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,11 @@ individual files.
 
 The changes are now listed with the most recent at the top.
 
+**May 23 2022 :: Bug-fix for RTPS inflation flavor. Tag: v9.16.4**
+
+- Order of operations changed to avoid inadvertent changes to ens 
+  when using RTPS.
+
 **May 16 2022 :: Installation documentation update. Tag: v9.16.3**
 
 - Improved installation documentation.
--- a/conf.py
+++ b/conf.py
@@ -21,7 +21,7 @@ copyright = '2021, University Corporation for Atmospheric Research'
 author = 'Data Assimilation Research Section'
 
 # The full version, including alpha/beta/rc tags
-release = '9.16.3'
+release = '9.16.4'
 master_doc = 'README'
 
 # -- General configuration ---------------------------------------------------


hkershaw-brown:
@mgharamti when you get a chance, can you give this change a try with your case in https://github.com/NCAR/DART/issues/353

https://github.com/NCAR/DART/pull/355/files
```
ens = ens * inflate + mean * (1.d0 - inflate)
```

and see if it behaves as expected.

mgharamti:
> @mgharamti when you get a chance, can you give this change a try with your case in #353
> 
> https://github.com/NCAR/DART/pull/355/files
> 
> ```
> ens = ens * inflate + mean * (1.d0 - inflate)
> ```
> 
> and see if it behaves as expected.

@hkershaw-brown That indeed does fix the problem. I'm getting the right behavior; i.e., flavor 4 with `inf_initial=0` is equivalent to flavor just 0. Thanks for looking into this. We should push this bug fix as soon as we can.

mgharamti (approved):

assimilation_code/modules/assimilation/adaptive_inflate_mod.f90
```diff
@@ -550,7 +550,7 @@ subroutine inflate_ens(inflate_handle, ens, mean, inflate, var_in, fsprd, asprd)
       ! only inflate if spreads are > 0
       if ( asprd .gt. 0.0_r8 .and. fsprd .gt. 0.0_r8) &
           inflate = 1.0_r8 + inflate * ((fsprd-asprd) / asprd) 
-          ens     = mean + (ens-mean) * inflate 
+          ens = ens * inflate + mean * (1.d0 - inflate)
```
mjs2369, line 553:
```suggestion
          ens = ens * inflate + mean * (1.0_r8 - inflate)
```

mjs2369 (changes requested):

mjs2369 (approved):

PR #371, opened 2022-07-14
Bugfix for polar, periodic x in WRF model_mod

### LINKED ISSUE #369: bug: inconsistent output for variables polar and periodic_x
:bug: 

### Describe the bug

WRF ensemble runs that should be bitwise identical are not, as they display differences in their output NetCDF files for the variables polar and periodic_x. 

These variables are stored in the tmp array in the WRF model_mod, which is declared as size(num_domains), but this is due to an outdated version of DART in which a single NetCDF file could have multiple domains. Currently, DART only supports a single domain per NetCDF file. 

Screen Shot 2022-07-13 at 6 21 11 PM

When num_domains > 1 in &model_nml in input.nml, the array of values is written as output (see line 4300). Since there is only a single domain per NetCDF file and therefore there is only a single value for polar or periodic_x, the remaining indices are empty. This causes meaningless values to be inserted into the vacant spots in the array when they are written to the NetCDF file. This causes bitwise identical testing to fail by displaying differences in the output files for these variables.

Additionally, there are some incorrect comments in the model_mod. For example, the comment at line 304:
` ! Boundary conditions -- hopefully one day these will be in the global attributes of the
   !   input NetCDF file ("periodic_x" and "polar" are namelist items in the &bdy_control
   !   section of a standard WRF "namelist.input" file), but for now we have included them
   !   in the "model_nml" group of DART's own "input.nml".  Above, their default values are
   !   both set to .true. (indicating a global domain). `

does not reflect the change made to the default setting for these variables, which is now .false.

- Set the variable `num_domains = 2` in &model_nml in input.nml
- Perform bitwise identical testing on two runs of WRF by comparing the data written to the output NetCDF files.

The output files are identical across runs that are otherwise bitwise identical (ex: two WRF runs in which no changes were made to the code or input values).
 
3. What actually happened?  

Values in indices greater that 1 for the variables periodic_x and polar written into output NetCDF files will vary for almost every execution of WRF.

The following image shows the results of a diff command on the output file output_mean_d01.nc for two WRF runs on the main branch, having made no changes in between. 

Screen Shot 2022-07-13 at 5 21 15 PM

### Solution
This can be a simple fix by declaring the tmp variable as an integer instead of an array, and then reflecting this change in the calls to nf90_put_var().

However, there is more work that can be done to clean up the WRF model_mod by removing the variable num_domains entirely, as there are other variables in the subroutine nc_write_model_atts() that are declared as arrays of the size(num_domains) despite only ever holding a single value. 

WRF

### Version of DART     
v10.0.2

No  

### Build information
Please describe:  
 1. The machine you are running on (e.g. windows laptop, NCAR supercomputer Cheyenne).   
Any machine.
 2. The compiler you are using (e.g. gnu, intel).  
Intel.

mjs2369:
## Description:
The two variables polar and periodic_x in the WRF model_mod are logical variables, which are converted to ints and stored in arrays of size(num_domains).

However, the variable num _domains is no longer used as an attribute for output NetCDF files, as it is leftover from outdated code in which multiple domains could belong to a single NetCDF file. Updates to DART make it so that there can only a single domain per NetCDF file. Therefore, there is only a single value for these variables, and the tmp array is only filled in the first index.  

When num_domains > 1 in &model_nml in input.nml (line 216), meaningless values are written to the NetCDF file for the vacant indices in the tmp array. This will cause bitwise identical testing in WRF to fail due to the inconsistent values of these variables being put in the output files. 

Changes made to fix this:
- Change the definition of the tmp variable from an array to a integer
- Adjust the calls to nf90_put_var() to reflect this change to the tmp variable
- Fix the incorrect comment that states the default setting for these variables is .true.

### Fixes issue
Fixes #369 

### Types of changes

- [ ] Documentation update

### Tests
Tested using WRF test case listed below as well as ensemble runs in the wrf/work dir. 

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [x] No dataset needed

Tested with the test case found in /glade/scratch/hkershaw/DART/siparcs2022/test_cases/wrf 
However, the changes can also be tested with any wrf case, as long as the variable num_domains is set to `num_domains = 2` in &model_nml in input.nml


Benjamin-Gunn (approved):
Seems like a logical fix. Looks good!

hkershaw-brown (changes requested):
Hi Marlee,

The fix looks good, however there are other variables with dimensions(num_domains) in this routine. This seems like the perfect time to pull out the num_domains from nc_model_atts.  That way, no one is tempted to add 1:num_domains back in.

Cheers,
Helen

mjs2369:
Hi @hkershaw-brown ,

I agree. Would it be okay for me to go ahead and make those commits to this PR, or should I make a separate issue for it?

hkershaw-brown:
on this PR is perfect.

hkershaw-brown (approved):

hkershaw-brown:
@hkershaw-brown 

bundle with https://github.com/NCAR/DART/pull/380 for release

Issue #359, opened 2022-06-02
State IO


hkershaw-brown:
Notes for state IO

* read_variables is public and is used by mpas_atm/update_bc.f90
* out of date comments in state_vector_iod_mod.f90 (e.g reference to construct_file_name_in() which no longer exists)
    * read_model_time directly from model_mod rather than assim_model_mod
    * out of date comments on "dart format read"
* io_filenames_mod has a bunch of references to diagnostic files which no longer exist
* io_filanmes_mod is doing a bunch of netcdf variable checking 
   * state_structure_mod also has a mix of netcdf file info vs. state info
* stages_to_write is in state_vector_io_mod - more sense to have this info in io_fileaames_mod?
*  see #283 for round robin assumption

* write_augmented_state - why the restriction on  `! if there is only one timestep in your input file insert the mean and sd if requested`

hkershaw-brown:
In `create_and_open_state_output`, only 'time'  (lower case) can be the unlimited dimension.

https://github.com/NCAR/DART/blob/cbe0d41e9ce190366b3cc5cc1433becf2ebc421c/assimilation_code/modules/io/direct_netcdf_mod.f90#L1643-L1656

If you have an unlimited dimension in the state_structure_mod, e.g. WRF has 'Time' this gets created as a limited dimension.

This means that your created netcdf files have a different netcdf dimension structure than your model files. 

`has_unlimited` is a property of the state_stucture%domain, but it is (can be) different between netcdf files because of create_and_open_state_output.  I think this is why dart_time_io_mod is querying the netcdf file rather than the state_structure (see comment above). 

There is also `domain%variable(ivar)%var_has_unlim` which is per variable, set but never used.

hkershaw-brown:
This is a hard coded 0 however, there is a `integer, parameter :: SINGLE_IO_TASK_ID = 0` in the module.

https://github.com/NCAR/DART/blob/cc1da67976feb3b092ac9eae7388f6181ef33154/assimilation_code/modules/io/direct_netcdf_mod.f90#L431-L434

* I don't think this needs a broadcast, since if all tasks are in the initialize_single_file_io routine, all tasks have access to the local variable `local_model_mod_will_write_state_variables`
* double check pe vs task: since `initialize_single_file_io`(**ens_handle**, file_handle) should you be using **ens_handle**%my_pe rather than my_task_id for this routine? Maybe it does not matter. Similarly for `read_single_file`:
[...]

PR #374, opened 2022-07-19
TIEGCM 2.0 Manhattan

### LINKED ISSUE #309: TIEGCM
Notes for TIEGCM model_mod

X vs X_MN variables - Forcing X at the moment 25de28e77a731a5016ff57427cb53cdde841b0c8

Original note in Lanai model_mod.f90
```
! Note from FindVar_by_kind:
! FIXME There is some confusion about using the T-minus-1 variables
! in this construct. Both TN and TN_NM have the same dart_kind,
! so we use the first one ... but it is not guaranteed that TN
````

### PR DESCRIPTION
**Pull request by hkershaw-brown**
## Description:

TIEGCM model_mod updated to Manhattan
Tested with TIEGCM 2.0

* Supports 2.5-degree and 5-degree resolution grids.
* vtec extrapolation is done in model_mod::model_interpolate with 10 levels for 5deg, 20 levels for 2.5deg.
* Model_interpolate uses X rather than X_MN variable no matter the order of state
  variables in model_nml. Previously in Lanai the variable given first in the state was used.
  _MN variables are from the previous model time step.
* f10.7 can be added to the DART state. An example f10_7.cdl file is provided in the work directory. At the moment f10.7 is 
  in the CALCULATE/CONSTRUCTED domain. This could do with a better name. 
* Added hyperslice subroutine to state_structure_mod to read only part of a netcdf variable.
* Separated functions for io_dim_length and dim_length in state_structure_mod to
  allow for hyperslice.

Tiegcm has:
  - the lower boundary condition at the top level. This should not be altered by filter.
  - multiple timesteps per restart file.

**obs_def_upper_atm_mod.f90 changes**
  `get_expected_vtec` 'routine needs to be written' removed and replaced with common code.  

**state_structure_mod.f90 changes:**
new `hyperslice_domain` - cuts off the top level
new `has_unlimited_dim` query if the domain has an unlimited dimension.

`get_io_dim_length` -   returns io_info%dimlens so you can query the dimensions of the netcdf file vs. the state. 
`state_structure_info` print statement changed to not write out get_original_dim_ID - this segfaults for add_domain_from_spec.

**direct_netcdf_mod.f90 changes**
Read the latest slice of the unlimited dimension in the restart file.
Write the latest slice of the unlimited dimension in the restart file. 

**scripts**
There is are scripts in the directory in tiegcm/shell_scripts for running on Cheyenne
I am not sure if these are general enough to include in the release. I can remove them from this pull request. 

### Fixes issue

Fixes #309, #269

### Types of changes

- [x] Documentation update

### Tests

tiegcm model_mod_check
Cycled tiegcm on Cheyenne.  The shell scripts are included in this pull request,  I am not sure if these are general enough.  I can remove them from this pull request.

To build TIEGCM2.0 on Cheyenne I had to use esmf-7.1.0r. Cheyenne build script here: https://github.com/hkershaw-brown/tiegcm-cheyenne/blob/main/tiegcm-ch.job 

Tests for Netcdf changes:
lorenz_96 bitwise
wrf bitwise

## Checklist for merging

- [ ] Updated changelog entry
- [x] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [ ] No dataset needed

### CODE REVIEW

File: assimilation_code/modules/io/direct_netcdf_mod.f90 [hunk 1]
```diff
@@ -852,22 +851,41 @@ subroutine read_variables(ncfile_in, var_block, start_var, end_var, domain)
...
+   num_dims = get_io_num_dims(domain, i)
+   allocate(counts(num_dims))
+   allocate(slice_start(num_dims))
 
-   dims = get_io_dim_lengths(domain, i)
+   if (has_unlimited_dim(domain)) then
+
+      slice_start(:) = 1 ! default to read all dimensions start at 1
+      counts(num_dims) = 1 ! one slice of unlimited dimesion
+      counts(1:num_dims-1) = get_dim_lengths(domain, i) ! the state
```
nancycollins, line 863: a couple of comments for the future here.  newer versions of netcdf allow for multiple unlimited dimensions so this approach will have to be altered if we run into a model that uses this in the input/output state to dart.  we can always require a model_to_dart and dart_to_model as a temporary measure once we do see how people might be using multiple unlimited dims.

also, it's possible that netcdf variables in files have additional (not-unlimited) dimensions that shouldn't be part of the state.  i can't remember which example or test had this, so for now it's not an issue.   but eventually some namelist items may be needed to address i/o issues - for example, the name(s) of the unlimited dimension(s) (e.g. Time), other name(s) of dimensions to ignore when reading into the state (e.g. leapfrog_step1 or something), and then for the ignored dimensions, what index should be used if the length is more than 1 (e.g. last, 1, etc).

assimilation_code/modules/io/state_structure_mod.f90
```diff
@@ -936,7 +947,76 @@ subroutine load_common_cf_conventions(domain)
...
+
+domain_size = 0
+! variable size
+do ivar = 1, get_num_variables(dom_id)
+
+   variable_size = 1
+
+   do jdim = 1, get_num_dims(dom_id, ivar)
+      !HK there is a todo about this, what is going on with time and member?
+      if ((state%domain(dom_id)%variable(ivar)%dimname(jdim) == 'time') .or. &
```
nancycollins, line 997: i remember this being added to the existing code but i don't remember why, sorry.  "member" is the dimension created by dart when making a unified (single) file that holds all the ensemble member info.  "time" is most often an unlimited dim but nothing in netcdf forces it to be.

models/tiegcm/model_mod.f90
```diff
@@ -160,82 +161,52 @@ module model_mod
...
+real(r8)  :: bot_lat        = MISSING_R8
+real(r8)  :: top_lat        = MISSING_R8
+real(r8)  :: delta_lat      = MISSING_R8
+integer   :: zero_lon_index = MISSING_I
+
+
+! Obs locations are expected to be given in height [m],
+! and so vertical localization coordinate is *always* height.
+! Note that gravity adjusted geopotential height (ZG) is read in
+! "tiegcm_s.nc". ZG is 'cm', dart is mks(HK mks?))
```
nancycollins, line 189: "mks" = meter, kilogram, second as the standard units.

models/tiegcm/model_mod.f90
```diff
@@ -251,143 +222,36 @@ subroutine static_init_model()
...
+! error-check, convert namelist input to variable_table, and build the
+! state structure
+call verify_variables()
 
-! Might as well use the Gregorian Calendar
 call set_calendar_type('Gregorian')
 
-! Convert the last year/doy/hour/minute to a dart time.
-state_time = get_state_time(tiegcm_restart_file_name)
+! Convert the last year/day/hour/minute to a dart time.
```
nancycollins, line 238: does tiegcm time use 'doy' == day of year, 1-366?  if so, this isn't really a typo.  but i suppose day is just as clear, to be honest.

models/tiegcm/model_mod.f90
```diff
@@ -160,82 +161,52 @@ module model_mod
...
+real(r8)  :: bot_lon        = MISSING_R8
+real(r8)  :: top_lon        = MISSING_R8
+real(r8)  :: delta_lon      = MISSING_R8
+real(r8)  :: bot_lat        = MISSING_R8
+real(r8)  :: top_lat        = MISSING_R8
+real(r8)  :: delta_lat      = MISSING_R8
+integer   :: zero_lon_index = MISSING_I
+
+
+! Obs locations are expected to be given in height [m],
```
nancycollins, line 186: maybe adjust this comment?  it looks like model_interpolate also supports obs on model levels?  it's probably true that only height localization is supported, but asking for obs values on levels seems like it works?

models/tiegcm/model_mod.f90
```diff
@@ -396,371 +260,247 @@ function get_model_size()
...
-! Get lon and lat grid specs
-bot_lon        = lons(1)                         ! 180.
-delta_lon      = abs((lons(1)-lons(2)))          ! 5. or 2.5
-zero_lon_index = int(bot_lon/delta_lon) + 1      ! 37 or 73
-top_lon        = lons(nlon)                      ! 175. or 177.5
-bot_lat        = lats(1)                         !
-top_lat        = lats(nlat)                      !
-delta_lat      = abs((lats(1)-lats(2)))          !
+      ! HK pressure is not in the state so do not call get_state.
+      call error_handler(E_ERR, 'model_interpolate', 'vert is height for pressure interpolate')
```
nancycollins, line 360: if this is an error message, it needs to say why it's failing.  "pressure obs with vert in height not supported"  or something?

models/tiegcm/model_mod.f90
```diff
@@ -769,1225 +509,345 @@ end subroutine get_state_meta_data
...
+integer,  dimension(nmtime) :: mtime  ! day, hour, minute
+integer                     :: year, doy, utsec
+integer, allocatable, dimension(:,:) :: mtimetmp
+integer, allocatable, dimension(:)   :: yeartmp
 
-if ( .not. module_initialized ) call static_init_model
+if ( present(ncfileid) ) then
+   ncid = ncfileid
+else
+   call nc_check(nf90_open(filename, NF90_NOWRITE, ncid), &
```
nancycollins, line 785: eventually these calls could use the dart netcdf utility routines.

models/tiegcm/model_mod.f90
```diff
@@ -769,1225 +509,345 @@ end subroutine get_state_meta_data
...
-   call error_handler(E_MSG,'read_TIEGCM_namelist:',string1)
-endif
-
-end subroutine read_TIEGCM_namelist
-
-
-!-------------------------------------------------------------------------------
-
-
 subroutine read_TIEGCM_definition(file_name)
```
nancycollins, line 851: also eventually could be converted to using the dart netcdf util routines.

nancycollins (approved):
i left a few comments but nothing that prevents this from being merged with main.  i will suggest that the tiegcm test cases include obs locations in height that are exactly on model levels, both at the bottom, top, and in the middle to test when the fractions across the levels are exactly 0.0, 1.0, and inbetween.  but otherwise, getting rid of vector_to_model, model_to_vector, progvar, and moving up to the manhattan interfaces is a huge step forward.  

the release notes for this might ask users to report any changes in how dart reads their netcdf files.  i don't see how the state struct and i/o changes here would cause problems, but netcdf is so flexible and users are so creative in how they structure their data that it's hard to predict what changes might cause problems for someone.

assimilation_code/modules/io/direct_netcdf_mod.f90
hkershaw-brown: Totally agree with this, I'm keeping track State IO issues  here: https://github.com/NCAR/DART/issues/359 for a future refactor.

models/tiegcm/model_mod.f90
hkershaw-brown: oh, I read doy and day as the same word. 

tiegcm uses day,hour,minute. The tiegcm user guide uses 'day' for day-of-year, e.g.

```
STOP integer(3) Model stop time (day,hour,minute)
````

so it makes sense for DART to follow that.

models/tiegcm/model_mod.f90
hkershaw-brown: I'll strip this out, this is me catching pressure when testing.  Sorry.

models/tiegcm/model_mod.f90
hkershaw-brown: I'll do this in this pull request - good time to do it.

models/tiegcm/model_mod.f90
hkershaw-brown: ditto

models/tiegcm/model_mod.f90
hkershaw-brown: removed.

models/tiegcm/model_mod.f90
hkershaw-brown: done.

models/tiegcm/model_mod.f90
hkershaw-brown: done [cb492ec](https://github.com/NCAR/DART/commit/cb492ecf7a6469105de176a65a27660589180cde)

models/tiegcm/model_mod.f90
hkershaw-brown: done 8445726fee511fa19ecd2630a559a2dc2bdbd691

Issue #349, opened 2022-05-11
Feature request: Automated creation of new model interfaces


hkershaw-brown:
### Use case  

A user wants to add a new model, where do they start from?

### Is your feature request related to a problem?  

When a user wants to add a new model interface (model_mod) to DART they typically copy an existing model_mod. This has a few unwanted side effects:

* Copied code from an existing model is not necessary relevant to the new model (code is messier than it needs to be)
* Users are not necessarily getting the latest code/best practices (creates more work for DART developers during pull requests)
* Copied documentation can be misleading (it looks up to date, but it is for another model)

### Describe your preferred solution  

Create a script that automatically sets up a clean model interface for users to start developing and writing their documentation.  I believe with good template files for model_mod.f90 we can have a model interface that compiles from the get-go ( note: this may not be true).

Here is a spec:
https://docs.google.com/document/d/1u1v6MoGIQTKSRIxzeCESJMVOu-hU7DFpCU-eTNDbQt4/edit?usp=sharing

### Describe any alternatives you have considered  
I think we just need code that writes code.

hkershaw-brown:
Next steps:
* Use shell check to check the bash scripts https://www.shellcheck.net/  This may suggest some changes to the scripting.
* For the readme.rst created in work - something sensible here for a skeleton documentation.  This could be:

   ```
  ==============
   Name of the model
  ===============

   .. attention::
       Add your model documentation here.

   ```
   Note we do want to keep the template/readme.rst because this is the documentation: https://docs.dart.ucar.edu/en/latest/models/template/readme.html
 so you'll need to create template/new_model.rst -> work/readme.rst
* create a oned model_mod.f90 and input.nml template.  You can take a look at lorenz_96 for a working 1d model mod.  The goal is to have a template version similar to the template/model_mod.f90( works for threed_sphere).

PR #376, opened 2022-07-21
New model automation

Benjamin-Gunn:
## Description:

Previously, the creation of a new model would involve copying the files an already existing model. However, this process is not the most efficient and can lead to some unwanted side effects, like out of date standards. 

To make a new model now, users can call the script `new_model` alongside the desired model name and location module. If the model name is available and the location module is correct, the script will automatically create the necessary directories and files for a functional model. This creates a steady foundation for the user to commence further model development.

The new model copies from `DART/models/template` and begins by adjusting `quickbuild.sh` and `readme.rst` (renamed from `template/new_model.rst` by the script) to have the inputted name and location. The script automatically chooses from the currently two new templates for `model_mod.f90` and `input.nml` based on location given, then renames them accordingly. The currently supported location templates are for 3D and 1D modules, with the possibility for more in the future. 

At the moment, `threed_sphere`, `threed_cartesian`, and `oned` locations compiled with the scripts selections. If the location is not listed here, the script will default to 3D as the default as it provides a good example on the typical files structure.

### Fixes issue

fixes https://github.com/NCAR/DART/issues/349, #382

### Types of changes

- [ ] Documentation update

### Tests
Tested model creation and compilation on personal computer and Cheyenne.

## Checklist for merging

- [x] Updated changelog entry
- [x] Documentation updated
- [x] Version tag 

## Testing Datasets

- [ ] No dataset needed


hkershaw-brown:
See #379 for the note on documentation on the required routines.

models/new_model
```diff
@@ -0,0 +1,48 @@
...
+elif [[ -d "$model_name" ]]; then
+  echo "$model_name already exists! Please try a different model name."
+  exit 1
+elif [[ $# -eq 1 ]]; then
+  echo "Please enter a proper location! (Hint: check DART/assimilation_code/location for location modules avaliable.)"
+  exit 1
+elif [[ ! -d "../assimilation_code/location/$location_mod" ]]; then
+  echo "$location_mod is not a proper location module! Please try a different location name."
+  exit 1
+else
```
hkershaw-brown, line 22: Functionally this is good, but I think it is worth refactoring to make the script easier to add to in the future.

Here you have a giant if statement that runs from line 10 to line 48, but you do not need to:

pseudo code:
```
check the arguments
   if arguments are wrong, exit

create the template
```

The argument checking is good, I would create a function "usage" which prints out the usage.

models/new_model
```diff
@@ -0,0 +1,48 @@
...
+location_mod=$2
+
+if [[ ("$1" = "help") || ("$1" = "--help") || ($# -eq 0) ]]; then
+  echo "Usage: $0 [model_name] [location_mod]"
+  exit 0
+elif [[ -d "$model_name" ]]; then
+  echo "$model_name already exists! Please try a different model name."
+  exit 1
+elif [[ $# -eq 1 ]]; then
+  echo "Please enter a proper location! (Hint: check DART/assimilation_code/location for location modules avaliable.)"
```
hkershaw-brown, line 17: If `$# -eq 1` this could be the user has given a location, but not a model_name.
If the arguments are not what is expected, best to print the usage

models/new_model
```diff
@@ -0,0 +1,48 @@
...
+  exit 1
+elif [[ ! -d "../assimilation_code/location/$location_mod" ]]; then
+  echo "$location_mod is not a proper location module! Please try a different location name."
+  exit 1
+else
+  cp -R template "$model_name"
+  rm -r "$model_name"/shell_scripts "$model_name"/model_mod.nml "$model_name/readme.rst"
+  mv "$model_name/new_model.rst" "$model_name/readme.rst"
+  sed -i '' "s/template_model/$model_name/" "$model_name/readme.rst"
+  sed -i '' -e  "s/template_model/$model_name/" -e "s/template_location/$location_mod/" "$model_name/work/quickbuild.sh"
```
hkershaw-brown, line 27: I can't remember the ins and outs of sed. Does sed -i ' ' work on mac and linux?

models/new_model
```diff
@@ -0,0 +1,48 @@
...
+  case $location_mod in
+    threed | threed_cartesian | threed_sphere) # threed does not compile, but is a good framework
+      rm -r "$model_name"/oned_model_mod.f90 "$model_name"/work/oned_input.nml
```
hkershaw-brown, line 30: -r (recursive) not needed here, since you are not removing directories. 

Some systems will have rm aliased to rm -i. If you want to a command, e.g.  rm and not any aliases, add \ to the start of the command:

```
\rm
```    

Also, since users often do things that you are not expecting you can use -- 

`\rm --  "$model_name"/oned_model_mod.f90`

`--` means "end of the options". Otherwise if a user decides to give their model a name starting with -, e.g.

`model_name = -fish_pie`

rm will think you are trying to give -fish_pie as an option to rm.

hkershaw-brown: Can you only copy the files you need, rather than copying everything and then removing what is not needed? Two reasons to do this:

1. I'd like to have as few `rm`s as possible when 
2. It is possible that more files get added to the template directories, and if so, they will get copied, but not removed by this script.

hkershaw-brown (changes requested):

models/new_model
mgharamti: As a frequent sed user, I can confirm that sed -i '' does work on mac. I don't know about linux though

mgharamti: Because you're working with directories, I'd replace the slash with some other character to avoid any confusion. For instance: `sed -i '' "s*template_model*$model_name*"` instead of `sed -i '' "s/template_model/$model_name/"`

guide/advice-for-new-collaborators.rst
```diff
@@ -312,10 +312,26 @@ of the state variables then doing it on demand is more efficient.
...
-------------------------
-
-The ``models/template`` directory has files that can be used to start
-porting code to support a new model, but we also recommend looking at
-the existing supported models and reusing code from them if possible.
-Models with similar grid types or vertical coordiates are good candidates.
+New model automation and code reuse
+-----------------------------------
+
+Adding a new model can be done with the script ``models/new_model.sh``.
```
mgharamti, line 318: As it stands the script `new_model` does not have a `.sh` extension

models/new_model
mgharamti: Agree, or they might have entered something wrong e.g., `./new_model -help`

models/new_model
Benjamin-Gunn: Hello! I have done some testing and I have found when I substitute the slash in the commands, they usually can't find the file and I get ``sed: test*work*quickbuild.sh: No such file or directory``.  I have tried some other characters too, but no luck so far. So, I was wondering if there is some way I am meant to allow the code to read these characters as file paths? Also, if I do get another character to work, would you like me to replace all the slashes in the code to use this characters for the rest of the script? Finally, I was able to get the ``sed -i '' ...`` to work on my computer and Cheyenne. Thank you!

nancycollins: try semicolons as your separators. those often work without causing new problems.

Benjamin-Gunn: Ah, cool! I actually found with a bit more testing that I made an oversight by replacing the file path's slashes too. This just so happens to be the confusing I am trying to prevent, which is quite something! I guess this really shows how important this change is. I might just go with the `*`s for now too, because I feel like they look the cleanest. Thanks for your help!

nancycollins: i'm glad they work, but just an fyi: i think people may avoid using asterisks as separators because they are the "regular expression" wildcard character. there are times you want to use asterisks in the patterns you are matching so then they can't also be the separators.

Benjamin-Gunn: Oh, ok. Then would it be best for me to change them to your suggestion of the semicolons?

guide/advice-for-new-collaborators.rst
```diff
@@ -312,10 +312,26 @@ of the state variables then doing it on demand is more efficient.
...
-------------------------
-
-The ``models/template`` directory has files that can be used to start
-porting code to support a new model, but we also recommend looking at
-the existing supported models and reusing code from them if possible.
-Models with similar grid types or vertical coordiates are good candidates.
+New model automation and code reuse
+-----------------------------------
+
+Adding a new model can be done with the script ``models/new_model``.
```
hkershaw-brown, line 318: The new_model script needs to be run in the models directory, if you run models/new_model

```
$ models/new_model hello oned
oned is not a proper location module! Please try a different location name.
```

guide/advice-for-new-collaborators.rst
hkershaw-brown: renamed to new_model.sh

models/new_model.sh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+mkdir -p "$model_name/work"
+cp "template/new_model.rst" "$model_name/readme.rst"
+sed -i '' "s;template_model;$model_name;" "$model_name/readme.rst"
+cp "template/work/quickbuild.sh" "$model_name/work/quickbuild.sh"
+sed -i '' -e  "s;template_model;$model_name;" -e "s;template_location;$location_mod;" "$model_name/work/quickbuild.sh"
+
+case $location_mod in
+  threed | threed_cartesian | threed_sphere) # threed does not compile, but is a good framework
```
hkershaw-brown, line 30: The threed version does not compile because the threed location_mod.f90 is out-of-date - the location_mod.f90 should have a get_close_obs and a get_close_state.  
I'll fix that and then I think your script will produce code that compiles for threed.

guide/advice-for-new-collaborators.rst
Benjamin-Gunn: Ooh, that would be great. Then it would also get rid of the confusion that the threed template works for every 3D component but threed!

hkershaw-brown (approved):
This is great Benjamin, it has already proved useful for testing various location mods.

Approved.

PR #387, opened 2022-08-11
bug-fix: mpas_atm call get_num_varids_from_kind to check if a qty is in the state

### LINKED ISSUE #386: bug: mpas_atm check for fields in the state fails if there are multiple fields of the same quantity
🦔  :bug: Your bug may already be reported!

### Describe the bug

    Run MPAS with 2 (or more) variables of the same quantity in the state
    The check for the required state quantities should be ok with multiple variables of the required quantities.
3. What actually happened?  
    it is not.

### Error Message  
```
 ERROR FROM:
  source : state_structure_mod.f90
  routine: get_varid_from_kind
  message:  Found            2  > 1
  message: ...  for dart kind : QTY_POTENTIAL_TEMPERATURE
  message: ...  Please use get_varids_from_kind to get a list of indices
```

MPAS_ATM

### Version of DART      
v10.2.0

No  

### Build information
Please describe:  
 1. Cheyenne
 2. intel

The fix is to use get_num_varids_from_kind
on the mega-mpas branch at the moment: 45d6f018d9b0192f0438fb2b34a9d0dfbc87032c
Note This also has some obs_def_mods that Chris R. was using.

hkershaw-brown:
## Description:

There may be multiple variables with the same qty in the state.
`get_varid_from_kind` errors out if there is more than one variable
of that quantity.
This fix uses `get_num_varids_from_kind` instead

### Fixes issue

fixes #386

### Types of changes

- [ ] Documentation update

### Tests
run mpas with these two QTY_POTENTIAL_TEMPERATURE variables in the state.

```
&mpas_vars_nml
  mpas_state_variables =  'theta',  'QTY_POTENTIAL_TEMPERATURE',
                          'th2m',   'QTY_POTENTIAL_TEMPERATURE'
                                         ...
```

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [ ] No dataset needed

/glade/scratch/hkershaw/DART/MPAS/test_run

### DIFF
--- a/models/mpas_atm/model_mod.f90
+++ b/models/mpas_atm/model_mod.f90
@@ -131,7 +131,7 @@ use state_structure_mod, only :  add_domain, get_model_variable_indices, &
                                  state_structure_info, get_index_start, get_index_end, &
                                  get_num_variables, get_domain_size, get_varid_from_varname, &
                                  get_variable_name, get_num_dims, get_dim_lengths, &
-                                 get_dart_vector_index, get_varid_from_kind
+                                 get_dart_vector_index, get_num_varids_from_kind
 
 
 implicit none
@@ -923,12 +923,12 @@ endif
 ! basically we cannot do much without having at least these
 ! three fields in the state vector.  refuse to go further
 ! if these are not present:
-!print *, get_varid_from_kind(anl_domid, QTY_POTENTIAL_TEMPERATURE)
-!print *, get_varid_from_kind(anl_domid, QTY_DENSITY) 
-!print *, get_varid_from_kind(anl_domid, QTY_VAPOR_MIXING_RATIO) 
-if ((get_varid_from_kind(anl_domid, QTY_POTENTIAL_TEMPERATURE) < 0) .or. &
-    (get_varid_from_kind(anl_domid, QTY_DENSITY) < 0) .or. &
-    (get_varid_from_kind(anl_domid, QTY_VAPOR_MIXING_RATIO) < 0)) then
+!print *, get_num_varids_from_kind(anl_domid, QTY_POTENTIAL_TEMPERATURE)
+!print *, get_num_varids_from_kind(anl_domid, QTY_DENSITY)
+!print *, get_num_varids_from_kind(anl_domid, QTY_VAPOR_MIXING_RATIO)
+if ((get_num_varids_from_kind(anl_domid, QTY_POTENTIAL_TEMPERATURE) < 0) .or. &
+    (get_num_varids_from_kind(anl_domid, QTY_DENSITY) < 0) .or. &
+    (get_num_varids_from_kind(anl_domid, QTY_VAPOR_MIXING_RATIO) < 0)) then
    write(string1, *) 'State vector is missing one or more of the following fields:'
    write(string2, *) 'Potential Temperature (theta), Density (rho), Vapor Mixing Ratio (qv).'
    write(string3, *) 'Cannot convert between height/pressure nor compute sensible temps.'


syha:
Helen, I just want to give a quick input on this. I never ran into this kind of issue before because I never had the same QTY for different mpas_state_variables.  
For your particular example with theta and th2m, I usually defined t2m (temperature, not potential temperature, at 2 m) as QTY_2M_TEMPERATURE. 
Typically we have observations in t2m and q2, so we define QTY_2M_TEMPERATURE and QTY_2M_SPECIFIC_HUMIDITY for them, respectively. But if somebody has a good reason to use th2m instead, I think we can add QTY_2M_POTENTIAL_TEMPERATURE in assimilation_code/modules/observations/atmosphere_quantities_mod.f90.
In a nutshell, we'd encourage to differentiate the QTYs for different mpas variables.

hkershaw-brown:
Thanks Soyoung! 

If at some point you do want to enforce only one of each qty in the mpas domain, then you can use
`get_num_varids_from_kind(domid, QTY) == 1` or you could enforce it in verify_state_variables.

This fix does not change the existing functionally of checking the required variables are in the domain, it just stops the code erroring out if a user has chosen to give multiple variables the same quantity.

syha:
> Thanks Soyoung!
> 
> If at some point you do want to enforce only one of each qty in the mpas domain, then you can use `get_num_varids_from_kind(domid, QTY) == 1`.

That's a good idea. It would be good to include that line.

> 
> This fix does not change the existing functionally of checking the required variables are in the domain, it just stops the code erroring out if a user has chosen to give multiple variables the same quantity.

Thanks for the clarification, Helen! Sounds good to me.

mjs2369:
The updated code looks good. Just a few questions for my own knowledge. 

In what context would one run MPAS with 2 (or more) variables of the same quantity in the state? I don't really see how this could produce correct or meaningful results, unless the two variables of the same qty were the same variable as well and therefore stored identical values in their shared qty. In which case, why not just use one of those variables?

Why weren't we already using get_num_varids_from_kind in the MPAS_ATM model_mod? Why do we want DART to be ok with multiple variables of the required quantities?

When you said on the issue page that the rest of the mpas_atm/model_mod does not have problems with multiple same qtysdue to the progvar structure, does this mean that this problem is contained to the get_varid_from_kind routine? Why was the progvar structure not applied to this portion of the code?

hkershaw-brown:
> In what context would one run MPAS with 2 (or more) variables of the same quantity in the state? I don't really see how this could produce correct or meaningful results, unless the two variables of the same qty were the same variable as well and therefore stored identical values in their shared qty. In which case, why not just use one of those variables?

The new large test case we have uses QTY_POTENTIAL_TEMPERATURE for th2m and theta. 

> 
> Why weren't we already using get_num_varids_from_kind in the MPAS_ATM model_mod? 

I cannot answer this for certain, `get_varid_from_kind` came in this pull request on DART_development(May 2020) https://github.com/NCAR/DART_development/pull/65

My best guess is people were concerned with 'time to new science' vs. refactoring an existing model_mod
If you take a look at the code there are several FIXMEs in the model mod like this:
https://github.com/NCAR/DART/blob/0de0a8aa25ddbec49b3b278a19ad4136eb829322/models/mpas_atm/model_mod.f90#L365-L369

> Why do we want DART to be ok with multiple variables of the required quantities?

Internally, dart is just a vector.  No restriction on quantities.  Have as many variables with the same quantity as you want. 

Reasons why there may be multiple variables with the same quantity:
* multiple grids in the same, e.g. a high res and low res; or pressure on the interfaces, pressures on the levels
* multiple time steps in the same file,
* someone may want to define their state variables a certain way so they can use the obs_impact_tool to stop certain quantities being impacted by certain observations. 

_required_ quantities is a mpas_atm model model_mod restriction.
The restriction in the current mpas_atm (and this pull request) is that 
QTY_POTENTIAL_TEMPERATURE,   QTY_DENSITY, QTY_VAPOR_MIXING_RATIO exist in the mpas state.

There is not currently a restriction in the mpas_atm model_mod code that enforces 1 variable of each quantity.  From Soyoung the 1 variable of each quantity is convention, but it is not enforced in the code. 

> 
> When you said on the issue page that the rest of the mpas_atm/model_mod does not have problems with multiple same qtysdue to the progvar structure, does this mean that this problem is contained to the get_varid_from_kind routine? Why was the progvar structure not applied to this portion of the code?

I'm going to flip this question around.  Why is progvar in the model_mod a problem?

I was not precise enough in my note in https://github.com/NCAR/DART/issues/386, we don't error out because of multiple quantities anywhere else in mpas_atm model_mod, because the code does stuff like this: this function returns when it finds the first kind
https://github.com/NCAR/DART/blob/0de0a8aa25ddbec49b3b278a19ad4136eb829322/models/mpas_atm/model_mod.f90#L4295-L4312
Not erroring out and doing what is intended is not the same, right? 

The progvar structure is an upcoming problem, particularly this bit:

```
   integer :: index1        ! location in dart state vector of first occurrence
   integer :: indexN        ! location in dart state vector of last  occurrence
```

because the model_mod is assuming the structure of the state.  Similarly mpas_atm model_mod also seems to be doing its own clamping. 

https://github.com/NCAR/DART/blob/0de0a8aa25ddbec49b3b278a19ad4136eb829322/models/mpas_atm/model_mod.f90#L2763-L2771

Issue #394, opened 2022-09-08
CLM free run shell scripts cannot provide single instance run


braczka:
### Describe the bug

1. List the steps someone needs to take to reproduce the bug.  

Executing either `CLM5_hybrid_freerun` or `CLM5_startup_freerun` with number of instances = 1.  The CLM case is created and build successfullly, but case fails when executing ./case.submit.

Generally these scripts are used with multi-instance capability, but sometimes when testing configurations of CLM it is useful to generate single instances.

2. What was the expected outcome?

For the setup scripts to be able to successfully create a single instance run, and then that run could be executed.

3. What actually happened?  

The data stream files are not created correctly within the CLM run folder.  CLM expects a multi-instance run with the instance number appended to the stream files.  A single instance run does not require multi-instance capability, nor appended instances to stream files.

### Error Message  

Data stream files cannot be located.

### Which model(s) are you working with?
cesm2.2

### Version of DART      
latest version

### Have you modified the DART code?   
No -- but fix could include following similar setup as using the `CLM5_setup_pmo` scripting which uses single instance capability in case setup, and removes need for instance number to be appended data streams files.

Issue #393, opened 2022-09-08
CLM shell scripts give dtlimit error for single year runs


braczka:
### Describe the bug

1. List the steps someone needs to take to reproduce the bug.  
Executing any of the CLM shell scripts at models/clm/shell_scripts/cesm2_2, in which a single year simulation is performed.
Scripts include `CLM5_hybrid_freerun`,  `CLM5_startup_freerun`, `CLM5_setup_assimilation`, and  `CLM5_setup_pmo`.

2. What was the expected outcome?
For the CLM simulation to run the complete year, run to completion.

3. What actually happened?  

CLM error occurs on the final day of year (12/31)

### Error Message  

A 'dtlimit' error occurs within CLM -- which is an indicator that meteorology forcing timesteps are missing.  This occurs because
the CAM6 annual reanalysis files have a 6 hour shift, where last 6 hours in current year exist in the following years file.  For a single year run- only that CAM6 annual reanalysis file is listed in datm namelist -- thus CLM is forced to interpolate.
This can be corrected by loading all CAM6 reanalysis years into datm namelist.

### Which model(s) are you working with?
CESM2.2

### Version of DART      

latest version

### Have you modified the DART code?   
Not yet -- but the fix is straightforward, to use the multi year streamfiles (datm.streams.txt.CPLHISTForcing*.complete) in place of the single year streamfiles (datm.streams.txt.CPLHISTForcing*single_year) .

PR #396, opened 2022-09-13
Clm shell scripts

braczka:
## Description:

This pull request removes the option to load up a single year of CAM reanalysis within the datm data stream files, and replaces that with using the complete set of CAM reanalysis years.  This removes the potential for a 'dtlimit' error when performing a single year simulation within CLM.  A dtlimit error is the way that CLM identifies missing time steps within the atmospheric forcing data.   In reality, there are no missing time steps within the CAM reanalysis, only a 6 hour shift in the time steps (e.g. the final 6 hours of year 2011 are located in 2012 file).   Loading all the CAM reanalysis years eliminates the possibility of this error, with a negligible increase in computation time.  This change was imposed for all shell scripts, including freerun, pmo and assimilation scripts.

This pull request also fixes a bug where single instance runs cannot be performed with the 'freerun' CLM shell scripts.   Performing single instance simulations (even with scripts designed to perform multi-instance simulations) is useful when troubleshooting various CLM setups without DART, and minimizes computation time.

### Fixes issue
Fixes  https://github.com/NCAR/DART/issues/393 dtlimit error issue
Fixes https://github.com/NCAR/DART/issues/394  single-instance error

### Types of changes

- [ ] Documentation update

I have provided no changes to documentation because it fixes a bug.

### Tests

I have run single and multiple-instance ensemble runs for `CLM5_hybrid_freerun` and `CLM5_startup_freerun` for a full year.    I have also run these same scripts for monthly increments of time to test the stage_cesm_files script functionality for continuation runs  (CONTINUE_RUN=TRUE)

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Version tag 

## Testing Datasets

- [x] No dataset needed


hkershaw-brown (commented):
Hi Brett, great fixes. 

Anything critical you want looking at in the review?

braczka:
Hi Helen --- Nothing very specific that I can think of, these are very CLM specific changes, not really DART related.   I never did any formal testing on computation differences between using single reanalysis years and all CAM reanalysis years in setup scripts -- the assumption is that it is negligible when interacting with CLM.   I think that's fair and not worth investigating further.....

models/clm/shell_scripts/cesm2_2/CLM5_startup_freerun
```diff
@@ -280,57 +288,32 @@ while ( $inst <= $num_instances )
...
+     foreach FNAME ( user_datm.streams.txt* )
+        echo "modifying $FNAME"
+        sed s/NINST/0001/g $FNAME >! temp
+        sed s/RUNYEAR/${stream_year_first}/g temp >! $FNAME
+     end
+   else
    foreach FNAME ( user_datm.streams.txt*${inst_string} )
       echo "modifying $FNAME"
-      sed s/NINST/${inst_string}/g $FNAME >! temp
+        sed s/NINST/${inst_string2}/g $FNAME >! temp
```
hkershaw-brown, line 309: indent

```suggestion
      sed s/NINST/${inst_string2}/g $FNAME >! temp
```

models/clm/shell_scripts/cesm2_2/CLM5_hybrid_freerun
```diff
@@ -468,21 +486,30 @@ else     # CONTINUE_RUN == FALSE
 
    @ inst=1
    while (\$inst <= $num_instances)
+     
+      if ( \${num_instances} == 1) then
+         set inst_string = \`printf _%04d \$inst\`   
+         set inst_string2 = ''
+      else
+         set inst_string = \`printf _%04d \$inst\`
+         set inst_string2 = \`printf _%04d \$inst\`
```
hkershaw-brown, line 495: Unless I am mis-reading this (always possible)`inst_string` and `inst_string2` are identical.

I would just have `inst_string` if they are identical.

hkershaw-brown (changes requested):
Hi Brett, looks good. 

One nitpicking indenting suggestion, and then one place where it looks like you have inst_string and inst_string2 set to the same thing.

Also, note I went ahead an updated this branch to main (Ben's bug fix)

Cheers,
Helen

models/clm/shell_scripts/cesm2_2/CLM5_hybrid_freerun
braczka: Yes -- inst_string and inst_string2 are identical here, but only for the multi-instance case (see 'if' statement in lines 490-96).   On the other hand, In a single instance case these strings are different, because in a single instance case there is no appending of the instance number within the restart file name for the first time step.  Agreed --  it's confusing.    Other option would be to give multiple 'LINK' commands in a if statement, but that would generate more code.

hkershaw-brown: my bad, looks good as is. I'd missed the single instance case.

braczka:
Helen --  thanks -- should I go ahead and update the Changelog and then bump the tag to 10.3.2 on this?

hkershaw-brown:
Yup, go ahead and release it!

PR #418, opened 2022-10-27
move window creation and destruction for perfect_model_obs

### LINKED ISSUE #417: bug: Running perfect_model_obs in parallel
:bug: Apologies if this has been reported. I couldn't find any past issues similar to this one. 

### Describe the bug
Running `perfect_model_obs` in parallel and serially produce different `obs_seq.out` files. Though, they both create the same truth file: `true_state.nc`. I ran into this issue while running hundreds of experiments using the lorenz_96 model. The assimilation results only make sense when using the `obs_seq.out` produced by the serial `perfect_model_obs`. One other interesting fact of this issue: If you compare `obs_seq.out` from both serial and parallel runs, you'll find that the first cycle has exactly the same truth and observation values. Differences only start to appear for subsequent cycles (see attached screenshot). 

Not sure if this is a known behavior or whether it's related to my compiler. I vaguely remember someone telling me not to run `perfect_model_obs` in parallel in the bgrid model. But that was long time ago and not lorenz_96. 

### Error Message  
No error

Lorenz 96

### Screenshots   
Screen Shot 2022-10-26 at 4 51 28 PM

### Version of DART      
v10.5.1

No  

### Build information
 1. The machine you are running on: NCAR supercomputer Cheyenne.   
 2. The compiler you are using: intel.

hkershaw-brown:
## Description:
Perfect_model_obs was giving incorrect obs_seq.out (forward operators) when the  model was advanced in perfect_model_obs **and** perfect_model_obs was run with multiple processors (using the default namelist option of `distributed_state = .true.`)

This pull request moves the mpi window creation and destruction for perfect_model_obs inside the AdvanceTime do loop. The forward operator needs to use state from each step of the loop, so the window must be recreated.

Why do we need to create a window each time? Because currently the mpi window is created using a copy of the state. 

Edit:
This pull request also removes the namelist option `distribute_state` from pmo.
The perfect_model_obs code is not setup to do non-distrbuted state foward operators - the obs loop is around my_num_vars. Compare this to foward_operator_mod::get_obs_ens_distributed_state which has an if statement for transposed or distributed state forward operators. 

### Fixes issue
fixes #417

### Types of changes

- [ ] Documentation update

### Tests

Lorenz_96 perfect_model obs. serial, and mpirun n=4 give the same obs_seq.out and same true_state.nc

## Checklist for merging

- [x] Updated changelog entry
- [ ] Documentation updated
- [x] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed

  ### Notes 
  still to test if `distributed_state = .false.` works correctly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,13 @@ individual files.
 
 The changes are now listed with the most recent at the top.
 
+**November 3 2022 :: Bug-fix release. Tag v10.5.4**
+
+- Perfect_model_obs (pmo) fixed for running with MPI and advancing the
+  model inside pmo.
+- MPAS_ATM xtime string padded with blanks for easier bitwise comparison.
+- lorenz_96_tracer_advection quickbuild.sh fixed.
+
 **October 13 2022 :: Bug-fix for read variables. Tag v10.5.3**
 
 - Per-file check for unlimited dimension before variable read. Netcdf 
--- a/assimilation_code/programs/perfect_model_obs/perfect_model_obs.f90
+++ b/assimilation_code/programs/perfect_model_obs/perfect_model_obs.f90
@@ -79,7 +79,6 @@ integer  :: async              = 0
 logical  :: trace_execution    = .false.
 logical  :: output_timestamps  = .false.
 logical  :: silence            = .false.
-logical  :: distributed_state  = .true.
 
 ! if init_time_days and seconds are negative initial time is 0, 0
 ! for no restart or comes from restart if restart exists
@@ -118,7 +117,7 @@ namelist /perfect_model_obs_nml/ read_input_state_from_file, write_output_state_
                                  trace_execution, output_timestamps,                &
                                  print_every_nth_obs, output_forward_op_errors,     &
                                  input_state_files, output_state_files,             &
-                                 single_file_in, single_file_out, distributed_state
+                                 single_file_in, single_file_out
 
 !------------------------------------------------------------------------------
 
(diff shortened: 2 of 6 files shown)


hkershaw-brown:
models that have adv_1step subroutine.  Should these build serial perfect_model_obs?

lorenz_96
lorenz_96_2scale
bgrid_solo
lorenz_63
lorenz_04
null_model
lorenz_96_tracer_advection
forced_barot
pe2lyr
9var
ikeda
lorenz_84
forced_lorenz_96
simple_advection
dynamo
sqg

mgharamti:
@hkershaw-brown I am ok with making these models only build serial perfect model obs. Also, should we return an error if the user attempts to build pmo in parallel? If we know the answer is going to be wrong, then we definitely need to return with an error.

hkershaw-brown:
> @hkershaw-brown I am ok with making these models only build serial perfect model obs. Also, should we return an error if the user attempts to build pmo in parallel? If we know the answer is going to be wrong, then we definitely need to return with an error.

This pull request fixes the running in parallel error so running pmo with mpi now gives the same answer as running without mpi. 

I was thinking about changing the subroutine runable models to build pmo serial, because there is no speed advantage to running with mpi. But in the interests of paranoia (causing more problems than I solve), I'm backing off making those changes to all those model quickbuilds, and just keep this pull request as the bug-fix.

mjs2369 (approved):
Hi Helen, everything for the bug-fix and the removal of the distributed_state nml option looks great. Approved.

mgharamti (approved):
This looks good. I've just tested the changes in L96 by running pmo both serially and in parallel. I got the same truth netcdf file and more importantly the same `obs_seq.out` file. I see that you've removed `distributed_state` option from pmo's namelist parameters. Is that across the board for all models. Right now, it's only the `input.nml` for POP. Please make sure the change is consistent in other models.

hkershaw-brown:
>I see that you've removed `distributed_state` option from pmo's namelist parameters. Is that across the board for all models. Right now, it's only the `input.nml` for POP. Please make sure the change is consistent in other models.

will do. Thanks Moha!

Issue #381, opened 2022-08-04
bug: cam-fv no_normalization_of_scale_heights   = .false.


hkershaw-brown:
:bug: Your bug may already be reported!
Please search on the [issue tracker](https://github.com/NCAR/DART/issues) before creating a new issue.

### Describe the bug

1. List the steps someone needs to take to reproduce the bug.  

build with debugging flags
cam-fv filter run with
```
   no_normalization_of_scale_heights   = .false.
 ```

Test case is here: /glade/scratch/hkershaw/DART/pull_requests/pull_370/test_cam.main2

3. What was the expected outcome?
Successful run of filter

5. What actually happened?  
error: ptype being used without being defined.

### Error Message  
Please provide any error messages.

```
forrtl: severe (194): Run-Time Check Failure. The variable 'model_mod_mp_obs_vertical_to_scaleheight_$PTYPE' is being used in '/glade/scratch/hkershaw/DART/pull_requests/pull_370/DART.main/models/cam-fv/model_mod.f90(3471,33)' without being defined
Image              PC                Routine            Line        Source             
filter             00000000004B6CF7  model_mod_mp_obs_        3471  model_mod.f90
filter             00000000004B5EE4  model_mod_mp_conv        3302  model_mod.f90
filter             00000000006007E2  assim_tools_mod_m         478  assim_tools_mod.f90
filter             00000000006D6DD0  filter_mod_mp_fil         918  filter_mod.f90
filter             000000000065C15B  MAIN__                     20  filter.f90
filter             00000000004077A2  Unknown               Unknown  Unknown
libc-2.22.so       00002AC391147A35  __libc_start_main     Unknown  Unknown
filter             00000000004076A9  Unknown               Unknown  Unknown


```

In obs_vertical_to_scaleheight

ptype is only defined in the no_normalization_of_scale_heights = .true. part of the if statement, here:
https://github.com/NCAR/DART/blob/0de0a8aa25ddbec49b3b278a19ad4136eb829322/models/cam-fv/model_mod.f90#L3422-L3430

Here is ptype being used in  no_normalization_of_scale_heights = .false. part of the if statement. 

https://github.com/NCAR/DART/blob/0de0a8aa25ddbec49b3b278a19ad4136eb829322/models/cam-fv/model_mod.f90#L3471

I'm not sure you need to check ptype, because VERTISSURFACE obs is checked in the surrounding if statement, so by line 3471 ptype can only be QTY_PRESSURE. 

``` diff
[...]

PR #425, opened 2022-11-09
move code to set ptype variable outside the if() statement

nancycollins:
in the cam model_mod, referencing an unset ptype in the else clause could cause a crash.

## Description:
fixes bug #381.  this update moves the setting of the ptype variable outside the if() statement so it doesn't crash with an unset variable in the else clause.

in the bug report discussion there was mention of other changes but i believe the decision was that more extensive refactoring was outside the scope of this bug and we should fix this immediate issue only.

### Fixes issue
#381

### Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

### Tests
i ran helen's test case and it no longer crashes.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ x] Dataset download instructions included - helen has one in the original bug report.
- [ ] No dataset needed

### DIFF
--- a/models/cam-fv/model_mod.f90
+++ b/models/cam-fv/model_mod.f90
@@ -2077,18 +2077,21 @@ character(len=*), parameter :: routine = 'obs_vertical_to_scaleheight'
 
 ens_size = 1
 
+! if this location is on the surface, use the surface pressure field
+! in the computations below.  otherwise use the 3d pressure field.
+if (query_location(location) == VERTISSURFACE) then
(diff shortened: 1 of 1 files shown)


kdraeder (approved):
This looks like a robust fix to me.

hkershaw-brown:
@hkershaw-brown bundle this into the next release

Issue #426, opened 2022-11-09
bug: setting convert_all_obs_verticals_first = .false. prevents QC=8 from being recorded


nancycollins:
### Describe the bug

if convert_all_obs_verticals_first is set to .false. in the &assim_tools_nml namelist, any failed vertical conversions fail to set QC=8 in the output obs_seq.final file.

to reproduce:
on cheyenne in /glade/scratch/nancy/do_vert_first_test, copy model_mod.f90 to your bgrid_solo dir and copy the input.nml to your bgrid work dir.  
build and run.  
look at the DART QCs in the obs_seq.final file.  every other one should be 8 and is not.
set convert_all_obs_verticals_first to .true. in the input.nml and rerun.  the output is as expected.

### Version of DART    
release 10.5.5.

the bug with skipping posterior forward operators and not seeing QC=8 is an unrelated problem and that fix doesn't change these results.  

### Have you modified the DART code?   
i haven't changed the code.  my recommended modification is to require convert_all_obs_verticals_first to remain true.  it eventually should be removed from the namelist but that's quite disruptive.  the default is already true and i suspect hardly anyone is actually changing this to false.   but to prevent unexpected behavior, i suggest adding a test and error out if it has been set to false, including a message that this namelist item is deprecated and will be removed at some future time.

### Build information
i ran this on cheyenne with the intel compiler.

nancycollins:
the alternative to removing this option is to fix the code in the assimilation loop so it calls convert_vertical_obs() explicitly before doing the broadcast of values and setting QC=8 if the convert fails at that point.

hkershaw-brown:
closing this as won't fix. QC 8 is a failure from vertical conversion QC4 is from failure in the forward operator call.

PR #427, opened 2022-11-09
do not allow convert_all_obs_verticals_first to be false

nancycollins:
drop support for letting get_close do lazy vertical conversion on observations.   test if this namelist item is false and error out.  the default is true so users setting this to false seems unlikely.  but it shouldn't be left there because it results in wrong DART QCs in the output file.   eventually this option should be deprecated and removed from the namelist.

## Description:

add a check for a user setting this namelist false and error out with a message.

### Fixes issue
#426

### Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

### Tests
set convert_all_obs_verticals_first to false in the namelist and run filter on anything.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x ] No dataset needed

### DIFF
--- a/assimilation_code/modules/assimilation/assim_tools_mod.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.f90
@@ -292,6 +292,15 @@ endif
 
 is_doing_vertical_conversion = (has_vertical_choice() .and. vertical_localization_on())
 
+! this option is deprecated and will be removed in some future release.
+if (.not. convert_all_obs_verticals_first) then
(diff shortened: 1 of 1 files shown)


hkershaw-brown:
closing this, I don't think this is a settled issue about the correct thing to do here.

Issue #404, opened 2022-10-04
Feature request: Refactor wrf model_mod


hkershaw-brown:
### Use case  

Refactor the wrf model_mod to make it easier/possible to:
* add state variables
* swap between wrf v3 and v4
* not assume the structure of the state variable in model_mod.

### Is your feature request related to a problem?  
 
The WRF model_mod has `wrf_static_data_for_dart` which is replicating  functionality available in state_structure_mod. 

* Adding variable X to the state means altering the model_mod code to have a `wrf%dom(id)%type_X`, rather than simply adding a line in the model_nml::wrf_state_variables.  
* WRF has its own integer types (qtys), replicating (but not quite) the DART qty functionality.  
https://github.com/NCAR/DART/blob/7ee9ed9c86330ba231fa29395dbb2db47d3a7e80/models/wrf/model_mod.f90#L323-L332
so you see code like this:
https://github.com/NCAR/DART/blob/7ee9ed9c86330ba231fa29395dbb2db47d3a7e80/models/wrf/model_mod.f90#L889-L890
* WRF v3 vs. v4 - it is not clear for users what to change when moving from WRF v3 to v4. If possible, it would be good for the model_mod to be able to check the WRF version from the netcdf file. 

* Wider issue #389. Moving control of the state vector to core dart modules allows us to have general solutions for things like compression, removing missing values (so dart core code can assume every element of the state is valid), and other cool stuff.

### Describe your preferred solution  
Remove the state vector information from wrf_static_data_for_dart

### Describe any alternatives you have considered  
I think the wrf_static_data_for_dart is a road block for core dart development, however wrf-dart is widely used and has a lot of additional user developed code in circulation. There is 8000 lines of code in for wrf model_mod (+ unknown user modifications) so the refactoring needs to be approached with this in mind. 
I think it is possible to simplify the bounds checks for each grid (staggered and various flavors of periodic), but I do not think the bounds check affects development of core dart.

hkershaw-brown:
https://github.com/hkershaw-brown/DART/tree/wrf-refactor

hkershaw-brown:
wrf model_mod get_close has a test for missing_r8

https://github.com/NCAR/DART/blob/72e5740b71ef6c7cde4c6f946e2c1df2019365fb/models/wrf/model_mod.f90#L6419C3-L6428

should check for verti_is_undefined before doing this (or not have the missing_r8 check)

hkershaw-brown:
[...]

PR #510, opened 2023-07-10
bug-fix: get_dist calculation for observations with VERTISUNDEF in WRF

### LINKED ISSUE #486: bug: WRF observations with VERTISUNDEF
:bug: :ant: Your bug may already be reported!

### Describe the bug

User reported using WRF and radiance obs.

Run WRF with an observation that has VERTISUNDEF for the vertical location
The distance calculated in get_close should be the horizontal distance whether you have the 
```
&location_nml 
  horiz_dist_only =.true./.false.
```
3. What actually happened?  
The distance calculated changes for the VERTISUNDEF case depending on `horiz_dist_only =.true./.false.`

### Error Message  
No error, incorrect distances

WRF
User is working with WRF v4

### Screenshots   
Screen Shot 2023-05-31 at 10 36 10 AM

### Version of DART      
v10.7.2-2-ga56916533

Yes
If your code changes are available on GitHub, please provide the repository.
To reproduce the bug I am using a WRF v4 test case provided by the user. This requires a change to the WRF state (THM vs T in the wrf/model_mod.f90) 
see https://github.com/NCAR/DART/compare/main...lukas-vertundef-wrf

### Build information
Please describe:  
 1. mac laptop 
 2. intel (and gnu for kicks)

## Notes
I'm unclear yet on whether setting `whichvert_obs_in_localization_coord = 0` is a problem. 
https://github.com/NCAR/DART/blob/72e5740b71ef6c7cde4c6f946e2c1df2019365fb/assimilation_code/modules/assimilation/assim_tools_mod.f90#L592-L593

0 is undefined as a vertical localization coord for the 3d sphere lcoation_mod:
https://github.com/NCAR/DART/blob/72e5740b71ef6c7cde4c6f946e2c1df2019365fb/assimilation_code/location/threed_sphere/location_mod.f90#L59-L67

hkershaw-brown:
## Description:

Previously observations with VERTISUNDEF were given a large distance, because the vertical coordinate is missing: local_array(3) == missing_r8

https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/wrf/model_mod.f90#L6423-L6424

The location_mod::get_dist uses horizontal distance only if VERTISUNDEF for either location.

### Fixes issue

Fixes #486 

### Types of changes

- [ ] Documentation update

### Tests
Tested using an example case from Lukas K. 
#486 and https://github.com/NCAR/DART/compare/main...lukas-vertundef-wrf
Note lukas-vertundef-wrf has changes for wrf v4, this pull request only has the VERTISUNDEF bug fix.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/models/wrf/model_mod.f90
+++ b/models/wrf/model_mod.f90
@@ -6403,6 +6403,13 @@ if (istatus1 == 0) then
       local_loc   = locs(t_ind)
       local_which = nint(query_location(local_loc))
 
+      if (present(dist)) then
+         if (local_which == VERTISUNDEF) then
+            dist(k) = get_dist(base_loc, local_loc, base_type, loc_qtys(t_ind))
+            cycle
+         endif
+      endif
+
       ! Convert local vertical coordinate to requested vertical coordinate if necessary.
       ! This should only be necessary for obs priors, as state location information already
       ! contains the correct vertical coordinate (filter_assim's call to get_state_meta_data).


mjs2369:
@hkershaw-brown the general fix here looks great, which adds the check for VERTISUNDEF obs that calculates the 2D dist before converting the local vert coordinate. 

I have a few logistical questions: 

The first is about this comment ( https://github.com/NCAR/DART/issues/486#issuecomment-1584848283 ) where you explain why you changed the value of dist for testing. Should we maybe keep this change, even outside of testing, so that we have a different value for initialization and missing/failed vert coordinates?

And the second is just revisiting a question that you initially asked in the notes of the issue, which is whether or not we should set whichvert_obs_in_localization_coord = 0 here - https://github.com/NCAR/DART/blob/72e5740b71ef6c7cde4c6f946e2c1df2019365fb/assimilation_code/modules/assimilation/assim_tools_mod.f90#L593 

As you noted it doesn't match any of the values in the 3D sphere location_mod:
should this be -2 to match the VERT_IS_UNDEF parameter values ( https://github.com/NCAR/DART/blob/72e5740b71ef6c7cde4c6f946e2c1df2019365fb/assimilation_code/location/threed_sphere/location_mod.f90#L62 )? The ability to do this is dependent on whether or not setting the value to -2 will break the other location mods other than 3D sphere, but it's hard to tell because most location mods either don't have this set of parameters or only use a select number of them (like here https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/assimilation_code/location/threed_cartesian/location_mod.f90#L50 in 3D cartesian)

On the flip side, could we potentially change the parameter value for VERTISUNDEF to 0 for in threed_sphere/location_mod.f90?

hkershaw-brown:
> The first is about this comment ( https://github.com/NCAR/DART/issues/486#issuecomment-1584848283 ) where you explain why you changed the value of dist for testing. Should we maybe keep this change, even outside of testing, so that we have a different value for initialization and missing/failed vert coordinates?

I thought about this too.  I was using a debugger to trap on various values, which i think nice debugging technique, but
I think if you start putting extra information into the distance value, e.g.  
  dist (output from get_close) is both and distance and an indication of why the distance is the value it is,  
 you're breaking the contract between get_close and the code that calls get_close.

Note there is an upcoming project to simplify a lot of the wrf model_mod https://github.com/NCAR/DART/issues/404 and part of that is to not have so many if statements in get_close.

> whichvert_obs_in_localization_coord = 0

VERTISUNDEF  is in a world where the vertical dimension exists, but the observation is not at one point in the vertical (e.g. maybe it is the integral of a whole column of atmosphere) 
vs. in a one dimensional or two dimensional world (e.g. [flatland](https://en.wikipedia.org/wiki/Flatland)) where there is no concept of the vertical dimension. 

I think `whichvert_obs_in_localization_coord` is just set to something because it is broadcast (the task that owns an observation does vertical conversion and broadcasts the result out along with all the other stuff that task calculates). 
https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/assimilation_code/modules/assimilation/assim_tools_mod.f90#L613-L617

but `whichvert_obs_in_localization_coord` is never used if the model_mod does not a have a vertical dimension:
https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/assimilation_code/modules/assimilation/assim_tools_mod.f90#L293
I'm guessing that `0` was picked because it is not a valid value for the vertical so if you start to use it, it should trigger and error. 

I don't think `whichvert_obs_in_localization_coord` is great for readability in assim_tools_mod (since it confused me! hehehe).  But I'd push any refactoring of it into the future when we have a good look at assim_tool_mod.

mjs2369 (approved):
Sounds good. Like you said, we can revisit any of these issues when we refactor. Approved.

PR #522, opened 2023-07-28
Updates for w

syha:
## Description:
Major updates for w obs (for radial velocity) and a bug fix for vertical interpolation in the vertical localization in height. Zgrid is now considered at either full or half levels (vertically), and at the center or edge (horizontally), just like in the model. Some cleanups and clarifications are also made. 

### Fixes issue
The mpas_atm model uses a height vertical coordinate with the height defined at full levels (e.g., zgrid). Most prognostic variables (ex. theta, rho, and qv) are defined at half levels, which are calculated as the middle points between the full levels. Only w (vertical velocity) is defined at full levels. 
The model_mod has previously only checked "oncenters" to differentiate cell center vs edges for zgrid, but this only applies to horizontal location differences at the same half levels. Now, to support QTY_VERTICAL_VELOCITY, we added "onHalf", checking full vs. half levels, as well. Along the line, we found a bug related to the vertical interpolation in height and fixed it.

### Types of changes

- [ ] Documentation update

- [ ] I have updated the comments accordingly.

### Tests
The new model_mod.f90 has passed the test with a 2-member ensemble and a single obs.

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets
- [ ] No dataset needed


hkershaw-brown:
> Along the line, we found a bug related to the vertical interpolation in height and fixed it.

Can you describe the bug you found? What was the problem, what was the fix?

models/mpas_atm/model_mod.f90
```diff
@@ -661,29 +661,30 @@ subroutine static_init_model()
...
+            zGridEdge(kloc,iloc) = (zGridCenter(kloc,cel1) + zGridCenter(kloc,cel2))*0.5_r8
          else if (cel1>0) then
-            zGridEdge(iloc,kloc) = zGridCenter(iloc,cel1)
+            zGridEdge(kloc,iloc) = zGridCenter(kloc,cel1)
          else if (cel2>0) then
-            zGridEdge(iloc,kloc) = zGridCenter(iloc,cel2)
+            zGridEdge(kloc,iloc) = zGridCenter(kloc,cel2)
          else  !this is bad...
-            write(string1,*)'Edge ',kloc,' at vertlevel ',iloc,' has no neighbouring cells!'
+            write(string1,*)'Edge ',iloc,' at vertlevel ',kloc,' has no neighbouring cells!'
```
hkershaw-brown, line 687: what's the goal of renaming (swapping) the loop variables for these two do loops? 
The functionality is the same.

syha:
Sure. In Lines 4689-4690 in the main branch, fdata should not be always computed from zGridFace (which is not zGridFull, following the model's definition). It was supposed to be zGridCenter because we only considered half-level cases, so far. Now we split up the case for both half and full levels.

models/mpas_atm/model_mod.f90
syha: Yes, the model indices are (k,j,i), from the wrf days. So we tend to think of k as vertical index. Even within this code, we have "vloc" at the same first index, so I think it's more intuitive to go with "kloc" there.

hkershaw-brown:
There are a couple of name changes to variables, so it looks like there is a lot of code change here

hkershaw-brown (changes requested):
no not merge as is.

hkershaw-brown:
> Sure. In Lines 4689-4690 in the main branch, fdata should not be always computed from zGridFace (which is now zGridFull, following the model's definition). It was supposed to be zGridCenter because we only considered half-level cases, so far. Now we split up the case for both half and full levels.

To clarify this is not a bug in the main branch, because no variables were on the zGridFull (zGridFace)?

syha:
The model variable 'zgrid' was read as zGridFace in the main branch. Then it was used to compute zGridCenter, which was used for theta, qv, etc. So the lines 4689-4690 should have used zGridCenter, not zGridFace (which is now zGridFull). It was a bug, for sure, because find_vert_indices (above the lines) found vertical indices in zGridCenter.

Issue #520, opened 2023-07-27
bug: obs_window_days and obs_window_seconds are no-ops


nancycollins:
:bug:
### Describe the bug
Somewhere in the translation of the documentation from html to rst, the info about perfect_model_obs and filter ignoring the obs_window_days and obs_window_seconds namelist items got lost.   the docs should be updated to say these are not yet functional, and the code should warn the user if they try to use it.

(eventually i think they could be implemented and be useful, but that's a slightly larger task needing a lot of testing.)

it would be quick to at least flag that these aren't implemented yet.   the default values of these namelist items are -1.  in the filter_mod, in filter_set_window_time(), if the values aren't both -1, it could print a warning or error saying this option is not yet supported.  that would at least prevent someone from using them and thinking they were working.

to reproduce, set obs_window_days and obs_window_seconds to something other than -1.     nothing changes in which obs are assimilated.

another thing i'd fix - the filter_set_window_time() routine only tests obs_window_days to see if it is >=0.   it should also look at obs_window_seconds, i'd think.  both default to -1 but someone could leave days as -1 and try only setting seconds.  that should be a real error.  (same for any time pair of days/seconds.)

i'd be willing to put in code to at least catch this and print a warning or error (your vote here?) if you think that's a good thing to do.  let me know?

### Error Message  
no errors are printed - the settings are silently ignored. 

### Which model(s) are you working with?
any model.

### Version of DART      
Which version of DART are you using? 
You can find the version using `git describe --tags`  

### Have you modified the DART code?   
Yes/No  
If your code changes are available on GitHub, please provide the repository.

### Build information
Please describe:  
 1. The machine you are running on (e.g. windows laptop, NCAR supercomputer Cheyenne).   
 2. The compiler you are using (e.g. gnu, intel).

braczka:
@nancycollins thanks for looking at this.  I would be in favor, as you say, to print a warning message if either `obs_window_days` or `obs_window_seconds` are set to values other than -1.   Also the current dart documentation for these namelist settings in either `perfect_model_obs` or `filter` don't indicate (to me at least, maybe that's what 'reserved for future use' means?) that they are non functional.  In the short term maybe make that more explicit in the DART docs.

nancycollins:
ok - i will add some code to print warning messages if these are set to something other than -1 -- or i can make it a fatal error just to be safe?   and i will change the docs to say they are only reserved for future use and currently should not be used.

PR #525, opened 2023-08-02
make clear obs_window_seconds and obs_window_days are unsupported

nancycollins:
## Description
both filter and perfect_model_obs have obs_window_days and obs_window_seconds in their namelists.  these are for future use and are currently unsupported.  update the docs to be more clear about this and update the code to print out a warning message if they are changed to be >= 0.

also check both the days and seconds values for obs_window_xxx and also init_time_xxx.  if either are >= 0, call set_time(). that will catch and print a fatal error if one is set and the other is left at -1.  the old code only tested days, so if someone set seconds >= 0 but left days negative the init time would have been set to 0, 0 and the namelist values wouldn't have been used without any message.

### Fixes issue
#520

### Types of changes

- [x] Documentation update

### Tests
I ran Lorenz 96 pmo and filter with various combinations of values in the namelist for both init_time_xxx and obs_window_xxx and it gave me the results i expected.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed

### DIFF
--- a/assimilation_code/modules/assimilation/filter_mod.f90
+++ b/assimilation_code/modules/assimilation/filter_mod.f90
@@ -170,7 +170,9 @@ integer  :: first_obs_days      = -1
 integer  :: first_obs_seconds   = -1
 integer  :: last_obs_days       = -1
 integer  :: last_obs_seconds    = -1
-! Assimilation window; defaults to model timestep size.
+! Assimilation window; default is model timestep size. 
+! Changing these is currently unsupported/not implemented. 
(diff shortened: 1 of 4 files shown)


hkershaw-brown:
closing, 
* the original user reporting problems was getting a hang because of this bug https://github.com/NCAR/DART/issues/535

PR #538, opened 2023-09-02
Enable restriction of an even-sphere obs network to a lon-lat box

kdraeder:
## Description:
This is a new version of even_sphere.m, modified to restrict the observations to a lon-lat box,
but maintain the evenly-spaced pattern produced by the "golden section spiral".
It was needed to create an obs network restricted to the Tropics for a Zagar OSSE.
I generalized it to allow restrictions of longitude as well.
I did some refactoring to reduce the number of similar loops and gather closely-related tasks.

### Fixes issue
None (yet)

### Types of changes

- [x] Documentation update

The matlab script can print a "help" section, which now has an example of restricting the domain.

### Tests
I specified a subdomain that spans the prime meridian and the equator and examined the picture of the profile locations.

### Reviewers
 I don't know who knows Matlab and would like to review this.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed

### DIFF
--- a/observations/obs_converters/even_sphere/even_sphere_subdom.m
+++ b/observations/obs_converters/even_sphere/even_sphere_subdom.m
@@ -1,0 +1,303 @@
+function even_sphere_compressed(nprofiles, varargin)
+
+% Generate approximately evenly-distributed profiles on a sphere using Golden Section spiral algorithm
+%
+% This creates input to create_obs_sequence that will generate a set of radiosonde
+% profiles of T,U,V, that are approximately evenly-spaced on a sphere.
+% A subset of obs in a lon-lat box may be selected, but the spacing will be derived 
+% from nprofiles *on the whole sphere*.
+% An optional argument 'fill_obs' will trigger the creation of synthetic
+% observations complete with (silly) observation values.
+%
+% The vertical coordinate system is 'pressure'.
+% The pressure levels themselves may be input.
+% The observation error variances for each level must then also be input.
+% There are separate error variances for T and [U,V].
+%
+% Example 1: 30 profiles, 14 default levels and obs error variances
+% nprofiles   = 30;
+% even_sphere_subdom(nprofiles)
+%
+% Example 2: 30 profiles at 6 specified levels, 
+%            uses the default obs error variances for each layer
+% nprofiles   = 30;
+% levels      = [1000  850  500  300  200  100];
+% even_sphere_subdom(nprofiles, 'levels', levels)
+%
+% Example 3: 30 profiles at 6 specified levels, 
+%            specify the error variances for each layer 
+% nprofiles   = 30;
+% levels      = [1000  850  500  300  200  100];
+% T_error_var = [1.44 0.64 0.64 0.81 1.44 0.64];
+% W_error_var = [1.96 2.25 4.41 9.00 7.29 4.41];
+% even_sphere_subdom(nprofiles, 'levels', levels, ...
+%             'T_error_var', T_error_var, 'W_error_var', W_error_var)
+%
+% Example 4: restrict profiles to a horizontal subdomain (which spans the prime meridian).
+% nprofiles   = 2000;  only some of these will be written to the file.
+% lon_left = 300.; left boundary
+% lon_right =  30.; right boundary
+% lat_south = -20.; southern boundary
+% lat_north =  20.; northern boundary
+
+% 
+
+%% DART software - Copyright UCAR. This open source software is provided
+% by UCAR, "as is", without charge, subject to all terms of use at
+% http://www.image.ucar.edu/DAReS/DART/DART_download
+
+%%--------------------------------------------------------------------
+% Decode,Parse,Check the input
+%---------------------------------------------------------------------
+
+p = inputParser;
+
+addRequired(p,'nprofiles',@isnumeric);
+
+% obs types to create at each location.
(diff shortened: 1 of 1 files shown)


hkershaw-brown:
Hi @kdraeder I haven't looked too closely at this yet, but this looks like some added options to even_sphere.m.

Is the goal to replace the existing even_sphere.m so we have one .m file that can creates profiles on the whole sphere, or a restricted horizontal area of the sphere?

kdraeder:
It could be a replacement, but I saw that there is already a variant; even_sphere_pe2lyr.m.
So I thought maybe the philosophy here might be to have specialized scripts.
There are no log entries for even_sphere_pe2lyr.m, so it's not clear.
Generally I'm in favor of minimizing code and having fewer programs that do more things,
but I've been outvoted in the past.  So it's your call.

observations/obs_converters/even_sphere/even_sphere_subdom.m
```diff
@@ -0,0 +1,303 @@
...
+default_nlevels     = length(default_levels);
+default_YMD         = '2017-12-25';
+default_nlon         = 288;
+default_nlat         = 192;
+default_lon_left         =   0.;
+default_lon_right         = 360.;
+default_lat_south         = -90.;
+default_lat_north         =  90.;
+
+if (exist('inputParser/addParameter','file') == 2)
```
nancycollins, line 78: it would help if you could add a comment about why one section needs to use addParameter and the other needs to use addParamValue - it's a matlab thing i guess but it's not clear to me what the diff is.

observations/obs_converters/even_sphere/even_sphere_subdom.m
```diff
@@ -0,0 +1,303 @@
...
+    lon(k) = atan2(y(k), x(k)) + pi;
+    lat(k) = asin(z(k));
+end
+
+% Now convert latitude and longitude to degrees and restrict to the subdomain.
+
+num_plot = 0;
+for k = 1:nprofiles
+    % Input is in degrees of latitude and longitude; convert from radians
+    lon_deg(k) = rad2deg(lon(k));
```
nancycollins, line 182: so lon_deg and lat_geg seem to be global, and plot_lon and plot_lat are box bounded.  but when you're writing out the obs, they seem to be using the lon_deg/lat_deg global values?  (line 245).  am i missing something?

nancycollins:
if all the box does is restrict the output to that box and doesn't change where the points are generated, then the obs_sequence_tool has a box option to cull obs outside the box.  you could run the original code unchanged and then cut the obs down as a post processing step.

hkershaw-brown:
@kdraeder let's put your changes into even_sphere.m  
Leave even_sphere_pe2lyr.m as out of scope for this pull request

`git mv even_sphere_subdom.m even_sphere.m`  should replace even_sphere.m with your new changes.

observations/obs_converters/even_sphere/even_sphere_subdom.m
kdraeder: I don't know the difference either.  I was just following the template.
I'd need to dig into Matlab to figure it out.  I'll do it if it's important.

kdraeder:
@nancycollins comment about obs_sequence_tool reminds me that in Manhattan
we'll be using the preprocessor strategy for cam.  I'd forgotten that, as I was importing
this from the reanalysis branch where I needed it and there's no preprocessor step.
Creating the global obs set and extracting a subset using obs_sequence_tool
in the preprocessor step looks like the more consistent strategy.

So it appears to me now that we should leave even_sphere.m unchanged,
ignore even_sphere_subdom.m, and close this PR as "won't do" (?).
Sorry for the wasted effort.

hkershaw-brown:
@kdraeder no problem, do you want to switch the PR from main to the reanalysis branch to have a record of what what used?

observations/obs_converters/even_sphere/even_sphere_subdom.m
nancycollins: am still curious if this is a bug or if i'm not reading the code right.  when you used this did you only generate obs inside the specified box?  (or maybe not of interest anymore if we're going with the obs_sequence_tool extraction route.)

nancycollins:
even_sphere used to be a simple matlab script.  now it's a function - but you have to add a bunch of arguments to get it to run at all?  that doesn't seem good.  can't there be reasonable defaults in the script?  i can't figure out how to run it at all.  we should at least have an example matlab script that calls it with reasonable values.  someone who knows too much matlab made this pretty hard to use, in my opinion.  </end rant>

nancycollins:
ok  - i found you can run it with even_sphere(100) to gen 100 obs.  but i tried restricting it with the lat/lon box and it didn't generate obs inside the box.  the plot looks like the box you specified, but the locations in the output file aren't the same.  so we definitely don't want this version on main, and if this isn't the exact version kevin used, you don't want it on the reanalysis branch either.

kdraeder:
Agreed.  It's broken.  It's not needed.  I'm abandoning it.

jlaucar:
Kevin,
Let's chat about this tomorrow. Let it go for this evening. Jeff

PR #635, opened 2024-02-09
New interface to the Aether lat-lon model

### LINKED ISSUE #559: Develop Aether lon-lat interface
### Use case  
Collaboration with Aaron Ridley (U Mich) requires development of an interface to his newly developed Aether upper atmosphere model.  The first interface will be to the logically rectangular lon-lat grid version.
Later we will develop an interface to the cubed-sphere version.

We plan to use the tiegcm model_mod as a template representing DART's best practices.
We'll import pieces from gitm, which very closely match the Aether lon-lat restart file structure.
Checkout the "aether" branch to contribute work or see its status.
There are also rst files describing parts of the model and interface, written by @johnsonbk,
which have not been committed yet. 

Developing a model_mod, etc., from scratch.  That seems inefficient, given the existence of a gitm interface,
which is closely related to Aether, and the tiegcm interface, which is a recently updated interface 
to another upper atmosphere model.

kdraeder:
## Description:
Add a new interface to enable assimilation with the lat-lon formulation of the Aether space weather model.

### Fixes issue
Fixes #559

### Types of changes

- [x] Documentation update

  - [ ] documentation external to the interface has been updated

### Tests
We compiled and built state conversion programs aether\_to\_dart and dart\_to\_aether
and tested them on the latest Aether restart file sets.
Model\_mod\_check passes tests 1-5 and 7, except for expected failure of interpolation at the poles.
Filter is able to assimilate several observations of AIRS temperature 
and GPS profiles of electron density (even though that's not available in the restart files).

## Checklist for merging

- [ ] Confirm that we will not develop cycling scripting at this time (or develop it).
- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed


models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,191 @@
+Aether Rectangular Grid Interface
+=================================
+
+Overview
+--------
+
+The `Aether`_ ("eether") space weather model can be implemented 
+on a logically rectangular grid "lat-lon", 
+or on an the cubed-sphere grid (see ../aether_cubed_shere).
```
hkershaw-brown, line 9: There is no aether_cubed_sphere at the moment. 
```suggestion
or on an the cubed-sphere grid.
```

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,191 @@
...
+This is the interface to the lat-lon version.
+
+.. _Aether: https://aetherdocumentation.readthedocs.io/en/latest/
```
hkershaw-brown, line 12:
```suggestion
Aether is available on GitHub https://github.com/AetherModel/Aether
```

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,191 @@
...
+
+Aether writes history and restart files, with some overlap of the fields (?).
```
hkershaw-brown, line 14:
```suggestion
Aether writes history and restart files, with some overlap of the fields.
```

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,191 @@
...
+There's a unique set of these files for each member.
+The restart file names reflect this information:  
+
+|   {neutrals,ions}_mMMMM_gBBBB.nc
+|   MMMM = ensemble member (0-based)
+|   BBBB = block number (0-based)
+
+The restart files do not have grid information in them, which must be read from
+   grid_gBBBB.nc
+
```
hkershaw-brown, line 30:
```suggestion
The restart files do not have grid information in them. Grid information must be read from
   grid_gBBBB.nc

```

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,191 @@
...
+Aether_to_dart and dart_to_aether read the same namelist; transform_state_nml.
+The fields chosen to be part of the model state are specified in ``variables``.
+Program aether_to_dart will read the specified fields, from all the restarts
```
hkershaw-brown, line 33: link to aether_to_dart and dart_to_aether docs in this paragraph

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,191 @@
...
+      in the ``variables`` namelist variable in transform_state_nml.
+      The following information must be provided for each field
+      
+         1) Aether field name
+         2) which file contains the field ("neutrals" or "ions")
+      
+      Aether field names are not CF-compliant and are translated 
+      to CF-compliant forms by aether_to_dart.
+      The suggested DART quantity to associate with some fields are listed
+      in ./aether_to_dart.nml.
```
hkershaw-brown, line 77: DART quality is not part of aether_to_dart or dart_to_aether, and the file aether_to_dart.nml does not exist.  
```suggestion
```

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,191 @@
...
+> cd {aether_restart_dirname}
+> mkdir Orig
+> cp *m0000* Orig/
+> ./aether_to_dart  0
+> cp filter_input_0001.nc filter_output_0001.nc
+> ./dart_to_aether  0
+
+| The filter\_ files will contain the CF-compliant field names which must be used in model_nml:variables.
+| Compare the modified Aether restart files with those in Orig.
+.. NOTE:
```
hkershaw-brown, line 185: The note is not showing up in the documentation:  https://dart-documentation--635.org.readthedocs.build/en/635/models/aether_lat-lon/readme.html#usage
```suggestion

.. NOTE::
```

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,191 @@
...
+   Some halo parts may have no data in them because Aether currently (2024-2) 
+   does not use those regions.
+.. WARNING:
```
hkershaw-brown, line 188: The warning is not showing up in the documentation https://dart-documentation--635.org.readthedocs.build/en/635/models/aether_lat-lon/readme.html#usage
```suggestion

.. WARNING::
```

models/aether_lat-lon/matlab/new_varname_file.m
```diff
@@ -0,0 +1,155 @@
+function new_varname_file(data_dir, member, nblocks)
+
+% Converts Aether restart file names to updated (2024-1-17) versions.
```
hkershaw-brown, line 3: There is no mention of this matlab file 'new_varname_file.m' in the documentation.  Is this something that you would expect people to use?  If not, then I'd remove it from this pull request.

models/aether_lat-lon/model_mod.nml
```diff
@@ -0,0 +1,9 @@
+&model_nml
+   template_file  = 'if other than filter_input_0001.nc'
+   debug          = 0
+   variables = 'Temperature',   'QTY_TEMPERATURE',      '1000.0',  'NA',      'UPDATE',
```
hkershaw-brown, line 4: I think these should be set to sensible defaults
Do you want to clamp temperature at a minimum of 1000? If not have 'NA' for clamping. 

similarly, set:
template_file='filter_input_0001.nc'

models/aether_lat-lon/work/quickbuild.sh
```diff
@@ -0,0 +1,52 @@
...
+create_fixed_network_seq
+create_obs_sequence
+obs_diag
+obs_seq_to_netcdf
+)
+
+model_serial_programs=(
+aether_to_dart
+dart_to_aether)
+# transform_names
```
hkershaw-brown, line 31: remove program that no-longer exists
```suggestion
```

models/aether_lat-lon/work/input.nml
```diff
@@ -0,0 +1,326 @@
...
+   interp_test_yrange    = -888888.0, -888888.0
+   interp_test_zrange    = -888888.0, -888888.0
+   interp_test_vertcoord = 'VERTISHEIGHT'
+   test1thru             = -1
+   run_tests             = 1,2,3,4,5,7
+   verbose               = .FALSE.
+  /
+
+&quad_interpolate_nml
+    debug = 999
```
hkershaw-brown, line 325: Turn off debugging of quat_interpolate_mod
```suggestion
```

hkershaw-brown (changes requested):
Here's the first batch of review comments on the documentation.
A couple of your NOTES and WARNINGS are not showing up in the html documentation, so definitely change those.

hkershaw-brown (changes requested):
next batch, dart_to_aether, aether_to_dart, transform_state_mod

kdraeder:
> Here's the first batch of review comments on the documentation. A couple of your NOTES and WARNINGS are not showing up in the html documentation, so definitely change those.

Thanks for catching so many mistakes!
I'll work on the readme.rst and the state transform files.
If I finish all that and @johnsonbk is busy with other things, I'll work on the model_mod.

models/aether_lat-lon/aether_to_dart.f90
```diff
@@ -0,0 +1,465 @@
...
+! state_time is read in transform_state_mod and is available by the use statement.
+call write_model_time(ncid, state_time)
+
+! Define (non-time) variables
+call restarts_to_filter(aether_restart_dirname, ncid, member, define=.true.)
+
+! Read and convert (non-time) variables
+call restarts_to_filter(aether_restart_dirname, ncid, member, define=.false.)
+! subr. called by this routine closes the file only if define = .true.
+call nc_close_file(ncid)
```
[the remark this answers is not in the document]
kdraeder: That was my feeling too, but I went with the template that we settled on at the start.

models/aether_lat-lon/readme.rst
kdraeder: What's the advantage of referring to the model on github, which is more focused on code,
rather than the model documentation?

models/aether_lat-lon/readme.rst
kdraeder: @johnsonbk and @kdraeder  settled on a single readme for this model.
Should the aether_to_dart and dart_to_aether docs be split out, and point to those new rst files,
or should this point to the source code for the programs, or ...?

models/aether_lat-lon/matlab/new_varname_file.m
kdraeder: I don't expect people to use it as a regular part of assimilating.
But if Aether variable names change again in the future, and still use difficult characters 
(blanks, \s, etc.) this file would be very useful for transforming old files into new ones.
I was not able to make NCO or fortran do it.  Python might be able to, but that hasn't been written
(? @johnsonbk).
Given that, I'll either write documentation or remove it from the PR.  Which do you prefer?

models/aether_lat-lon/readme.rst
hkershaw-brown: I don't think the readthedocs site is aether right? It talks about installing CAM, CESM and WACMX.
Seems safest to point people at the Aether github repo.

models/aether_lat-lon/matlab/new_varname_file.m
hkershaw-brown: I'd say a line or two in the documentation saying new_varname_file as a tool

models/aether_lat-lon/readme.rst
hkershaw-brown: my bad, no links needed.

models/aether_lat-lon/readme.rst
kdraeder: The original link points to the "Aether Model Documentation" page.
The next page is the Aether github instructions.  After that are system requirements for Aether.
Then it gets into the CAM and WACCM pieces which were useful for building Aether.
Whichever you think is best.

hkershaw-brown: lets point people direct to Aether.

models/aether_lat-lon/aether_to_dart.f90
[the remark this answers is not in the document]
hkershaw-brown: fair enough, let's leave this as-is.

models/aether_lat-lon/model_mod.f90
```diff
@@ -0,0 +1,616 @@
...
+
+integer(i8),         intent(in)  :: index_in
+type(location_type), intent(out) :: location
+integer, optional ,  intent(out) :: qty
+
+! Local variables
+
+integer :: lat_index, lon_index, lev_index
+integer :: my_var_id, my_qty
+! character(len=*), parameter :: routine = 'get_state_meta_data'
```
hkershaw-brown, line 328:
```suggestion
```

models/aether_lat-lon/model_mod.f90
```diff
@@ -0,0 +1,616 @@
...
+
+! longitude
+nlon = nc_get_dimension_size(ncid, trim(LON_DIM_NAME), routine)
+allocate(lons(nlon))
+call nc_get_variable(ncid, trim(LON_VAR_NAME), lons, routine)
+
+end subroutine assign_dimensions
+
+!-----------------------------------------------------------------------
+! Parse the table of variables' characteristics into arrays for easier access.
```
hkershaw-brown, line 413: Just to make life easier for syntax highlighting remove ' from comment 
```suggestion
! Parse the table of variables characteristics into arrays for easier access.
```

models/aether_lat-lon/model_mod.f90
```diff
@@ -0,0 +1,616 @@
...
+
+function assign_var(variables, MAX_STATE_VARIABLES) result(var)
```
hkershaw-brown, line 415: nitpick: the indenting for function assign_var does not match the rest of the file.

models/aether_lat-lon/model_mod.f90
```diff
@@ -0,0 +1,616 @@
...
+   integer, intent(in)                         :: MAX_STATE_VARIABLES
+
+   type(var_type) :: var
+
+   integer        :: ivar
+
+   !-----------------------------------------------------------------------
+   ! Codes for interpreting the NUM_STATE_TABLE_COLUMNS of the variables table
+   integer, parameter :: NAME_INDEX      = 1 ! ... variable name
+   integer, parameter :: QTY_INDEX       = 2 ! ... DART kind
```
hkershaw-brown, line 427:
```suggestion
   integer, parameter :: QTY_INDEX       = 2 ! ... DART qty
```

models/aether_lat-lon/model_mod.f90
```diff
@@ -0,0 +1,616 @@
...
+   enddo
+  
+   ! Allocate the arrays in the var derived type
+   allocate(var%names(var%count), var%qtys(var%count), var%clamp_values(var%count, 2), var%updates(var%count))
+  
+   do ivar = 1, var%count
+   
+      var%names(ivar) = trim(variables(NAME_INDEX, ivar))
+  
+      var%qtys(ivar) = get_index_for_quantity(variables(QTY_INDEX, ivar))
```
hkershaw-brown, line 448: DART QTYs are case sensitive,  so the following will error:
```
model_nml
   variables      = 'Temperature',    'qty_TEMPERATURE',      'NA',  'NA',   'UPDATE',
```


to_upper will convert your character TO_UPPER

models/aether_lat-lon/model_mod.f90
```diff
@@ -0,0 +1,616 @@
...
+          var%clamp_values(ivar,1) = MISSING_R8
+      endif
+  
+      if (variables(MAX_VAL_INDEX, ivar) /= 'NA') then
+          read(variables(MAX_VAL_INDEX, ivar), '(d16.8)') var%clamp_values(ivar,2)
+      else
+          var%clamp_values(ivar,2) = MISSING_R8
+      endif
+  
+      if (variables(UPDATE_INDEX, ivar) == 'UPDATE') then
```
hkershaw-brown, line 462: case sensitve, so 'update' in the input.nml would give you var%updates(ivar) = .false.

hkershaw-brown (changes requested):
last batch, model_mod

For reading in which variables in the state, the case matters in the current code. 
I'd definitely use to_upper() for the UPDATE since it is likely someone will write 'update'.
Unless you are adding qtys with a mixture of upper and lower case, I'd to_upper(QTY) as well.

models/aether_lat-lon/model_mod.f90
johnsonbk: Added subroutine call to_upper(table_entry) before the var%qtys element is assigned.

models/aether_lat-lon/model_mod.f90
johnsonbk: Added subroutine call to_upper(table_entry) before the var%updates element is assigned.

johnsonbk:
Thanks for everyone's work on this pull request. I ran into two problems following the directions in `readme.rst`.

### First problem: `model_mod_check` test number 2 fails

Starting with a fresh clone of DART

```
cd <abs path to local DART installation>
git clone https://github.com/NCAR/DART.git DART
cd DART/
git fetch origin
git switch aether
cd build_templates
mv <suitable mkmf.template> mkmf.template
cd ../models/aether_lat-lon/work
sftp <user>@data-access.ucar.edu
get /glade/work/raeder/Exp/Aether/Ens_renamed_derecho.tgz
exit
tar -zxvf Ens_renamed_derecho.tgz
cp Ens_renamed_derecho/filter_input_0001.nc ./
./quickbuild.sh
# The executables compile
./model_mod_check
...
***************** RUNNING    TEST 2    ***********************
 -- Read and write restart file
**************************************************************
--------------------------------------------------------------
 Reading File : filter_input_0001.nc
--------------------------------------------------------------
 ERROR FROM:
  source : dart_time_io_mod.f90
  routine: read_model_time:
  message:  inconsistent calendar types between DART program and input file.
  message: ...  DART initialized with: NO_CALENDAR File uses: GREGORIAN
  message: ...  You may need to supply a model-specific "read_model_time()" to read the time.
```

Perhaps this failure is related to commit 3c7d8d?

### Second problem: `aether_to_dart` doesn't use the `aether_restart_dirname` namelist entry in `open_block_file`

While trying to run `aether_to_dart`:

```
pwd 
<abs path to local DART installation>/DART/models/aether_lat-lon/work
vim input.nml
# Edit input.nml to set
aether_restart_dirname = 
      '<abs path to local DART installation>/DART/models/aether_lat-lon/work/Ens_renamed_derecho/'

# Continuing with the directions in readme.rst:
cd <aether_restart_dirname>
mkdir Orig
cp *m0000* Orig/
```
The following command to change directory back to the work directory is not in the directions in `readme.rst` but that's where `aether_to_dart` is compiled.
```
cd <abs path to local DART installation>/DART/models/aether_lat-lon/work
./aether_to_dart  0
...
ERROR FROM:
  source : aether_lat-lon/transform_state_mod.f90
  routine: open_block_file
  message: cannot open file grid_g0000.nc 
                                                                                                                   for read
```
If `aether_to_dart` and `input.nml` are in the aether work directory and the aether netcdf files are in a different directory, `<aether_restart_dirname>` this error is thrown.

## Attempt number 1 to fix this error involves copying the grid files to the work directory
```
cp <aether_restart_dirname>/grid_g000?.nc ./
./aether_to_dart 0
...
ERROR FROM:
  source : aether_lat-lon/transform_state_mod.f90
  routine: open_block_file
  message: cannot open file neutrals_m0000_g0000.nc                                                                                                                                                                                                                                          for read
```

## Attempt number 2 to fix this error involves copying the `aether_to_dart` executable and `input.nml` to `<aether_restart_dirname>`.
```
cp aether_to_dart <aether_restart_dirname>
cp input.nml <aether_restart_dirname>
cd <aether_restart_dirname>
./aether_to_dart
...
aether_to_dart Successfully converted the Aether restart files to 'filter_input_0001.nc'
 
 --------------------------------------
 Finished ... at YYYY MM DD HH MM SS = 
                 2024  2 14 11 23 29
 --------------------------------------
```

The cause of this unexpected behavior seems to be that the function open_block_file is not passed a filename including a the path of `<aether_restart_dirname>`. The executable only works if it is copied to `<aether_restart_dirname>` or if `<aether_restart_dirname>` is the same directory as the aether work directory. The reason this behavior is unexpected is that `aether_restart_dirname` is set in `transform_state_nml` so a user would expect that `aether_to_dart` would actually use that namelist entry as the directory in which the restart files are stored.

hkershaw-brown:
> ...
> ***************** RUNNING    TEST 2    ***********************
>  -- Read and write restart file
> **************************************************************
> --------------------------------------------------------------
>  Reading File : filter_input_0001.nc
> --------------------------------------------------------------
>  ERROR FROM:
>   source : dart_time_io_mod.f90
>   routine: read_model_time:
>   message:  inconsistent calendar types between DART program and input file.
>   message: ...  DART initialized with: NO_CALENDAR File uses: GREGORIAN
>   message: ...  You may need to supply a model-specific "read_model_time()" to read the time.
> ```
> 
> Perhaps this failure is related to commit 3c7d8d?

This is my bad,  I meant force the calendar to be Gregorian rather than having a variable calder (implying a user choice).  I'll go ahead and fix this.

kdraeder:
@johnsonbk  Thanks for working through the file transformation instructions.
I'll work on that, unless you already started fixing the issues.

> The cause of this unexpected behavior seems to be that the function open_block_file is not passed a filename including the path of `<aether_restart_dirname>`. The executable only works if it is copied to `<aether_restart_dirname>` or if `<aether_restart_dirname>` is the same directory as the aether work directory. The reason this behavior is unexpected is that `aether_restart_dirname` is set in `transform_state_nml` so a user would expect that `aether_to_dart` would actually use that namelist entry as the directory in which the restart files are stored.

johnsonbk:
I haven't begun to start fixing the transform_state bug, @kdraeder. Please proceed.

kdraeder:
> Second problem: `aether_to_dart` doesn't use the `aether_restart_dirname` namelist entry in `open_block_file`

This brings up a strategy question, which I thought about at the beginning, but lost track of.
We've decided not to develop full cycling scripting at this time,
so we don't have that context to define the use of aether_to_dart, filter, and dart_to_aether.
The experiment structure I'm most familiar with has the following:

1. All of the DART executables are copied to an exec_dir.
2. All of the ensemble model output is written to a run_dir and has a full date in the filenames.  (The Aether restart file names do not have the date in them, so the scripting will need to prevent overwriting of files which need to be kept.)
3. The assimilation script runs the executables in the run_dir using full pathnames of the executables.
4. All of the DART output is written to the run_dir.
5. The updated state variables are written into the existing model restart files.

An advantage of this is that the run_dir does not need to be specified in the DART executables 
because the scripting changes to that directory before running the executables.
They only need to know file names.  So far I don't see any disadvantages.

Is this the structure we want here?
Or do we want the executables to be able to find the restart files wherever they are?
And write their output to an output location, which may be different from the input files location?

nancycollins:
the wrf folks seem to prefer putting each ensemble member's input and output in a separate subdir, probably because that makes the model advances easier to run in parallel by letting each member run in their own subdirectory without interfering with each other.

as long as it doesn't complicate the code too much, enabling flexible options for how someone wants to run both the assimilation and the model advances would be good.  i know in cesm that the model advances can run in parallel using cesm options, but that may not be true for aether.  

also, you probably don't want to assume that the scripting is always going to do a model advance before an assimilation, like cesm does.  it's very useful for exploring parameter/namelist settings or debugging to run filter alone from a simple batch script without any other scripting requirements, as long as all the files are in the correct place with the correct names.

having said all that, it sounds like getting something running is the highest priority, so whatever works quickly might be the strategic thing here.

kdraeder:
I opted to make aether_to_dart and dart_to_aether use the same directory for the Aether and filter restart files,
but allow those programs to executed from any directory.
I tested and pushed those commits.

index.rst
```diff
@@ -439,6 +439,7 @@ References
    :hidden:
 
    models/9var/readme
+   models/aether_lon-lat/readme
```
hkershaw-brown, line 442:
```suggestion
   models/aether_lat-lon/readme
```

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,242 @@
...
+Overview
+--------
+
+The Aether ("eether") space weather model can be implemented 
+on a logically rectangular grid "lat-lon", or on an the cubed-sphere grid.
+This is the interface to the lat-lon version.
+The model code is available on 
+`GitHub <https://github.com/AetherModel/Aether>`_ .
+Additional documentation can be found 
+`here <https://aetherdocumentation.readthedocs.io/en/latest/>`_ .
```
hkershaw-brown, line 13:
```suggestion
```

models/aether_lat-lon/readme.rst
```diff
@@ -0,0 +1,242 @@
...
+which is just the transformation of files for a member (0) 
+use the following steps.  
+These build the ``aether_to_dart`` and ``dart_to_aether`` executables
+in $DART/models/aether_lat-lon/work directory.
+Also in that directory, edit input.nml to set ``transform_state_nml:`` ``aether_restart_dirname``
+to be the full pathname of the directory where the Aether restart and grid files are.
+
+::
+
+> set exec_dir = $DART/models/aether_lat-lon/work
```
hkershaw-brown, line 221: say this c shell

models/aether_lat-lon/aether_to_dart.f90
```diff
@@ -0,0 +1,472 @@
...
+call initialize_utilities(progname)
+
+!----------------------------------------------------------------------
+! Get the ensemble member
+!----------------------------------------------------------------------
+num_args = command_argument_count()
+if (num_args == 0) then
+   write(error_string_1,*) 'Usage: ./aether_to_dart member_number (0-based)' 
+   call error_handler(E_ERR, progname, error_string_1)
+endif
```
mgharamti, line 86: I guess whoever runs this would need to write a loop to generate all filter restarts one by one. Any reason why not generate all members internally? Is there a scenario where a user doesn't need to create all members?

models/aether_lat-lon/aether_to_dart.f90
```diff
@@ -0,0 +1,472 @@
...
+                1-nghost:nx_per_block+nghost))
+
+! TODO: We need all altitudes, but there might be vertical blocks in the future.
+!       But there would be no vertical halos.
+!       Make transform_state_mod: zcount adapt to whether there are blocks.
+!       Temp needs to have C-ordering, which is what the restart files have.
+! temp array large enough to hold 1 species, temperature, etc
+allocate(temp3d(1:nz_per_block, &
+                1-nghost:ny_per_block+nghost, &
+                1-nghost:nx_per_block+nghost))
```
mgharamti, line 224: I tried to ncview individual members are running aether_to_dart but the fields don't show up as I expect. I was hoping to see a global field and then change through the levels. However, I could only change through the lon dimension which is not very useful. This is because the dimensions of the 3D variables are ordered as lon, lat, alt. I ran a simple 'ncpdq -a alt,lon filter_input_0001.nc filter_input_0001_swap.nc' on the file to permute the dimensions and then I was able to see the entire globe at different levels. I suggest modifying aether_to_dart to adjust the dimension order.

models/aether_lat-lon/aether_to_dart.f90
```diff
@@ -0,0 +1,472 @@
+! DART software - Copyright UCAR. This open source software is provided
+! by UCAR, "as is", without charge, subject to all terms of use at
+! http://www.image.ucar.edu/DAReS/DART/DART_download
+!
+
+program aether_to_dart
```
mgharamti, line 6: If I run aether_to_dart with an argument that is larger than the ensemble size, it fails. Yet, it creates a filter_input netcdf file including only dimension variables. I think it should fail before creating any file

models/aether_lat-lon/work/input.nml
```diff
@@ -0,0 +1,325 @@
...
+   nblocks_lon = 2
+   nblocks_lat = 2
+   nblocks_lev = 1
+   debug = 0
+   /
+
+&model_nml
+   template_file  = 'filter_input_0001.nc'
+   variables      = 'Temperature',    'QTY_TEMPERATURE',      'NA',  'NA',   'UPDATE',
+                    'Opos',           'QTY_DENSITY_ION_OP',   'NA',  'NA',   'UPDATE'
```
mgharamti, line 135: Do we need any clamping? maybe for O+?

models/aether_lat-lon/work/input.nml
```diff
@@ -0,0 +1,325 @@
...
+   allow_missing_clm            = .false.
+   /
+
+
+
+&ensemble_manager_nml
+   /
+
+&assim_tools_nml
+   cutoff                          = 0.2
```
mgharamti, line 110: How about the cutoff, is 0.2 the default that we want users to use?

models/aether_lat-lon/work/input.nml
```diff
@@ -0,0 +1,325 @@
...
+&quality_control_nml
+/
+
+&state_vector_io_nml
+   /
+
+&perfect_model_obs_nml
+   read_input_state_from_file = .true.
+   single_file_in             = .false.
+   input_state_files          = 'wrfinput_d01'
```
mgharamti, line 17: in pmo: the input_state_files is set to 'wrfinput_d01'. We probably need to change that

models/aether_lat-lon/work/input.nml
```diff
@@ -0,0 +1,325 @@
+&probit_transform_nml
+   /
+
+&algorithm_info_nml
+   qceff_table_filename = ''
+   /
+
+&quality_control_nml
```
mgharamti, line 8: add outlier_threshold = 3

models/aether_lat-lon/work/input.nml
```diff
@@ -0,0 +1,325 @@
...
+   inf_initial_from_restart    = .false.,                 .false.,
+   inf_sd_initial_from_restart = .false.,                 .false.,
+   inf_deterministic           = .true.,                  .true.,
+   inf_initial                 = 1.0,                     1.0,
+   inf_lower_bound             = 0.0,                     1.0,
+   inf_upper_bound             = 1000000.0,               1000000.0,
+   inf_damping                 = 1.0,                     1.0,
+   inf_sd_initial              = 0.0,                     0.0,
+   inf_sd_lower_bound          = 0.0,                     0.0
+   inf_sd_max_change           = 1.05,                    1.05,
```
mgharamti, line 93: set inf_flavor to 5. Also, change inf_sd_initial and inf_sd_lower_bound to 0.6

models/aether_lat-lon/aether_to_dart.f90
kdraeder: We needed working code as soon as possible and a single member program was simpler to write.
It's possible to run multiple aether_to_darts in parallel, so it shouldn't be much slower than
having aether_to_dart handle all members at once.
There are testing and development scenarios which only need 1 member, and perfect_model_obs.

models/aether_lat-lon/aether_to_dart.f90
kdraeder: The coordinate order is annoying.  Aether won't change, and we decided to use it in order to make the conversion programs simpler to code and run faster (no transposing).  
Ncview can show a lat-lon slice by changing the axes.  But the lat and lon are transposed, so it's not perfect.  Other common viewing tools can transpose it.

models/aether_lat-lon/aether_to_dart.f90
kdraeder: Good catch!  I'll try to fix that.

models/aether_lat-lon/work/input.nml
kdraeder: We could certainly make 0.0 be the lower limit for the ion densities and temperatures, since we expect that inflation will sometimes make those be negative.  Or leave that as an exercize for the user.  
Whether it's needed or not depends on whether  Aether and the assimilation setup work well together.  One advantage of not clamping (especially the max values) is that it helps users see when an assimilation is failing, instead of silently correcting itself.

models/aether_lat-lon/work/input.nml
kdraeder: My guess is that this is a science question the users will need to answer.
0.2 has been reasonable in other global models, but if @johnsonbk has any insight, we can set a different default value.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
```diff
@@ -63,7 +63,7 @@ module netcdf_utilities_mod
           nc_begin_define_mode,           &
           nc_end_define_mode,             &
           nc_synchronize_file,            &
-          NF90_MAX_NAME, NF90_MAX_VAR_DIMS
+          NF90_MAX_NAME, NF90_MAX_VAR_DIMS, NF90_FILL_REAL
```
nancycollins, line 66: i'm not sure when the NF90 constants got added to this fie, but the intent was never to replicate the actual netcdf utilities mod.  adding more constants to this is just going down the wrong path.  these functions were intended to be more user-friendly ways to access netcdf without having to get into the weeds of the netcdf interface.  if you need these constants in your code, 'use netcdf' or make a cover function that has the name of what you're doing and uses this constant inside the function.  e.g. from the name i'm guessing it fills an array.  make a 
nc_fill_array_blank(array) that doesn't require netcdf constants in the calling code.  

i know this isn't an important fix, but it's a long-term bad idea.  maybe after the crush of getting this working is done, someone can go back and make this better?

models/aether_lat-lon/work/input.nml
kdraeder: That will need to be changed, but pmo is not in the scope of this stage of development,
so we haven't looked at that.  I'll replace wrfinput_d01 with "pmo not ready for use".

nancycollins (commented):
i just left a comment on the change to the dart version of the netcdf utilities cover functions.  it's not important for getting aether running, but sometime in the future someone should go back and remove all the NF90 constants and make better named functions that either return those values or do the code that uses the constants.

models/aether_lat-lon/work/input.nml
kdraeder: I added that, and 'input_qc_threshold   = 3.0'.  This input.nml hasn't been used directly for assimilation, so it may be missing other variables, or have suboptimal values.
@johnsonbk may also have an input.nml that's been updated to do an assimilation.

models/aether_lat-lon/work/input.nml
kdraeder: Done.

assimilation_code/modules/utilities/netcdf_utilities_mod.f90
kdraeder: Yes, I didn't do much work to figure out the existing strategy for using netcdf parameters.
The template I was working from used some, so I assumed it was not terrible to do that.

nancycollins: which is the only reason i brought it up.  once the camel's nose is in the tent...   
it's *not* important for this pull request and a later change to your code will be easy.  i will open an issue on this because i'm not requesting a change to this pull request, so this is the wrong place to address this.

models/aether_lat-lon/aether_to_dart.f90
kdraeder: The fix is to read the model time from the actual member instead of member 0.  
It will fail before the filter_input.nc file is created.

mgharamti:
@kdraeder I tested your changes. I am still getting a restart file even though the argument is larger than the ensemble size. I was expecting it to fail before creating a new netcdf file. And btw, we need the same logic applied to dart_to_aether. 
This, however, is not a deal breaker and I'm OK to move forward with the PR in its current form.

mgharamti:
Great, thanks Kevin. This works in both directions now!

PR #686, opened 2024-06-06
On demand GitHub action for build everything

ann-norcio:
## Description:

Added an on demand GitHub action for running all ./quickbuild.sh in DART with the gfortran compiler. 

### Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

### Tests
Ran the existing build_everything developer test and compared/verified outputs.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed


.github/workflows/action_on_demand.yml
```diff
@@ -0,0 +1,72 @@
...
+          find $DART -name input.nml -exec sed -i -e "/^[[:space:]]*#/! s|.*output_obs_def_mod_file.*|output_obs_def_mod_file = './obs_def_mod.f90'|g" \
+            -e "/^[[:space:]]*#/! s|.*output_obs_qty_mod_file.*|output_obs_qty_mod_file = './obs_kind_mod.f90'|g" \
+            -e "/^[[:space:]]*#/! s|.*output_obs_kind_mod_file.*|output_obs_qty_mod_file = './obs_kind_mod.f90'|g" {} \;
+
+      - name: Run quickbuild.sh scripts
+      # Run quickbuild.sh scripts and log failures
+        run: |
+          find $DART -executable -type f -name quickbuild.sh | while read filepaths; do
+            (cd "$(dirname "$filepaths")" && ./quickbuild.sh) || echo "FAILED: $filepaths" >> $GITHUB_WORKSPACE/quickbuild-failures.txt
+          done
```
hkershaw-brown, line 49: This is serial at the moment, running one quickbuild.sh after another

.github/workflows/action_on_demand.yml
```diff
@@ -0,0 +1,72 @@
...
+        # "./developer_tests/forward_operators/work/quickbuild.sh"
+        # "./observations/obs_converters/GOES/work/quickbuild.sh"
+        # "./observations/obs_converters/GMI/work/quickbuild.sh"
+        # "./observations/obs_converters/NSIDC/work/quickbuild.sh"
+        # "./observations/obs_converters/gps/work/quickbuild.sh"
+        # "./observations/obs_converters/var/work/quickbuild.sh"
+        # "./observations/obs_converters/AIRS/work/quickbuild.sh"
+        # "./observations/obs_converters/quikscat/work/quickbuild.sh"
+        # "./observations/obs_converters/GSI2DART/work/quickbuild.sh"
+        # "./models/template/work/quickbuild.sh"
```
hkershaw-brown, line 72: This looks like it _always_ gives "FAIL" for the action, and a person has to manually check the builds against the known failures. 
The action should:
 - pass if everything that is expected to build builds successfully
 - fail if something that should build doesn't

.github/workflows/action_on_demand.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: |
+          if [ -f $GITHUB_WORKSPACE/quickbuild-failures.txt ]; then
+            echo "Failed builds:"
+            cat $GITHUB_WORKSPACE/quickbuild-failures.txt
+            exit 1
+          else
+            echo "All runs passed"
+          fi
+      # The following should be expected to fail:
+        # "./developer_tests/quad_interpolate/work/quickbuild.sh"
```
hkershaw-brown, line 62: "./developer_tests/quad_interpolate/work/quickbuild.sh" should (and does) pass

hkershaw-brown (changes requested):
Hi Ann, 

This approach is following the Derecho build_everything developer tests, so while it is building everything, it is taking a long time (~35 minutes) and it is hard to dig into the into the individual failures.  

I think it is better to use a matrix since we are using GitHub actions, so each quickbuild.sh is its own runner. 
https://github.com/hkershaw-brown/DART/actions/runs/9484699298

If you follow the Derecho version and run the quickbuild.sh in parallel on a single runner, I think you're really going to be limited in speedup (4 cpus vs 128 on Derecho).

hkershaw-brown (commented):
Ann good work on this, however I'm going to switch this out for the matrix method since this the matrix method is coming out ~10 times faster.  

I don't think doing on more work on this pull request is worth your time since no matter how well you parallelize with the script using the methods for the Derecho version, when you only have a single runner it won't go that much faster because you only have [4 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) to play with vs. the 128 available on Derecho.

PR #593, opened 2023-12-11
Quickbuild tests

hkershaw-brown:
## Description:

building on Ann's previous pull request https://github.com/NCAR/DART/pull/575
improved build_everything, submits a job on Derecho for each compiler, each job runs every quickbuild.sh

time:
ccc ~12minutes
gfortran ~2:30 minutes
nvhpc ~4:40 minutes
ifort ~4:40 minutes

Note several of the converters require external libraries or code to be added before compiling, ignoring these for this pull request (rttov, hdfeos, wrf code, ncep prepbuf code):

```
intel RESULT: 1 /glade/derecho/scratch/hkershaw/build_everything/intel/DART/developer_tests/forward_operators/work/ FAILED
intel RESULT: 10 /glade/derecho/scratch/hkershaw/build_everything/intel/DART/observations/obs_converters/GOES/work/ FAILED
intel RESULT: 11 /glade/derecho/scratch/hkershaw/build_everything/intel/DART/observations/obs_converters/gps/work/ FAILED
intel RESULT: 12 /glade/derecho/scratch/hkershaw/build_everything/intel/DART/observations/obs_converters/NSIDC/work/ FAILED
intel RESULT: 28 /glade/derecho/scratch/hkershaw/build_everything/intel/DART/observations/obs_converters/var/work/ FAILED
intel RESULT: 29 /glade/derecho/scratch/hkershaw/build_everything/intel/DART/observations/obs_converters/AIRS/work/ FAILED
intel RESULT: 31 /glade/derecho/scratch/hkershaw/build_everything/intel/DART/observations/obs_converters/quikscat/work/ FAILED
intel RESULT: 42 /glade/derecho/scratch/hkershaw/build_everything/intel/DART/observations/obs_converters/GMI/work/ FAILED
```
Other failures: #592 #352, #594

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

### Documentation changes needed?

- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.

### Tests
Derecho: build everything (that has a quickbuild.sh) in DART for cce, intel, gcc, nvhpc

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
[...]

PR #691, opened 2024-06-21
on-demand GitHub action which runs every quickbuild.sh in DART 

hkershaw-brown:
## Description:

On demand github action to run every quickbuild.sh in DART
Uses container with dart dependencies hdfeos2, hdf4, hdf5, rttov12
Takes ~4 minutes

### Fixes issue

Github version the build_everything developer tests https://github.com/NCAR/DART/pull/593

### Types of changes

- [ ] Documentation update

### Tests
Run the action on main branch of your fork. 

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed


mjs2369 (commented):
@hkershaw-brown I've reviewed and tested your code by using your new action on my fork. The action reported three failures from the quickbuilds of the following three obs converters:

- Gps
```
/usr/bin/ld: cannot find ../../NCEP/prep_bufr/lib/bufrlib.a: No such file or directory
collect2: error: ld returned 1 exit status
make: *** [Makefile:227: convert_cosmic_gps_cdf] Error 1
```

- GSI2DART 
```
  173 |     call mpi_bcast(sprd_ob,nobs_tot,mpi_real4,0,mpi_comm_world,ierr)
      |                   1
Error: Type mismatch between actual argument at (1) and actual argument at (2) (REAL(4)/INTEGER(4)).
```

- Var 
`Fatal Error: Cannot open module file 'da_constants.mod' for reading at (1): No such file or directory`

While some failures are to be expected, it causes the entire action to report back as a failure.

image
 
Is this the behavior we want for the actions, or is there a way to set which jobs we expect to fail and compare against the results so that the whole check passes?

hkershaw-brown:
I'm kind of digging the failure as a mark of shame, but you're right you have to check the annotations. Do you think skipping the expected fails is ok (not run them in the matrix)? 

Screenshot 2024-07-03 at 11 57 23 AM

mjs2369:
> I'm kind of digging the failure as a mark of shame, but you're right you have to check the annotations. Do you think skipping the expected fails is ok (not run them in the matrix)? 

I think that sounds like a good plan. I'm just a little confused on how we would go about setting the expected fails to have them not run in the matrix? Just like how you are currently skipping the template model, but for the three failing obs converters instead?

```
for item in "${files_to_process[@]}"; do
            if [[ $item == *"template"* ]]; then
              continue  # template model not for compiling
            fi
```

Also side note is that GSI2DART compiles with the use of -fallow-argument-mismatch

hkershaw-brown:
> @hkershaw-brown I think that sounds like a good plan. I'm just a little confused on how we would go about setting the expected fails to have them not run in the matrix? Just like how you are currently skipping the template model, but for the three failing obs converters instead?
> 
> ```
> for item in "${files_to_process[@]}"; do
>             if [[ $item == *"template"* ]]; then
>               continue  # template model not for compiling
>             fi
> ```

yup, just skip them.

hkershaw-brown:
skipping known failures, using  -fallow-argument-mismatch for GSI2DART 21399b953b12c444214436bc37a617ee5b2d809e

mjs2369 (approved):
@hkershaw-brown I've reviewed your updates and everything looks good. I went ahead and ran the updated workflow on my fork and it passed now that we have skipped the known failures. Approved

PR #701, opened 2024-07-11
MARBL-DART

### LINKED ISSUE #614: bug: assuming all state variables have unlimited dimension
Making a separate issue for this since Robin has hit this while working on MOM6-MARBLE.

Assumptions in direct_netcdf_mod.f90:
* If there is an unlimited dimension, every variable has the unlimited dimension. 
* The unlimited dimension is time, and we are interested in the latest time slice in a file. 

In the cesm 2.3 clm, there is an unlimited dimension cohorts which is not used by any of the state variables (and, kind of strangely, not used at all)

So this line: 

https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/assimilation_code/modules/io/direct_netcdf_mod.f90#L865
num_dims - 1 = 0

Gustavo's compiler is picking up this 1:0 and erroring out. 

intel 19.1.1.21 (running on Cheyenne) just merrily continues, but the counts are incorrect.  The count is set to 1 rather than the length of the dimension. 

For fill_inflation_restart it does not matter if the state is read in incorrectly, since we just fill the output with (mean and sd (e.g. 1.0, 0.6).  Side note here is we don't need to be calling read_state in fill_inflation_restart. 

For filter, which does need to read the state, I think the read does not happen correctly (count is not the size of the variable) .

On write, only 'time' (lower case) can be the unlimited dimension. (This is another assumption in dart. It is a weakness in how we deal with unlimited dimensions (https://github.com/NCAR/DART/issues/359#issuecomment-1187818595))
For dart created files, dart won't add an unlimited dimension unless its name is 'time'.  So creating and writing a file with fill inflation restart gets you a file with the correct info.  

All the assumptions above are a bit of a hack.  

I put a fix on https://github.com/NCAR/DART/tree/fix-unlimited_dim-read which checks the variable for unlimited dimension before adjusting the counts reading and writing.  It is a narrow solution (it really just fixes the case where there is an unused unlimited dimension).

I'm not sure if the cohorts dimension will be used in the state, if it is then improving the state read/write to cope with various (and multiple) unlimited dimensions gets bumped up the priority list. 

Let me know if this makes sense or not.

_Originally posted by @hkershaw-brown in https://github.com/NCAR/DART/issues/519#issuecomment-1644469352_

robin-armstrong:
## Description:

Contains model-mods for data assimilation with the Marine Biogeochemistry Library (MARBL/MOM6).

also fixes #614 

## Testing Datasets:

A test package for MARBL_joint_estimation can be found at `/glade/u/home/rarmstrong/DART_test_package`.


hkershaw-brown:
Thanks for this Robin!

For reviewers, summary of July 11th 2024 meeting as todo list:

model mom6-marbl:
- [ ] MARBL_joint_estimation is the version of the model directory that should be released.
- [ ] remove other MARBL_ model directories 
    - [ ] MARBL_param_estimation
    - [ ] MARBL_state_estimation
- [ ] documentation
    - [ ] options to set in marbl-mom6 to do DA.
    - [ ] what is available (parameter estimation?) models/MARBL_param_estimation has metadata.txt is this needed?
 
observation converter(s):
- [ ] check the difference between these:
  - [ ] BATS
  - [ ] BATS_clim 
 - [ ] documentation
   - [ ] observations/obs_converters/BATS_clim/bats_climatology.py 

General git tidiness:
- [ ] remove netcdf files (may remove from history if they are large)
- [ ] remove binary files (may remove from history if they are large)
- [ ] revert mpi/fix system changes
- [ ] update to main
- [ ] rewrite history so Robin's commit 
- [ ] .gitignores dotted around, remove or move to root/.gitignore

Testing:
- [ ] test case /glade/work/hkershaw/DART/marbl/DART_test_package (copy of /glade/u/home/rarmstrong/DART_test_package)

backup of this branch since we're rewriting history:
https://github.com/hkershaw-brown/DART/tree/marbl-dart

assimilation_code/modules/observations/default_quantities_mod.f90
```diff
@@ -167,7 +170,9 @@
 !   QTY_LIVE_STEM_CARBON
 !   QTY_LIVE_STEM_NITROGEN
 !   QTY_MEAN_SOURCE
+!   QTY_MESOZOOPLANKTON_CARBON
 !   QTY_MICROWAVE_BRIGHT_TEMP
+!   QTY_MICROZOOPLANKTON_CARBON
```
hkershaw-brown, line 175: default_quantities_mod is missing many of the new ocean qtys (ocean_quantities_mod.f90)

assimilation_code/modules/observations/ocean_quantities_mod.f90
```diff
@@ -32,11 +32,53 @@
...
 !   QTY_DISSOLVED_ORGANIC_NITROGEN
 !   QTY_DISSOLVED_ORGANIC_P
 !   QTY_DISSOLVED_INORGANIC_IRON
 !   QTY_SURFACE_CHLOROPHYLL
 !   QTY_LAYER_THICKNESS
+!   QTY_COLUMN_DEPTH
+!   QTY_DISSOLVED_INORGANIC_SIO3
+!   QTY_MESOZOOPLANKTON_CARBON
+!   QTY_MICROZOOPLANKTON_CARBON
+!   QTY_PARAM_AUTOTROPH1_KCO2
```
hkershaw-brown, line 45: What's the goal of having _PARAM_ in the QTY name?  Just to communicate that it is a parameter?

Do you need a qty for each parameter?  Could you have QTY_PARAMETER?

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+character(len=256), parameter :: source = "model_mod.f90"
+logical :: module_initialized = .false.
+integer :: state_dom_id  ! used to access the state structure
+integer :: param_dom_id  ! used to access MARBL internal parameters
+integer :: nfields ! number of fields in the state or parameter vector
+integer :: nz      ! the number of vertical layers
+integer :: model_size
+type(time_type) :: assimilation_time_step
+real(r8), parameter :: geolon = 360 - 64.0
+real(r8), parameter :: geolat = 31.0
```
hkershaw-brown, line 87: these are hardcoded, should these be namelist options for people to put the column anywhere on Earth?

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+           param_qty_list(modelparams_table_height)
+real(r8):: state_clamp_vals(modelvar_table_height, 2), &
+           param_clamp_vals(modelparams_table_height, 2)
+logical :: update_var_list(modelvar_table_height), &
+           update_param_list(modelparams_table_height)
+
+module_initialized = .true.
+
+! Print module information to log file and stdout.
+call register_module(source)
```
hkershaw-brown, line 138:
```suggestion
```

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+character(len=*), intent(in) :: filename
+type(time_type) :: read_model_time
+
+integer :: ncid
+character(len=*), parameter :: routine = 'read_model_time'
+real(r8) :: days
+type(time_type) :: mom6_time
+integer :: mom_base_date_in_days, mom_days
+
+mom_base_date_in_days = 139157 ! 1982 1 1 0 0
```
hkershaw-brown, line 200: Is this always Jan 1st 1982?

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+!------------------------------------------------------------------
+! Given a state handle, a location, and a state quantity,
+! interpolates the state variable fields to that location and returns
+! the values in expected_obs. The istatus variables should be returned as
+! 0 unless there is some problem in computing the interpolation in
+! which case a positive istatus should be returned.
+!
+! For applications in which only perfect model experiments
+! with identity observations (i.e. only the value of a particular
+! state variable is observed), this can be a NULL INTERFACE.
```
hkershaw-brown, line 235:
```suggestion
```

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+                            num_close, close_ind, dist, ens_handle)
+
+
+end subroutine get_close_state
+
+
+!------------------------------------------------------------------
+! Does any shutdown and clean-up needed for model. Can be a NULL
+! INTERFACE if the model has no need to clean up storage, etc.
+
```
hkershaw-brown, line 463:
```suggestion
```

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+if ( .not. module_initialized ) call static_init_model
+
+! put file into define mode.
+
+call nc_begin_define_mode(ncid)
+
+call nc_add_global_creation_time(ncid)
+
+call nc_add_global_attribute(ncid, "model_source", source )
+call nc_add_global_attribute(ncid, "model", "template")
```
hkershaw-brown, line 487:
```suggestion
call nc_add_global_attribute(ncid, "model", "marbl-column")
```

or whatever the model_mod ends up being called.

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+
+call nc_get_variable_size(ncid, 'Layer', nz)
+
+call nc_close_file(ncid)
+
+end subroutine read_num_layers
+
+!------------------------------------------------------------
+! ocean_geom are 2D state sized static data
+! HK Do these arrays become too big in high res cases?
```
hkershaw-brown, line 632:
```suggestion
```

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+subroutine read_ocean_geometry()
+
+integer :: ncid
+
+character(len=*), parameter :: routine = 'read_ocean_geometry'
+
+! Need nx, ny
```
hkershaw-brown, line 639:
```suggestion
```
I think this comment is from the MOM6 model_mod where you need nx, ny for the size of the basin_depth variable. The column marble is (2,2) always.

models/MARBL_joint_estimation/work/input.nml
```diff
@@ -0,0 +1,241 @@
...
+   /
+
+&preprocess_nml
+   input_obs_def_mod_file  = '/glade/derecho/scratch/rarmstrong/marbl_dart/DART/observations/forward_operators/DEFAULT_obs_def_mod.F90'
+   output_obs_def_mod_file = '/glade/derecho/scratch/rarmstrong/marbl_dart/DART/observations/forward_operators/obs_def_mod.f90'
+   input_obs_qty_mod_file  = '/glade/derecho/scratch/rarmstrong/marbl_dart/DART/assimilation_code/modules/observations/DEFAULT_obs_kind_mod.F90'
+   output_obs_qty_mod_file = '/glade/derecho/scratch/rarmstrong/marbl_dart/DART/assimilation_code/modules/observations/obs_kind_mod.f90'
+   obs_type_files          = '/glade/derecho/scratch/rarmstrong/marbl_dart/DART/observations/forward_operators/obs_def_ocean_mod.f90'
+   quantity_files          = '/glade/derecho/scratch/rarmstrong/marbl_dart/DART/assimilation_code/modules/observations/default_quantities_mod.f90', '/glade/derecho/scratch/rarmstrong/marbl_dart/DART/assimilation_code/modules/observations/ocean_quantities_mod.f90'
+   /
```
hkershaw-brown, line 178: needs changing to relative paths (within DART) rather than pointing at Robin's directories.

models/MARBL_joint_estimation/work/input.nml
```diff
@@ -0,0 +1,241 @@
+&perfect_model_obs_nml
```
hkershaw-brown, line 1: will need qceff namelists in this file.

observations/obs_converters/BATS/README.md
```diff
@@ -0,0 +1,2 @@
+This folder contains files adapted from those in DART/observations/obs\_converters/text, for the purpose of converting data from the [Bermude Atlantic Time-Series Study](https://bats.bios.asu.edu/) (BATS) into a format readable by DART.
```
hkershaw-brown, line 1: move to .rst file for online docs

observations/obs_converters/BATS_clim/README.md
```diff
@@ -0,0 +1,2 @@
+This folder contains files adapted from those in DART/observations/obs\_converters/text, for the purpose of converting data from the [Bermude Atlantic Time-Series Study](https://bats.bios.asu.edu/) (BATS) into a format readable by DART.
```
hkershaw-brown, line 1: move to .rst for online docs.
Description needed for BATS_clim vs BATS

hkershaw-brown (changes requested):
first batch of review comments

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+mom_days = int(days) + mom_base_date_in_days
+
+read_model_time = set_time(0,mom_days)
+
+end function read_model_time
+
+!------------------------------------------------------------------
+! Returns the number of items in the state vector as an integer. 
+
+function get_model_size()
```
mgharamti, line 216: This is not very clean, especially if the parameter domain is not active. I would count the domains and loop over them to get the `model_size` (which is by the way defined on line 84 but not used). I suggest a code similar to the one below: 
```
domain_count = get_num_domains()
model_size = 0
do i_dom = 1,domain_count
     model_size = model_size + get_domain_size(i_dom)
enddo
```
This is related to my previous comment on active domains and the nature of the DA configuration

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+        case default
+            write(string1,'(A)')  'only UPDATE or NO_COPY_BACK supported in model_state_variable namelist'
+            write(string2,'(6A)') 'you provided : ', trim(varname), ', ', trim(dartstr), ', ', trim(update)
+            call error_handler(E_ERR,'verify_state_variables',string1, text2=string2)
+    end select
+
+    ! reading the clamp values
+
+    if (table(i, 3) /= 'NA') then
+        read(table(i,3), '(d16.8)') clamp_vals(i,1)
```
mgharamti, line 569: Is the format specification needed here? What if people added the clamps in a different format?

models/MARBL_param_estimation/model_mod.f90
```diff
@@ -0,0 +1,591 @@
+! DART software - Copyright UCAR. This open source software is provided
```
mgharamti, line 1: I thought we agreed that we will only do the joint state-parameter estimation one. I would stick to one marbl directory and one model mod for now. These could be consolidated into one in the future. I mean something that comes to mind now is to do no_copy_back  on the state variables to turn it into a parameter estimation only.

mgharamti (commented):
This is my first rundown of the code. I have not run the code but I plan to do that soon.

assimilation_code/modules/io/direct_netcdf_mod.f90
```diff
@@ -858,16 +858,17 @@ subroutine read_variables(ncfile_in, var_block, start_var, end_var, domain)
    counts(:) = 1
 
    slice_start(:) = 1 ! default to read all dimensions start at 1
-
-   if (has_unlimited_dim(domain)) then
+   
+   ret = nf90_inquire(ncfile_in, unlimitedDimID=unlim_dimID)
+   call nc_check(ret, 'read_variables: nf90_inquire', 'unlimitedDimID')
+   
+   if (has_unlimited_dim(domain) .and. any(get_io_dim_ids(domain, i) ==  unlim_dimID )) then
```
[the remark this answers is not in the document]
hkershaw-brown: yup it is an issue already #614 
It affects models which have some state variables with an unlimited dimension and some without.

assimilation_code/modules/observations/ocean_quantities_mod.f90
robin-armstrong: If I remember correctly, I created multiple `QTY_PARAM_...` types in order to facilitate experiments where multiple parameters are being estimated at once. For experiments of this type, we need a way to distinguish one parameter from another, although they do indeed have the same physical properties. But maybe there's some way for us to get by with only one QTY code which I'm not seeing.

models/MARBL_joint_estimation/model_mod.f90
robin-armstrong: Unless changes have been made to the MARBL date/time format, yes, it's always Jan 1st 1982.

models/MARBL_param_estimation/model_mod.f90
robin-armstrong: Yes, we did agree to only do the joint state-parameter estimation model-mod. I had all three model-mods being developed on the same branch to make it easier to switch between them, which is why they're all in the same pull request; I apologize if that's bad form. In any case, `MARBL_joint_estimation` is the one that we agreed to pull into DART.

models/MARBL_joint_estimation/model_mod.f90
```diff
@@ -0,0 +1,654 @@
...
+                            state_qty_list, state_clamp_vals, update_var_list)
+
+    state_dom_id = add_domain(template_file(1), nfields, &
+                              var_names = variable_table(1:nfields, 1), &
+                              kind_list = state_qty_list(1:nfields), &
+                              clamp_vals = state_clamp_vals, &
+                              update_list = update_var_list(1:nfields))
+
+! setting up the DART parameter vector
+if(estimate_params) then
```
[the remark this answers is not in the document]
robin-armstrong: It's been a while since I was developing the parameter-only version, but as I remember, that setup still needs a separate domain for state variables. The reasoning here was that the forward operator in the parameter-only case involves running MARBL to produce climatological averages, and this operation cannot be coded into DART. Maintaining a separate domain for state variables allows us to compute those averages externally, so that DART can read them as identity observations from a "climatology" component of the state vector. That said, it might be better to rename that component within the model-mod, so that instead of `state_dom_id` we have something like `climatology_dom_id`.

models/MARBL_joint_estimation/model_mod.f90
robin-armstrong: I don't see any reason to not make them namelist options, but a user will need to know that any mismatch between the location in the obs-seq file and `geolat`/`geolon` will cause DART to reject the observation. I think that was our motive for hard-coding those coordinates. We wanted to make sure that they always agreed with the coordinates written in the obs-seq file, which in turn are hard-coded into the obs converter.

hkershaw-brown:
note for test package, I switched to inflation not read from file

hkershaw@dec1004:/glade/work/hkershaw/DART/marbl/DART_test_package/test/DART$ diff input.nml input.nml.robin 
1,6d0
< &probit_transform_nml
< /
< 
< &algorithm_info_nml
< /
< 
73,74c67,68
<    inf_initial_from_restart    = .false.,                 .false.,
<    inf_sd_initial_from_restart = .false.,                  .false.,
---
>    inf_initial_from_restart    = .true.,                 .false.,
>    inf_sd_initial_from_restart = .true.,                  .false.,
108a103
>    filter_kind                     = 1,

mgharamti:
I checked out the code and I was not able to run any of the DART executables. Here is the error that I was getting: 
```
ERROR FROM:
  source : ensemble_manager_mod.f90
  routine: set_up_ens_distribution
  message: not enough MPI tasks for the model size, suggest at least ***** tasks
  ```
  So, it turns out the length of the state vector is `7036994676854122`!!
  I realize the issue is in computing the variable `model_size`. Right now, the function `get_model_size` in the `model_mod` is computing it using local variables. When someone calls `get_model_size()` outside the `model_mod`, this fails. So, the correct approach is to compute `model_size` inside `static_init_model` as I outlined in my previous review and then `get_model_size` should simply look like this: 
  ```
function get_model_size()
integer(i8) :: get_model_size
get_model_size = model_size 
end function get_model_size
```
With this change, I can now run `model_mod_check` and the `model_size` is now correctly computed as `3434`. 
It's still puzzling to me that this version of the code was running during Robin's internship. Could it be that we have a different version?

mgharamti (changes requested):
I was able to run filter and state and parameter updates. The only necessary change that is needed to run filter properly is the handling of `model_size`.

models/MARBL_joint_estimation/work/input.nml
```diff
@@ -0,0 +1,242 @@
...
+   output_timestamps          = .false.,
+   print_every_nth_obs        = -1,
+   output_forward_op_errors   = .false.,
+   silence                    = .false.,
+   /
+
+&filter_nml
+   single_file_in               = .false.,
+   input_state_files = '',
+   input_state_file_list = 'states_input.txt', 'params_input.txt',
```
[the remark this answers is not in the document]
hkershaw-brown: > If I set `estimate_params = .false.` in the `model_nml`, the code will still look for a list of parameter files. I guess that's ok given that this is a joint state-parameter estimation routine. This can be easily fixed once we work on merging the 3 different filtering strategies together.

merging the 3 different filtering strategies is happening as part of this pull request, right?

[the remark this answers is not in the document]
mgharamti: Helen, I'm not quite sure. I'm happy to do it but it would take some time as I'm leaving. I say, we push the joint one now after small corrections. Once I'm back, I can get to it and add flexibility to perform the 3 using one code.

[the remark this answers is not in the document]
hkershaw-brown: There is quite a bit to do before releasing on this code.  I'll go ahead an close the pull request.  We can reopen it when it its ready.

hkershaw-brown:
pushed robins branch to https://github.com/NCAR/DART/tree/marbl-dart

hkershaw-brown:
> pushed robins branch to https://github.com/NCAR/DART/tree/marbl-dart
deleting this branch, since Moha's pull request from his fork:

https://github.com/mgharamti/DART/tree/marbl-dart

PR #735, opened 2024-09-10
remove copy of state into mpi window

### LINKED ISSUE #718: Feature request: Reduce memory usage for forward operators MPI window
### Use case  
Currently doing an copy of data from the ensemble handle to the window

Not incorrect results, but using more memory per core than we need to. 

Create the window with the whole state_ens_handle%copies array. 'simply contiguous' <sup>1</sup>
get_state  copies_in_window != copies you need to get so the window logic would need to be updated for the whole %copies array

1. MPI_Win_create In Fortran, one can pass the first element of a memory region or a whole array, which must be 'simply contiguous' (for 'simply contiguous', see also MPI 3.0, Section 17.1.12 on page 626).
https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report.pdf

[Spec](https://docs.google.com/document/d/1XVpP9tiBYn4zSu5R6BmOOpwTp9zKTck-ctVQ2Tr_lCs/edit?usp=sharing)

None

hkershaw-brown:
## Description:

Puts the %copies array into the mpi window for get_state (forward operator)
Removes the copy of the %copies(1:real_ens_members, :) to the window.
This reduces the per-code memory usage (removes a %copies(1:real_ens_members, 1:my_num_vars)
~1/2 for the state per memory core usage.

The offset is calculated from %num_copies, the get_state is only the real ensemble members

ink

I think this is a good time to remove the cray window module. 

### Fixes issue

fixes #718

### Types of changes

- [ ] Documentation update

### Tests
There is a developer test included. Note this needs fortran-testanything code to run.
Bitwase lorenz_96 mpi, mpif08, nompi

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed


hkershaw-brown:
@hkershaw-brown check PMO

mjs2369 (commented):
Remove these identical comments -

[https://github.com/NCAR/DART/blob/8c549735550e7120ee926f3b194ee8f0cba5ada7/assimilation_code/modules/utilities/no_cray_win_mod.f90#L32-L33](https://github.com/NCAR/DART/blob/8c549735550e7120ee926f3b194ee8f0cba5ada7/assimilation_code/modules/utilities/no_cray_win_mod.f90#L32-L33)

[https://github.com/NCAR/DART/blob/8c549735550e7120ee926f3b194ee8f0cba5ada7/assimilation_code/modules/utilities/no_cray_winf08_mod.f90#L32-L33](https://github.com/NCAR/DART/blob/8c549735550e7120ee926f3b194ee8f0cba5ada7/assimilation_code/modules/utilities/no_cray_winf08_mod.f90#L32-L33)

assimilation_code/modules/utilities/no_cray_winf08_mod.f90
```diff
@@ -65,8 +61,7 @@ subroutine create_state_window(state_ens_handle, fwd_op_ens_handle, qc_ens_handl
 integer :: bytesize !< size in bytes of each element in the window
 integer :: my_num_vars !< my number of vars
 
-! Find out how many copies to put in the window
-! copies_in_window is not necessarily equal to ens_handle%num_copies
+! Find out how many copies to get, maybe different to %num_copies
```
mjs2369, line 64:
```suggestion
! Find out how many copies to get, maybe different to ens_handle%num_copies
```

assimilation_code/modules/utilities/no_cray_winf08_mod.f90
```diff
@@ -82,20 +77,15 @@ subroutine create_state_window(state_ens_handle, fwd_op_ens_handle, qc_ens_handl
    my_num_vars = state_ens_handle%my_num_vars
 
    call mpi_type_size(datasize, bytesize, ierr)
-   window_size = my_num_vars*data_count*bytesize
-
-   allocate(contiguous_fwd(data_count*my_num_vars))
-   contiguous_fwd = reshape(state_ens_handle%copies(1:data_count, :), (/my_num_vars*data_count/))
+   window_size = my_num_vars*state_ens_handle%num_copies*bytesize
```
mjs2369, line 80:
```suggestion
   window_size = my_num_vars*data_count*bytesize
```
If we don't want to use data_count here and instead want to use state_ens_handle%num_copies, then we don't need the data_count variable at all. This contradicts the comment on line 64 though

assimilation_code/modules/utilities/no_cray_winf08_mod.f90
mjs2369: Same thing in no_cray_win_mod.f90

mjs2369:
Do we want to rename the window mods to not mention cray?

no_cray_win_mod.f90 -> win_mod.f90
no_cray_winf08_mod.f90 -> winf08_mod.f90

developer_tests/window/work/threed_model_mod.f90
```diff
@@ -0,0 +1,298 @@
...
+                                 nc_begin_define_mode, nc_end_define_mode
+
+use state_structure_mod, only : add_domain, get_domain_size
+
+use ensemble_manager_mod, only : ensemble_type
+
+! These routines are passed through from default_model_mod.
+! To write model specific versions of these routines
+! remove the routine from this use statement and add your code to
+! this the file.
```
mjs2369, line 37: remove this comment?

mjs2369:
In buildfunctions.sh and buildconvfunctions.sh, you can remove all mentions of `windowsrc`. It's not needed anymore. 

I just pushed up my commits from when I removed the cray window a while back if you want to take a look at these or cherry pick and commits. It has `windowsrc` already taken out of the code. It also has the window files already renamed and updates buildfunctions.sh and buildconvfunctions.sh accordingly

assimilation_code/modules/utilities/no_cray_winf08_mod.f90
hkershaw-brown: data_count is the number of copies that get_state will get, state_ens_handle%num_copies is all the copies
Screenshot 2024-10-31 at 10 56 02 AM

developer_tests/window/work/threed_model_mod.f90
hkershaw-brown: This is the template model_mod, it it just in there to compile the tests.

hkershaw-brown: I can change the quickbuild.sh to get the template model mod rather than a copying (lazy see commit https://github.com/NCAR/DART/commit/9d0f897b0827fcfef8ecd564d03237eb27d14247)

hkershaw-brown:
> In buildfunctions.sh and buildconvfunctions.sh, you can remove all mentions of `windowsrc`. It's not needed anymore.
> 
> I just pushed up my commits from when I removed the cray window a while back if you want to take a look at these or cherry pick and commits (on branch remove_craywin). It has `windowsrc` already taken out of the code. It also has the window files already renamed and updates buildfunctions.sh and buildconvfunctions.sh accordingly

Good point. I shall grab your changes from remove_craywin. 
Since we only have null/mpi/mpif08 windows we don't have to have a separate window file (cray vs not cray). I quite like the window being in a separate file rather than adding more stuff to {null|mpi}_utilties_mod.f90

developer_tests/window/work/threed_model_mod.f90
mjs2369: Doesn't matter if we aren't going to put the tests in the repo, but if we do I think we should use the template

hkershaw-brown:
> In buildfunctions.sh and buildconvfunctions.sh, you can remove all mentions of `windowsrc`. It's not needed anymore.
> 
> I just pushed up my commits from when I removed the cray window a while back if you want to take a look at these or cherry pick and commits (on branch remove_craywin). It has `windowsrc` already taken out of the code. It also has the window files already renamed and updates buildfunctions.sh and buildconvfunctions.sh accordingly

cherry-picked your commits onto this branch

developer_tests/window/work/threed_model_mod.f90
hkershaw-brown: used template, added fortran-test anything to developer tests.

assimilation_code/modules/utilities/no_cray_winf08_mod.f90
hkershaw-brown: done

hkershaw-brown:
> Remove these identical comments -
> 
> https://github.com/NCAR/DART/blob/8c549735550e7120ee926f3b194ee8f0cba5ada7/assimilation_code/modules/utilities/no_cray_win_mod.f90#L32-L33
> 
> https://github.com/NCAR/DART/blob/8c549735550e7120ee926f3b194ee8f0cba5ada7/assimilation_code/modules/utilities/no_cray_winf08_mod.f90#L32-L33

done

mjs2369 (approved):
Everything looks good, developer test working well.

PR #771, opened 2024-11-06
fix: outgoing vertical coordinate should be 'height' in meters using zGridCenter

### LINKED ISSUE #524: bug: when localizing in height, the mpas_atm model_mod may compute an incorrect vertical location
### Describe the bug

In reviewing the mpas_atm model_mod code, the subroutine find_vert_indices()  calls find_vert_level() which calls find_height_bounds().  find_height_bounds uses either the zGridCenter array or the zGridEdge for vertical locations.  it returns the lower and upper model level numbers that enclose that vertical location, along with the fraction of the level.   

in convert_vert_distrib() if the requested outgoing vertical is VERTISHEIGHT, after computing the level numbers based on cell centers, it indexes into the zGridFace array for the heights instead of zGridCenter or zGridEdge.

this seems wrong and needs to be tested - possibly with a version of model_mod_check which includes a call to convert_vert.

note that if W interpolation is added, this code will need to distinguish between cell centered data in the vertical and data on the cell faces, where the existing code might be correct for the new case.

### To reproduce/test the bug

set the namelist so it will localize in the vertical on height.
test a location of a variable located on the cell center (T, Q) in the vertical on a known level (VERTISLEVEL)
look at the location after vertical conversion.  the height should match the height of the cell center, not the cell face.

### Error Message  
This will not provoke any error messages, but would increase the distances used in localization by half a vertical cell.

mpas_atm

### Version of DART      
This code is in the main branch

there is a modified version that includes other changes which is not in the dart repo. 

### Build information
This was discovered by inspecting the code.  The tests can be run anyplace but based on the size of the mpas test cases, it probably needs to be run on cheyenne.

hkershaw-brown:
## Description:

mpas, observation vertical conversion was using zGridFace rather than zGridCenter

Following the call tree, using oncenters always

```
convert_vertical_obs
  -> convert_vert_distrib
     case (VERTISHEIGHT)
     -> find_vert_indices
        -> find_vert_level(oncenters = true)
           case (VERTEIGHT)
             if (oncenters)
             -> find_height_bounds(zGridCenter)
```

### Fixes issue

fixes #524

### Types of changes

- [ ] Documentation update

### Tests

Skipped the early return so an obs with VERTISHEIGHT gets converted to VERTISHEIGHT.
https://github.com/NCAR/DART/compare/main...hkershaw-brown:DART:bug_524_reproducer
see https://github.com/NCAR/DART/issues/524#issuecomment-2447818389 for results.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

You can grab my one obs test from /glade/derecho/scratch/hkershaw/DART/Bugs/bug_524/runs_524


mjs2369 (approved):
Went ahead and tested this change using the one obs test in `/glade/derecho/scratch/hkershaw/DART/Bugs/bug_524/runs_524` and print statements committed to https://github.com/hkershaw-brown/DART/tree/bug_524_reproducer 

With the code skipping the early return for ztypein == ztypeout == VERTISHIEGHT, the 
vert values coming in and out of convert_vert_distrib are the same when using zGridCenter instead of zGridFace, as they should be. 

```
MS zin(:)    12045.3370000000
MS zout(:)   12045.3370000000
```

PR #783, opened 2024-12-18
Generalized subroutine to read the state variables from the namelist (parse_variables)

### LINKED ISSUE #448: Feature request: routine to read a table of state variables from namelist
### Use case  

reading a table of state variables from a namelist, there are several of these in various model_mods (not including Lanai only models):

[parse_variable_table](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/clm/model_mod.f90#L551)
[verify_variables](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/tiegcm/model_mod.f90#L897)  [verify_variables](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/wrf_hydro/model_mod.f90#L1232)
[vertify_variables](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/wrf_hydro/model_mod.f90#L1232)
[verify_state_variables](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/MOM6/model_mod.f90#L875)
[verify_state_variables](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/mpas_atm/model_mod.f90#L3803)
[verify_state_variables](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/cice/model_mod.f90#L2598)
[verify_state_variables](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/POP/model_mod.f90#L2757)
[fill_dart_kinds_table](https://github.com/NCAR/DART/blob/1b76f3afa5978469d6a119710bb638c1c077ae20/models/wrf/model_mod.f90#L7420)

Every time a new model is added to DART, we get a variation on one of these routines. 

Code repetition and all the associated problems that come with that.

 `vtablenamelength` is in types_mod. This is one of several non-types in types_mod 
https://github.com/NCAR/DART/blob/16595cad595b6349605c9bc8e021c7ecc2c2391a/assimilation_code/modules/utilities/types_mod.f90#L24-L26 

Generalize the routine:
netcdf variable name, dart qty, update

Define the state outside the model_mod

mjs2369:
## Description:

There are many variations in the model_mods of the subroutine to take the table of state variables read in from the namelist and put the info into various data structures, which can then be passed into add domain. A new variation of this subroutine would need to be written every time a new model is added to DART.

This PR creates two generalized subroutines that convert the table of state variables to a state_var_type : 
parse_variables reads netcdf variable name, dart qty, and update values from the nml
parse_variables_clamp reads netcdf variable name, dart qty, clamp_values, and update values from the nml

Alternate versions for this subroutine were replaced with get_state_variables in the models MOM6, wrf_hydro, aether_lat-lon, cam-fv, cam-se, POP, and cice

New dev test test_parse_variables added

### Fixes issue
Fixes #448 

### Types of changes

- [ ] Documentation update

### Tests
Compiled all models with updated model_mods
Build_everything runs as expected with all compilers
Bitwise tested with CAM-FV
New dev test runs successfully and reports pass/fails correctly

## Checklist for merging

- [x] Updated changelog entry
- [x] Documentation updated
- [x] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed


hkershaw-brown:
@mjs2369 Do you have restart files so you can test the changed model_mods?

models/utilities/default_model_mod.f90
```diff
@@ -283,6 +300,162 @@ subroutine pert_model_copies(state_ens_handle, ens_size, pert_amp, interf_provid
...
+!--------------------------------------------------------------------
+
+!> Reads in model_nml:model_state_variables and returns a
+!> state_var_type state_vars with nvars ; netcdf variable names ;
+!> qtys (kinds) ; clamp values ; updates
+!
+!> Verifies that the namelist was filled in correctly, and checks
+!> that there are valid entries for the dart_kind.
+
+subroutine get_state_variables_clamp(nml_state_vars, MAX_STATE_VARIABLES, use_clamping, state_vars)
```
hkershaw-brown, line 312: What's the thinking in have MAX_STATE_VARIABLES passed in vs. in default_model_mod?

models/utilities/default_model_mod.f90
```diff
@@ -283,6 +300,162 @@ subroutine pert_model_copies(state_ens_handle, ens_size, pert_amp, interf_provid
...
+
+character(len=*),   intent(inout) :: nml_state_vars(:)
```
hkershaw-brown, line 314: nitpick: don't use nml in a variable name unless it is a namelist

models/utilities/default_model_mod.f90
```diff
@@ -283,6 +300,162 @@ subroutine pert_model_copies(state_ens_handle, ens_size, pert_amp, interf_provid
...
+type(state_var_type), intent(out) :: state_vars
+
+character(len=NF90_MAX_NAME) :: netcdf_var_name, dart_qty_str, update
+character(len=256) :: string1, string2
+integer :: i, ivar
+
+if(.not. use_clamping) then
+   string1 = 'logical use_clamping set to false when values for clamping are required in this model nml'
+   call error_handler(E_ERR, 'get_state_variables_clamp', string1)
+endif
```
hkershaw-brown, line 326: I don't follow the logic of the use_clamping argument and error checking for it. 

call get_state_variables_clamp
or call get_state_variables

If you are overloading then the arguments should be different, so no need to check.

If you're overloading but also passing a logical for what the routine should do, it is a good time to step back and reassess whether overloading is a good thing to do here.

models/utilities/default_model_mod.f90
```diff
@@ -283,6 +300,162 @@ subroutine pert_model_copies(state_ens_handle, ens_size, pert_amp, interf_provid
 
 end subroutine pert_model_copies
 
+!--------------------------------------------------------------------
+
+!> Reads in model_nml:model_state_variables and returns a
+!> state_var_type state_vars with nvars ; netcdf variable names ;
+!> qtys (kinds) ; clamp values ; updates
```
hkershaw-brown, line 307: This comment is not true, there is no namelist reading in this routine

models/utilities/default_model_mod.f90
```diff
@@ -283,6 +300,162 @@ subroutine pert_model_copies(state_ens_handle, ens_size, pert_amp, interf_provid
...
+         write(string1,'(A)')  'Invalid update variable in &model_nml:model_state_variable - only UPDATE or NO_COPY_BACK are supported'
+         write(string2,'(6A)') 'Issue: ', trim(netcdf_var_name), ', ', trim(dart_qty_str), ', ', trim(update)
+         call error_handler(E_ERR,'verify_state_variables',string1, text2=string2)
+   end select
+
+   ! Checking that the rows in the nml entry are all complete
+   if ( dart_qty_str == '' .or. update == '' ) then
+      string1 = 'model_nml:model_state_variables not fully specified'
+      call error_handler(E_ERR, 'verify_state_variables', string1)
+   endif
```
hkershaw-brown, line 453: Do your unit tests catch this?

models/wrf_hydro/model_mod.f90
```diff
@@ -140,9 +138,9 @@ module model_mod
 character(len=256)  :: perturb_distribution           = 'lognormal'
 
 character(len=obstypelength) :: &
-    lsm_variables(  NUM_STATE_TABLE_COLUMNS,MAX_STATE_VARIABLES) = '', &
-    hydro_variables(NUM_STATE_TABLE_COLUMNS,MAX_STATE_VARIABLES) = '', &
-    parameters(     NUM_STATE_TABLE_COLUMNS,MAX_STATE_VARIABLES) = ''
+    lsm_variables(  NUM_STATE_TABLE_COLUMNS*MAX_STATE_VARIABLES) = '', &
+    hydro_variables(NUM_STATE_TABLE_COLUMNS*MAX_STATE_VARIABLES) = '', &
+    parameters(     NUM_STATE_TABLE_COLUMNS*MAX_STATE_VARIABLES) = ''
```
hkershaw-brown, line 143: Why have N*M here?

You could have default_model_mod parameter that is = N*M

models/utilities/default_model_mod.f90
```diff
@@ -283,6 +300,162 @@ subroutine pert_model_copies(state_ens_handle, ens_size, pert_amp, interf_provid
...
+!> Reads in model_nml:model_state_variables and returns a
+!> state_var_type state_vars with nvars ; netcdf variable names ;
+!> qtys (kinds) ; updates
+!
+!> Verifies that the namelist was filled in correctly, and checks
+!> that there are valid entries for the dart_kind.
+
+subroutine get_state_variables_noclamp(nml_state_vars, MAX_STATE_VARIABLES, state_vars)
+
+character(len=*),   intent(inout) :: nml_state_vars(:)
```
hkershaw-brown, line 400: nitpick: not a namelist, only an character array

models/wrf_hydro/model_mod.f90
```diff
@@ -110,17 +111,14 @@ module model_mod
...
 integer(i8) :: model_size
 type(time_type) :: time_step
 
-! Codes for interpreting the columns of the variable_table
-integer, parameter :: VT_VARNAMEINDX  = 1 ! ... variable name
-integer, parameter :: VT_KINDINDX     = 2 ! ... DART kind
-integer, parameter :: VT_MINVALINDX   = 3 ! ... minimum value if any
-integer, parameter :: VT_MAXVALINDX   = 4 ! ... maximum value if any
-integer, parameter :: VT_STATEINDX    = 5 ! ... update (state) or not
+! For interpreting the columns of the variable_table
```
hkershaw-brown, line 121: Not sure this comment is true anymore, it is just M*N no info about which column is which

models/wrf_hydro/model_mod.f90
```diff
@@ -250,13 +242,13 @@ subroutine static_init_model()
 
       call configure_hydro()
       call read_hydro_global_atts(domain_shapefiles(domainID))
-      call verify_variables(hydro_variables, domain_shapefiles(domainID), &
-                  n_hydro_fields, var_names, var_qtys, var_ranges, var_update)
+
+      call get_state_variables(hydro_variables, MAX_STATE_VARIABLES, use_clamping, state_vars_hydro)
```
hkershaw-brown, line 246: It is worth thinking about the name 'get_state_variables'.
Here it is called for three domains, so it is only 'getting' info for one domain, not the whole state.

I think also  it is good to think about what the routine is doing. It is not getting the data of any netcdf variables, it is just converting a table of characters to somehting that can be passed to add_domain.

models/utilities/default_model_mod.f90
hkershaw-brown, line 312: Can this be a function not a subroutine?

state_vars = get_state_variables_clamp(..)

models/MOM6/model_mod.f90
```diff
@@ -180,15 +170,16 @@ subroutine static_init_model()
...
+call get_state_variables(model_state_variables, MAX_STATE_VARIABLES, state_vars)
 
 ! Define which variables are in the model state
-dom_id = add_domain(template_file, nfields, &
-                    var_names = variable_table(1:nfields, VAR_NAME_INDEX), &
-                    kind_list = state_qty_list(1:nfields), &
-                    update_list = update_var_list(1:nfields))
+dom_id = add_domain(template_file, state_vars%nvars, &
+                    var_names = state_vars%netcdf_var_names(1:state_vars%nvars), &
+                    kind_list = state_vars%qtys(1:state_vars%nvars), &
```
hkershaw-brown, line 181: Why not overload add_domain with state_var_type?

```
dom_id = add_domain(state_vars)
```

hkershaw-brown: If get_state_variables is a function:

```
dom_id = add_domain(get_state_variables(table))
```

hkershaw-brown (changes requested):
Hi Marlee, 

Good work finding all the model_mods routines that can be switched out by this. 

I've put some comments in. The thing to sort out first, is the `get_state_variables` routine.  
1. Does the overloading and having to set clamping = .true. make. sense? '
2. Can you keep state_vars type private, so the model_mod does not access the internals (see the comments on add_domain)?

The 3d template model_mod needs updating to use the new routines, since this is what is used when the 'new_model.sh' script is run. 

Documentation for routine also needs adding, maybe  'adding a new model' is the best place for this, I'm not sure. 

I think the tests should comprehensively test the whole table read - if we are running the test, should test all the values.

mjs2369:
> @mjs2369 Do you have restart files so you can test the changed model_mods?

Only for cam-fv and pop - do you want me to test that all models are bitwise?

hkershaw-brown:
> > @mjs2369 Do you have restart files so you can test the changed model_mods?
> 
> Only for cam-fv and pop - do you want me to test that all models are bitwise?

If you're changing the model_mod then yes, check that the new code does the expected.
Bitwise model_mod_check, or pmo, or filter if you have test cases set up - with clamping. 
Or compare the domain info.

developer_tests/namelist/test_get_state_variables.f90
```diff
@@ -0,0 +1,78 @@
...
+
+call get_state_variables(state_variables, MAX_STATE_VARIABLES, state_vars)
+
+write(*,*) 'state_vars%nvars: ', state_vars%nvars
+write(*,*) 'state_vars%netcdf_var_names: ', state_vars%netcdf_var_names
+write(*,*) 'state_vars%qtys: ', state_vars%qtys
+write(*,*) 'state_vars%updates: ', state_vars%updates
+
+call ok(state_vars%nvars == 5) !OK
+call ok(state_vars%netcdf_var_names(2) == '') !NOT OK
```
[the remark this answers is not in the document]
mjs2369: Yes, it's supposed to fail. We discussed this a bit, but I can still change it if you'd rather they all pass. Otherwise, I will update the comment to be clearer

models/utilities/default_model_mod.f90
mjs2369: Yeah I guess my thought here was that the character array is obtained directly from the nml, so that's why I added that - I'll change it

models/wrf_hydro/model_mod.f90
mjs2369: @hkershaw-brown do you have any suggestions for the name?

developer_tests/namelist/test_get_state_variables.f90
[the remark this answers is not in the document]
hkershaw-brown: why not have 
```
call ok(state_vars%netcdf_var_names(2) /= '') 
```
Check for not equal?

models/utilities/default_model_mod.f90
nancycollins: this is a utility routine, so you shouldn't predict or constrain how the calling code chooses to use it.  a model_mod could read a character array out of a config file that wasn't the namelist file, or construct one by an interactive script, couldn't it?

models/utilities/default_model_mod.f90
mjs2369: @hkershaw-brown I agree that MAX_STATE_VARIABLES should just be in default_model_mod. Why I ended up passing it in as an argument from the various model_mods is that the value for MAX_STATE_VARIABLES varies in each. For example, aether has it set to 100, some have it at 20, but most have it set to 10. Therefore, I thought that we might want to allow the various model_mods to use different values for MAX_STATE_VARIABLES as opposed to having it as a parameter in default_model_mod. 

So I guess my main question to do this instead is - what should we have be the value for MAX_STATE_VARIABLES in default_model_mod? It seems inefficient to use 100, but then I'm not sure if this will be suitable for aether and other future models that might use more than 10 or 20

hkershaw-brown: Inefficient in terms of memory?  Are you worried about stack size?

(vtablenamelength = 64) * (5 columns) * (100 rows) = 31.74KB 
I would not worry about this. 

For computational efficiency, 

* you exit the loop early, so you only loop over the actual number of state variables
```
! Loop through the variables array to get the actual count of the number of variables
do ivar = 1, MAX_STATE_VARIABLES
   ! If the first element in the row is an empty string, the loop has exceeded the extent of the variables
   if (nml_state_vars(5*ivar-4) == '') then
      state_vars%nvars = ivar-1
      exit
   endif
enddo
```

* in get_state_variables_x
```
subroutine get_state_variables_clamp(nml_state_vars, MAX_STATE_VARIABLES, use_clamping, state_vars)

character(len=*),   intent(inout) :: nml_state_vars(:) <- this here you're better off being explicit about the (size)
                                                                                             also intent(out). The compiler can make better 
                                                                                            optimizations. But again tiny array, called once. Not that big 
                                                                                            a deal
integer,               intent(in) :: MAX_STATE_VARIABLES
logical,               intent(in) :: use_clamping
type(state_var_type), intent(out) :: state_vars
```

assimilation_code/modules/io/state_structure_mod.f90
```diff
@@ -365,6 +366,58 @@ function add_domain_from_file(info_file, num_vars, var_names, kind_list, clamp_v
...
+function add_domain_from_state_type(info_file, vars) result(dom_id)
+
+character(len=*), intent(in) :: info_file
+type(state_var_type), intent(in) :: vars
+integer :: dom_id
+
+integer :: ivar
+
+! add to domains
+call assert_below_max_num_domains('add_domain_from_file')
```
hkershaw-brown, line 386:
```suggestion
call assert_below_max_num_domains('add_domain_from_state_type')
```

hkershaw-brown:
Hi Marlee, it will be next week before I look at this, I'm not super worried about it, but it is a lot of model_mod changes to drop on a Friday! Have a good weekend

hkershaw-brown:
as in release on a Friday 'drop it like its hot'

developer_tests/namelist/test_parse_variables.f90
```diff
@@ -0,0 +1,81 @@
...
+state_vars = parse_variables(state_variables)
+
+call ok(state_vars%nvars == 3)
+call ok(state_vars%netcdf_var_names(1) == 'SALT_CUR')
+call ok(state_vars%netcdf_var_names(2) == 'TEMP_CUR')
+call ok(state_vars%netcdf_var_names(3) == 'UVEL_CUR')
+call ok(state_vars%qtys(1) == 263)
+call ok(state_vars%qtys(2) == 254)
+call ok(state_vars%qtys(3) == 356)
+call ok(state_vars%updates(1) == .true.)
```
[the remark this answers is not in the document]
hkershaw-brown: I'll go ahead and change to .eqv.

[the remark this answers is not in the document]
hkershaw-brown: You can just have ok(state_vars%updates(1)) if the value is .true.
Probably better in the tests to be explicit, and have 
state_vars%updates(1) == .true.

and 
state_vars%updates(1) == .false. (and have something set to false)

developer_tests/namelist/test_parse_variables.f90
```diff
@@ -0,0 +1,81 @@
...
+
+state_vars = parse_variables(state_variables)
+
+call ok(state_vars%nvars == 3)
+call ok(state_vars%netcdf_var_names(1) == 'SALT_CUR')
+call ok(state_vars%netcdf_var_names(2) == 'TEMP_CUR')
+call ok(state_vars%netcdf_var_names(3) == 'UVEL_CUR')
+call ok(state_vars%qtys(1) == 263)
+call ok(state_vars%qtys(2) == 254)
+call ok(state_vars%qtys(3) == 356)
```
hkershaw-brown, line 49: This hardcoding the QTY index, 
The QTY actual number is created by preprocess, from the qty files. 
So this will break if someone adds, for example, a new land qty

assimilation_code/modules/observations/land_quantities_mod.f90

! BEGIN DART PREPROCESS QUANTITY DEFINITIONS
!
!   QTY_BOURBON_BISCUIT <- the QTY numbers 263, 254, 356 will no longer match the QTY_strings you have
!   QTY_ABSORBED_PAR

developer_tests/namelist/work/input.nml
```diff
@@ -0,0 +1,67 @@
+&model_nml_clamp
+    state_variables_clamp = 'SALT_CUR ', 'QTY_SALINITY', '0.0', '0.0', 'UPDATE',
+                            'TEMP_CUR ', 'QTY_POTENTIAL_TEMPERATURE', '0.0', '0.0', 'UPDATE',
+                            'UVEL_CUR ', 'QTY_U_CURRENT_COMPONENT  ', '0', '0.0', 'UPDATE',
+  /
+
+&model_nml
+    state_variables = 'SALT_CUR ', 'QTY_SALINITY', 'UPDATE',
+                      'TEMP_CUR ', 'QTY_POTENTIAL_TEMPERATURE', 'UPDATE',
+                      'UVEL_CUR ', 'QTY_U_CURRENT_COMPONENT  ', 'update'
```
hkershaw-brown, line 10: nice mixture of uppercase lower and extra 'spaces    '.

developer_tests/namelist/test_parse_variables.f90
hkershaw-brown: fixed in e5057dadd609099e7405c9093fcb6baa39cc24f0

developer_tests/namelist/test_parse_variables.f90
```diff
@@ -0,0 +1,83 @@
...
+state_vars_clamp = parse_variables_clamp(state_variables_clamp)
+
+call ok(state_vars_clamp%nvars == 3)
+call ok(state_vars_clamp%netcdf_var_names(1) == 'SALT_CUR')
+call ok(state_vars_clamp%netcdf_var_names(2) == 'TEMP_CUR')
+call ok(state_vars_clamp%netcdf_var_names(3) == 'UVEL_CUR')
+call ok(state_vars_clamp%qtys(1) == QTY_SALINITY)
+call ok(state_vars_clamp%qtys(2) == QTY_POTENTIAL_TEMPERATURE)
+call ok(state_vars_clamp%qtys(3) == QTY_U_CURRENT_COMPONENT)
+call ok(state_vars_clamp%clamp_values(1,1) == 0.0)
```
hkershaw-brown, line 71: 0.0_r8

developer_tests/namelist/test_parse_variables.f90
```diff
@@ -0,0 +1,83 @@
...
+call ok(state_vars%nvars == 3)
+call ok(state_vars%netcdf_var_names(1) == 'SALT_CUR')
+call ok(state_vars%netcdf_var_names(2) == 'TEMP_CUR')
+call ok(state_vars%netcdf_var_names(3) == 'UVEL_CUR')
+call ok(state_vars%qtys(1) == QTY_SALINITY)
+call ok(state_vars%qtys(2) == QTY_POTENTIAL_TEMPERATURE)
+call ok(state_vars%qtys(3) == QTY_U_CURRENT_COMPONENT)
+call ok(state_vars%updates(1) .eqv. .true.)
+call ok(state_vars%updates(2) .eqv. .true.)
+call ok(state_vars%updates(3) .eqv. .true.)
```
hkershaw-brown, line 54: It would be good to test that state_vars%clamp_values is not allocated, since that is what your add_domain call is using to set clamping or not.

models/utilities/default_model_mod.f90
```diff
@@ -49,8 +51,24 @@ module default_model_mod
...
+          write_model_time,       &
+          parse_variables,        &
+          parse_variables_clamp,  &
+          state_var_type,         &
+          MAX_STATE_VARIABLE_FIELDS, &
+          MAX_STATE_VARIABLE_FIELDS_CLAMP
+
+type :: state_var_type
+    integer                        :: nvars
+    character(len=64), allocatable :: netcdf_var_names(:)
```
hkershaw-brown, line 63: the max length of a netcdf variable name is a constant set by the netcdf library

NF90_MAX_NAME

Its 256 at the moment:
https://github.com/Unidata/netcdf-fortran/blob/6302daa1376f076add3d97794195d070ab81fdd9/fortran/netcdf_constants.F90#L78
```suggestion
    character(len=NF90_MAX_NAME), allocatable :: netcdf_var_names(:)
```

models/utilities/default_model_mod.f90
```diff
@@ -283,6 +302,157 @@ subroutine pert_model_copies(state_ens_handle, ens_size, pert_amp, interf_provid
...
+
+   netcdf_var_name = trim(vars_table(5*i-4))
+   state_vars%netcdf_var_names(i) = trim(netcdf_var_name)
+
+   dart_qty_str = trim(vars_table(5*i-3))
+   call to_upper(dart_qty_str)
+   ! Make sure DART qty is valid
+   state_vars%qtys(i) = get_index_for_quantity(dart_qty_str)
+   if( state_vars%qtys(i) < 0 ) then
+      write(string1,*) 'The quantity specified in the &model_nml "', dart_qty_str, '", is not present in obs_kind_mod.f90'
```
hkershaw-brown, line 347: This write to the string will segfault, because string1 is 256 & dart_qty_str is 256 
256 characters in dart_qty_str + 85 characters in the message > 256 in string1

```suggestion
      write(string1,*) 'The quantity specified in the &model_nml "', trim(dart_qty_str), '", is not present in obs_kind_mod.f90'
```

models/MITgcm_ocean/model_mod.f90
```diff
@@ -533,11 +522,7 @@ subroutine static_init_model()
 if (do_output()) write(     *     , *) 'Using grid size : '
 if (do_output()) write(     *     , *) '  Nx, Ny, Nz = ', Nx, Ny, Nz
 
-call parse_variable_input(mitgcm_variables, model_shape_file, nvars, &
-                      var_names, quantity_list, clamp_vals, update_list)
-
-domain_id = add_domain(model_shape_file, nvars, &
-                    var_names, quantity_list, clamp_vals, update_list )
+domain_id = add_domain(model_shape_file, parse_variables_clamp(mitgcm_variables))
```
hkershaw-brown, line 525: I don't think you can replace the routine for MITgcm_ocean because the original code adjusts the clamping values if you are doing a log transform for BGC variables:

https://github.com/NCAR/DART/blob/f2deaa27499fea386d4b72d44a4134f4adf2ab62/models/MITgcm_ocean/model_mod.f90#L2113-L2123

models/POP/model_mod.f90
```diff
@@ -356,9 +338,10 @@ subroutine static_init_model()
 call init_interp()
 
 !> @todo 'pop.r.nc' is hardcoded in dart_pop_mod.f90
-domain_id = add_domain('pop.r.nc', nfields, &
-                       var_names = variable_table(1:nfields, VAR_NAME_INDEX), &
-                       update_list = update_var_list(1:nfields))
+! parse_variables converts the character table that was read in from
+! model_nml:model_state_variables and returns a state_var_type that
+! can be passed to add_domain
```
hkershaw-brown, line 343: This model_mod is missing the logic to use the default state variables if no model_state_variables = ' '
before calling add_domain

https://github.com/NCAR/DART/blob/f2deaa27499fea386d4b72d44a4134f4adf2ab62/models/POP/model_mod.f90#L2774-L2778

```
if model_state_variables = ' '
  call use_default_state_variables( state_variables )
  'blah blah using default state variables'
endif 
domain_id = add_domain('pop.r.nc', parse_variables(model_state_variables))

File: models/cice/model_mod.f90 [hunk 22]
```diff
@@ -431,10 +411,10 @@ subroutine static_init_model()
...
 ! The assimilate.csh, perfect_model.csh must ensure the cice restart file
 ! is linked to this filename.
-domain_id = add_domain('cice.r.nc', nfields, &
-                       var_names = variable_table(1:nfields, VAR_NAME_INDEX), &
-                       kind_list = state_kinds_list(1:nfields), &
-                       update_list = update_var_list(1:nfields))
+! parse_variables converts the character table that was read in from
+! model_nml:model_state_variables and returns a state_var_type that can be
+! passed to add_domain
+domain_id = add_domain('cice.r.nc', parse_variables(model_state_variables))
```
hkershaw-brown, line 417: Missing the logic for model_state_variables = ' ' 

https://github.com/NCAR/DART/blob/f2deaa27499fea386d4b72d44a4134f4adf2ab62/models/cice/model_mod.f90#L2618-L2622

models/cice/model_mod.f90
```diff
@@ -2684,7 +2530,7 @@ subroutine use_default_state_variables( state_variables )
 character(len=*),  intent(inout) :: state_variables(:)
 
 ! strings must all be the same length for the gnu compiler
-state_variables( 1:5*num_state_table_columns ) = &
+state_variables( 1:5*3 ) = &
```
hkershaw-brown, line 2533:
```suggestion
state_variables( 1:15 ) = &
```

hkershaw-brown (changes requested):
Hi Marlee, looks good.

I checked the template/threed_model_mod.f90 & threed_input.nml  with new_model.sh

There are 3 model_mods that do need changing:
POP, CICE to have default state variable tables if none is specified in the namelist
MITgcm I don't think you can replace because the clamping values depend on whether a log transform is applied to BGC variables

Cheers,
Helen

models/cice/model_mod.f90
mjs2369: Putting this change in the commit with other fixes to the POP and CICE model_mods

hkershaw-brown:
Marlee still testing this.

mjs2369:
Tested and ready to merge @hkershaw-brown 

Cam-fv: bitwise filter executions
Cam-se, Aether_lat-lon, wrf_hydro: bitwise model_mod_check executions
POP, MOM6, CICE: domain_info in the state_structure after the call to add_domain is identical to the main branch

models/utilities/default_model_mod.f90
```diff
@@ -283,6 +302,157 @@ subroutine pert_model_copies(state_ens_handle, ens_size, pert_amp, interf_provid
...
+integer :: i, ivar
+
+! Loop through the variables array to get the actual count of the number of variables
+do ivar = 1, MAX_STATE_VARIABLES
+   ! If the first element in the row is an empty string, the loop has exceeded the extent of the variables
+   if (vars_table(5*ivar-4) == '') then
+      state_vars%nvars = ivar-1
+      exit
+   endif
+enddo
```
hkershaw-brown, line 332: if  state_var%nvars == MAX_STATE_VARIABLES 

then MAX_STATE_VARIABLES might not be big enough.

models/utilities/default_model_mod.f90
```diff
@@ -49,8 +51,24 @@ module default_model_mod
...
+          MAX_STATE_VARIABLE_FIELDS_CLAMP
+
+type :: state_var_type
+    integer                        :: nvars
+    character(len=NF90_MAX_NAME), allocatable :: netcdf_var_names(:)
+    integer, allocatable           :: qtys(:)
+    real(r8), allocatable          :: clamp_values(:, :)
+    logical, allocatable           :: updates(:)
+end type state_var_type
+
```
hkershaw-brown, line 68: there is a relationship between these parameters that needs to remain the same if MAX_STATE_VARIABLES is increased. 

You can't code this, since they are parameters, so note in comment

MAX_STATE_VARIABLE_FIELDS = MAX_STATE_VARIABLES*3
MAX_STATE_VARIABLE_FIELDS_CLAMP = MAX_STATE_VARIABLES*5

hkershaw-brown: added comment in d43f7ee13b583300340b0d86de863063d139c292

models/utilities/default_model_mod.f90
hkershaw-brown: @mjs2369 do you think an error E_ERR message here?
That way if people have >= 100 state variables this is caught here and they can increase MAX_STATE_VARIABLES as needed?

mjs2369: @hkershaw-brown yes definitely, good catch

hkershaw-brown: sounds good. I'll put the error message in and we can do the release tomorrow!

developer_tests/namelist/work/quickbuild.sh
```diff
@@ -0,0 +1,43 @@
...
+export DART=$(git rev-parse --show-toplevel)
+source "$DART"/build_templates/buildfunctions.sh
+
+MODEL="none"
+EXTRA="$DART"/models/template/threed_model_mod.f90
+dev_test=1
+TEST="namelist"
+LOCATION="threed_sphere"
+
+programs=(
```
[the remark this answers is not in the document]
hkershaw-brown: fixed in a95013b1d8098c104c8d2fd3f6b1ff167cdc3975

models/utilities/default_model_mod.f90
hkershaw-brown: this actually leaves state_vars%nvars as unset

hkershaw-brown: so you don't get to this point if you have > 100 entries in the namelist. You get an invalid namelist entry:

message:  INVALID NAMELIST ENTRY:                       'TEMP_CUR33 ', 'QTY_POTENTIAL_TEMPERATURE', 'NO_COPY_BACK', in namelist model_nml

models/utilities/default_model_mod.f90
```diff
@@ -49,8 +51,25 @@ module default_model_mod
...
-
+          write_model_time,       &
+          parse_variables,        &
+          parse_variables_clamp,  &
+          state_var_type,         &
+          MAX_STATE_VARIABLE_FIELDS, &
+          MAX_STATE_VARIABLE_FIELDS_CLAMP
+
+type :: state_var_type
+    integer                        :: nvars
```
hkershaw-brown, line 62: note for chat today 
`nvars = -1`

mjs2369:
> @mjs2369 can you point me to where your test restart files are on Derecho, thanks!

@hkershaw-brown 
Sorry, just seeing this particular comment:

/glade/work/masmith/test_cases/

hkershaw-brown (approved):
Great work on this Marlee, 

I added the error check for nvars vs. MAX_STATE_VARIABLES in parse_variables, and parse_variables_clamp, if you can double check those that would be great. 

Feel free to release when you're ready!

Cheers,
Helen

Edit: parse_variables -> parse_variables_clamp

Issue #749, opened 2024-10-03
bug: inflation creating out-of-bounds values qceff


hkershaw-brown:
:bug: Your bug may already be reported!
Please search on the [issue tracker](https://github.com/NCAR/DART/issues) before creating a new issue.

### Describe the bug

1. List the steps someone needs to take to reproduce the bug.  
Run CAM-FV reanalysis single run of filter example /glade/derecho/scratch/hkershaw/DART/CAM-out-of-bounds/Rean_run 
inf_flavor 2

2. What was the expected outcome?
Run to completion

3. What actually happened?  

filter errors out:

```
ERROR FROM:
  source : bnrh_distribution_mod.f90
  routine: bnrh_cdf
  message:  Smallest ensemble member less than lower bound -4.009123520117420E-164  0.000000000000000E+000
```
### Error Message  
Please provide any error messages.

### Which model(s) are you working with?
CAM-FV 

### Screenshots   

Here are the lines of code:
https://github.com/NCAR/DART/blob/464aa57a1261fa098e5f0f999bad41a464f8d7db/assimilation_code/modules/assimilation/adaptive_inflate_mod.f90#L559-L560

### Version of DART      
Which version of DART are you using? 
v11.8.1

### Have you modified the DART code?   
No  

Here is a small reproducer

```
program test_maths

! Using selected_real_kind for double precision
integer, parameter :: dp = selected_real_kind(15, 307)
real(kind=dp) :: ens
real(kind=dp) :: mean
real(kind=dp) :: inflate  
real(kind=dp) :: sd_inflate

ens = 3.2342393548165711e-191_dp
inflate = 1.4082289753147326_dp
mean = 2.1474965713154784e-163_dp

print *, ens    

sd_inflate = sqrt(inflate) 
ens = ens * sd_inflate + mean * (1.0_dp - sd_inflate)

print *, ens    

end program test_maths  
```

```
hkershaw@derecho3:/glade/derecho/scratch/hkershaw/DART/Bugs/qceff-maths$  ifort fortran_maths.f90 
hkershaw@derecho3:/glade/derecho/scratch/hkershaw/DART/Bugs/qceff-maths$ ./a.out 
  3.234239354816571E-191
 -4.009123520117420E-164
```

Or you can run CAM-FV filter
/glade/derecho/scratch/hkershaw/DART/CAM-out-of-bounds/Rean_run

The numbers for the test_maths.f90 program are from proc3 (out of 128) j==61138

### Build information
Please describe:  
[...]

PR #794, opened 2025-01-09
Qceff inflation fix; Explicitly handling BNRHF transform failures.

jlaucar:
## Description:
The BNRHF transform fails when the computed ensemble standard deviation is not positive. Existing code returns an untransformed ensemble but no error return. The transforms and the probit transforms that call them were modified to return an error. The inflation in filter_mod and the increments in assim_tools_mod were updated to check for this error. These routines now explicitly do nothing to variables that fail in the transform step. 

### Fixes issue
fixes https://github.com/NCAR/DART/issues/749

### Types of changes

- [ ] Documentation update

### Tests
Was tested with Lorenz_96_tracer model QCEFF test suites. Produced answer changes in 7 of the tests but rest were bitwise. The answer changes do not appear to be associated with significantly changed perfomance.
Was also tested with the original CAM reanalysis case that led to identification of this problem. Tests there bitwise reproduced over several days of testing with previous results that had to use clamping; clamping is no longer required with the BNRHF as would be expected.

## Checklist for merging

- [x] Updated changelog entry
- [ ] Documentation updated
- [x] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed


assimilation_code/modules/assimilation/probit_transform_mod.f90
```diff
@@ -334,38 +349,43 @@ subroutine to_probit_bounded_normal_rh(ens_size, state_ens, p, probit_ens, &
...
    ! Get all the info about the rank histogram cdf
    call bnrh_cdf_params(state_ens, ens_size, bounded_below, bounded_above, &
       lower_bound, upper_bound, p, quantile)
 
-   ! Do not know what to do if sd is 0 (or small, work on this later)
+   ! Fail if sd is not positive (or small, work on this later)
    if(get_bnrh_sd(p) <= 0.0_r8) then
+      ierr = 1
       ! Just return the original ensemble
       probit_ens = state_ens
```
hkershaw-brown, line 386: this is quite a nice contract, that if the transform fails, just return the original ensemble.

assimilation_code/modules/assimilation/assim_tools_mod.f90
```diff
@@ -541,8 +545,9 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
       ! Convert all my obs (extended state) variables to appropriate probit space
       call transform_to_probit(ens_size, obs_ens_handle%copies(1:ens_size, i), dist_for_obs, &
          obs_dist_params(i), probit_ens, .false., &
-         bounded_below, bounded_above, lower_bound, upper_bound)
-      obs_ens_handle%copies(1:ens_size, i) = probit_ens
+         bounded_below, bounded_above, lower_bound, upper_bound, ierr)
+      obs_probit_trans_ok(i) = (ierr == 0)
+      if(obs_probit_trans_ok(i)) obs_ens_handle%copies(1:ens_size, i) = probit_ens
```
hkershaw-brown, line 550: The added  if(obs_probit_trans_ok(i)) & at line 511 gives you the same answer as before, because of the bnrh 

fail -> `! Just return the original ensemble`
I guess this is for algorithm readability?

assimilation_code/modules/assimilation/assim_tools_mod.f90
```diff
@@ -635,9 +640,11 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
          orig_obs_prior_var  = obs_ens_handle%copies(OBS_PRIOR_VAR_START:  &
             OBS_PRIOR_VAR_END, owners_index)
 
-         ! If QC is okay, convert this observation ensemble from probit to regular space
-         call transform_from_probit(ens_size, obs_ens_handle%copies(1:ens_size, owners_index) , &
-            obs_dist_params(owners_index), obs_ens_handle%copies(1:ens_size, owners_index))
+         ! Convert this observation ensemble from probit back to regular space if to_probit was successful
+         if(obs_probit_trans_ok(owners_index)) then 
+            call transform_from_probit(ens_size, obs_ens_handle%copies(1:ens_size, owners_index) , &
+               obs_dist_params(owners_index), obs_ens_handle%copies(1:ens_size, owners_index))
```
hkershaw-brown, line 646: this is still dangerous since you have the same input for intent(in) and intent(out), but the original code does that so yippee ki yay I guess.

https://github.com/NCAR/DART/blob/f0e7f99138df52600ef36caf0af1187ba81da83c/assimilation_code/modules/assimilation/probit_transform_mod.f90#L596-L601

assimilation_code/modules/assimilation/assim_tools_mod.f90
```diff
@@ -698,10 +705,14 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
       ! Convert the prior and posterior for this observation to probit space
       call transform_to_probit(grp_size, obs_prior(grp_bot:grp_top), dist_for_obs, &
          temp_dist_params, probit_obs_prior(grp_bot:grp_top), .false., &
-         bounded_below, bounded_above, lower_bound, upper_bound)
+         bounded_below, bounded_above, lower_bound, upper_bound, ierr)
+      ! If probit transform fails for any group, this observation cannot be assimimlated
+      if(ierr /= 0) cycle SEQUENTIAL_OBS
```
mgharamti, line 710: I think it could be helpful to add a CAUTION comment here. 
Because of the new error check, the interaction between the transformations and the groups can become messy. If it happens and the probit transform fails, then only part of the ensemble may be updated for a certain obs. Because of the `cycle SEQUENTIAL_OBS` the remaining groups, which could be transformed without issues, are disregarded.

mgharamti (commented):
All looks good to me. I just have one comment regrading the interaction between QCEFF and groups with the newly added error check.

assimilation_code/modules/assimilation/assim_tools_mod.f90
```diff
@@ -635,9 +640,11 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
          orig_obs_prior_var  = obs_ens_handle%copies(OBS_PRIOR_VAR_START:  &
             OBS_PRIOR_VAR_END, owners_index)
 
-         ! If QC is okay, convert this observation ensemble from probit to regular space
-         call transform_from_probit(ens_size, obs_ens_handle%copies(1:ens_size, owners_index) , &
-            obs_dist_params(owners_index), obs_ens_handle%copies(1:ens_size, owners_index))
+         ! Convert this observation ensemble from probit back to regular space if to_probit was successful
+         if(obs_probit_trans_ok(owners_index)) then
```
hkershaw-brown, line 644: If the to_probit was not ok, do you want to set the qc to failed fwd op?

The other processes don't have access to obs_probit_trans_ok(owners_index) and so they will go ahead and calculate obs_increments because the qc is 0, then transform_to_probit which may or may not work?

assimilation_code/modules/assimilation/assim_tools_mod.f90
hkershaw-brown: since you return the original if the transform fails, might be a good time to change this to inout

hkershaw-brown (changes requested):
A few comments, summary here actual comments are on the code.

Question about the error code.

Performance question on setting the qc if the observation ensemble.

Definite change is to remove the unused routine `transform_all_to_probit`, because it is unused and i think incorrect.

The intent(in) intent(out) I think will bite us in the future, but I think fixing it in this pull request is going to confuse what this pull request is fixing.

assimilation_code/modules/assimilation/assim_tools_mod.f90
nancycollins: i see that calling a routine with the same array with both an intent(in) and intent(out) goes several levels deep in this code so it would be pretty disruptive to change it.  but in the future i'd avoid passing in the same array multiple times with different intents.  it's leaving a trap for future code maintainers because someone looking only at the called subroutines could decide to set the outgoing state_ens(:) to 0 or MISSING_R8 as an initial value for debugging, which would result in the probit_ens(:) mysteriously getting overwritten without any protection from the intent(in).  you're requiring some behavior in the called subroutine because of how the calling code is written.

hkershaw-brown: I don't think it is standard conforming to alias in this way so not just people changing the code, its undefined what the compiler will do.

hkershaw-brown:
QCEFF lab request pull request being finished before anyone looks at this.

hkershaw-brown:
From @jlaucar at standup, this is not getting worked on until the DART_LAB & DART tutorial updates, so closing for now.

assimilation_code/modules/assimilation/probit_transform_mod.f90
```diff
@@ -62,7 +62,7 @@ module probit_transform_mod
 !------------------------------------------------------------------------
 
 subroutine transform_all_to_probit(ens_size, num_vars, state_ens, distribution_type, &
-   p, probit_ens, use_input_p, bounded_below, bounded_above, lower_bound, upper_bound)
+   p, probit_ens, use_input_p, bounded_below, bounded_above, lower_bound, upper_bound, ierr)
```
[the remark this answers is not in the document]
hkershaw-brown: removed in [894c9f2](https://github.com/NCAR/DART/commit/894c9f2d3ad9e217b8d2ac355f8823f824b756d7)
[built everything](https://github.com/NCAR/DART/actions/runs/12992529739) (caveat #805)

hkershaw-brown (approved):
Removed the unused routine, put in the error handler calls rather than 'stop'
Converted the unaddressed comments into issues.  #816 #817 #818 #819

PR #838, opened 2025-03-12
Check for THM USE_THETA_M attribute

### LINKED ISSUE #837: Check for WRF THM variable  USE_THETA_M setting
### What's the issue? 
In WRF-DART, a user can operate on the incorrect perturbed *moist* adiabatic potential temperature (instead of perturbed adiabatic potential temperature) if USE_THETA_M is not 0.  This is well documented, but sometimes not readily checked if user is relying on older simulations of WRF. 

### Where did you find the issue?
WRF-DART OSSE using WRF 4.1.2

### What needs to be fixed?
The DART code already makes THM in &model_nml mandatory, however doesn't check wrfinput_d01 attributes to confirm USE_THETA_M =0. 

As a side effect the use of THM becomes mandatory when using perfect_model_obs, which technically shouldn't be a requirement.....

### Suggestions for improvement
Perhaps error out if USE_THETA_M is not equal to zero in wrfinput.   Do nothing is probably also a good option too.  Placeholder until I get more feedback.

braczka:
## Description:
Mandate in WRF-DART that if THM variable is present in &model_nml it must be ' perturbed dry adiabatic potential temperature'.   This is set through USE_THETA_M option in WRF &dynamics namelist.

### Fixes issue
Fixes #837 

### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

None required.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,11 @@ individual files.
 
 The changes are now listed with the most recent at the top.
 
+**March 14 2025 :: WRF use_theta_m check. Tag v11.10.5**
+
+- Assert USE_THETA_M=0 in WRF when initializing wrf model_mod.
+- fixed documentation broken links.
+
 **February 6 2025 :: WRF tutorial and prepbufr documentation. Tag v11.10.4**
 
 - Observation section of WRF tutorial updated to match provided obs.   
--- a/conf.py
+++ b/conf.py
@@ -21,7 +21,7 @@ copyright = '2023, University Corporation for Atmospheric Research'
 author = 'Data Assimilation Research Section'
 
 # The full version, including alpha/beta/rc tags
-release = '11.10.4'
+release = '11.10.5'
 root_doc = 'index'
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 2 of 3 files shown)


braczka:
Tested this code using USE_THETA_M = 0 and USE_THETA_M = 1 and gives the expected behavior.   /glade/work/bmraczka/DART/models/wrf/MattW_bug/perfect_model_obs

Current WRF 4 and older WRF 3 wrfinput_d01 versions all contain USE_THETA_M as a global attribute, thus no need to set default value behavior in this DART issue.  @hkershaw-brown please give this a check, but I think it is ready to go.  Please include in the next release, as this problem is not really detectable in practice, and not documented well within WRF unless we mandate exact configuration of THM here within DART.

hkershaw-brown (approved):

PR #848, opened 2025-03-28
inverse distance weighting interpolation for CICE

hkershaw-brown:
## Description:

Switched out bilinear interpolation for inverse distance weighting interpolation in CICE model_mod.f90
Power for IDW is fixed at 2.
The inverse distance weight keeps the interpolation result with the range of the 4 corners of the quad. 

### Fixes issue

fixes #833 

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

### Documentation changes needed?

- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.

### Tests
Please describe any tests you ran to verify your changes.

Ran the CICE egregious bounds case.
I know this is not sufficient to test the interpolation, but getting the ball rolling with this pull request. 

## Checklist for merging

- [x] Updated changelog entry
- [ ] Documentation updated
- [x] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Create release from the main branch with appropriate tag
- [ ] Delete feature-branch

## Testing Datasets

- [ ] Dataset needed for testing available upon request
- [ ] Dataset download instructions included
- [ ] No dataset needed

Issue #833, opened 2025-03-05
bug: CICE get_expected_agreg_over_grid giving > 100%


hkershaw-brown:
:bug: 🧊 

User reported

### Describe the bug

1. Run cice with QCEFF with bounds on sea ice concentration 
2. What was the expected outcome? Run to completion, bounds respected
3. What actually happened?  CICE fwd op returning out of bounds causing QCEFF filter to die with "Largest ensemble member greater than upper bound"

### Error Message  
"Largest ensemble member greater than upper bound" 

### Which model(s) are you working with?
CESM CICE5(6?)

### Version of DART      
Which version of DART are you using? 
You can find the version using `git describe --tags`  
Will reproduce, believe it is v11

### Have you modified the DART code?   
No  

Work around is forcing the fwd op to be <100%

### Build information
Please describe:  
 1. Derecho
 2. unknown, will confirm if it makes a difference.

hkershaw-brown:
See also #819 #681

hkershaw-brown:
restarts: /glade/derecho/scratch/hkershaw/DART/Tickets/CICE

hkershaw-brown:
```
/glade/derecho/scratch/hkershaw/DART/Tickets/CICE/DART/models/cice/cice_to_dart.f90(124): warning #6843: A dummy argument with an explicit INTENT(OUT) declaration is not given an explicit value.   [TABLE]
subroutine verify_parameters(parameters, ngood,table)
-----------------------------------------------^
```

hkershaw-brown:
reproducer in /glade/derecho/scratch/hkershaw/DART/Tickets/CICE/run

fix_bound_violations        = .true.,

```
Before observation assimilation TIME: 2025/03/06 13:40:17
 ERROR FROM:
  source : probit_transform_mod.f90
  routine: fix_bounds
  message:  Egregious upper bound violoation first check(see code)   1.01133114890742        1.00000000000000
 
MPICH ERROR [Rank 55] [job id bf1d451c-e89f-41b3-adf1-710fecc3105e] [Thu Mar  6 13:40:17 2025] [dec0222] - Abort(99) (rank 55 in comm 496): application called MPI_Abort(comm=0x84000001, 99) - process 55
```

fix_bound_violations        = .false.,

```
Before observation assimilation TIME: 2025/03/06 13:47:20
 ERROR FROM:
  source : bnrh_distribution_mod.f90
  routine: bnrh_cdf
[...]

PR #887, opened 2025-06-03
Replace quad_bilinear_interp with inverse distance weighting in quad_utils. 

### LINKED ISSUE #453: quad_util_mod rotations
Reported in the standup that the quad_utils_mod may have problems when the quad is rotated at 45 degrees. 

model_mods that would be affected by this:
* cam-fv 
* mom6 (unreleased)
* quad_utils is used by people with their own DART development for TIEGCM

Not confirmed (yet).

There is no record of the error (or fix) in the log:
https://github.com/NCAR/DART/commits/main/models/utilities/quad_utils_mod.f90

There is the namelist option do_rotate (default .false.)
https://github.com/NCAR/DART/blob/16595cad595b6349605c9bc8e021c7ecc2c2391a/models/utilities/quad_utils_mod.f90#L2179

The notes talk about the degenerate case where the grid is rotated **90** degrees:
https://github.com/NCAR/DART/blob/16595cad595b6349605c9bc8e021c7ecc2c2391a/models/utilities/quad_utils_mod.f90#L2201-L2202

Mentioned at the standup (note I might not be recalling this correctly): cam-se has a fix for this rotation problem.  Is this it?
https://github.com/NCAR/DART/blob/16595cad595b6349605c9bc8e021c7ecc2c2391a/models/cam-se/model_mod.f90#L2531

Q. Why does cam-se not use quad_utils_mod?
Q. What is the goal of rotating?
Q. Why is the default `do_rotate = .false.`?

I think quad_interp_mod is a generalization (and replacement?) for the original POP model_mod interpolation code:
https://github.com/NCAR/DART/blob/16595cad595b6349605c9bc8e021c7ecc2c2391a/models/utilities/quad_utils_mod.f90#L894-L897
Q. Can we switch out the POP model_mod code?

There are 21 fixmes is the quad_util_mod module.

# Other quad_util_mod issues:
* see https://github.com/NCAR/DART/issues/249 for a note only only one of the quad_utils routines checking for missing values in the state.

* Note from pull #247
  ```
  quad_utils_mod.f90 uses  invals(4, nitems)
  cam-fv/model_mod.f90 uses quad_vals(ens_size,4) !< array of interpolated values
  ```
  which seems ripe for bugs such as https://github.com/NCAR/DART/issues/246. I'm not sure why the change 
  in order was made, maybe should be reverted.

### LINKED ISSUE #860: quad_utils_mod interpolation method
https://github.com/NCAR/DART/pull/848 replaced bilinear interpolation with IDW for CICE. 
The quad_utils_mod uses the bilinear interpolation method. 

See #453 for discussion of rotations of the quad - does not fix out-of-bounds 
See #833 for the reported out-of-bounds values for CICE

hkershaw-brown:
## Description:

The current quad_utils_mod interpolation method (bilinear) can give out-of-bounds values for some quads. 
This problem was originally reported by Nuo Chen in CICE #833 when using the QCEFF.  
The CICE bilinear interpolation was replaced by inverse distance weighting in #848.  This is the same code as quad utils. Note, replacing the CICE interpolation with quad utils is not done in this pull request. 

There is a notebook to call dart from python in https://github.com/hkershaw-brown/DART-probe/blob/main/ice-interp.ipynb if you want to reproduce the interpolation plots (see [README](https://github.com/hkershaw-brown/DART-probe/blob/main/README.md) for dart library build).  

### Fixes issue

Fixes #860
Fixes #453 since no rotation needed in the IDW. See #833 for plots of rotations

### Types of changes

- [ ] Documentation update

There is no documentation for quad utils at the moment. 

### Tests
This is a draft at the moment.
Questions to answer: 
bitwise differences for which models - who cares, it is a bug fix vs. loads of people care. 

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/build_templates/buildconvfunctions.sh
+++ b/build_templates/buildconvfunctions.sh
@@ -109,7 +109,7 @@ local modelsrc="$DART/models/template/model_mod.f90"
 local loc="$DART/assimilation_code/location/$LOCATION \
           $DART/assimilation_code/location/utilities/ \
           $DART/models/model_mod_tools/test_interpolate_$LOCATION.f90"
-local misc="$DART/models/utilities/ \
+local misc="$DART/models/utilities/default_model_mod.f90 \
             $DART/models/model_mod_tools/model_check_utilities_mod.f90 \
             $DART/observations/forward_operators/obs_def_mod.f90 \
             $DART//observations/forward_operators/obs_def_utilities_mod.f90 \
@@ -143,7 +143,12 @@ else #nompi
    core=${core//$winf08/}
 fi
 
-convsrc="${core} ${conv} ${obserrsrc} ${modelsrc} ${misc} ${loc}"
+quad=""
+if [ "$LOCATION" == "threed_sphere" ]; then
+   quad="$DART/models/utilities/quad_utils_mod.f90"
+fi
+
+convsrc="${core} ${conv} ${obserrsrc} ${modelsrc} ${misc} ${loc} ${quad}"
 
 # remove nuisance files
 nuisance=(\
--- a/build_templates/buildfunctions.sh
+++ b/build_templates/buildfunctions.sh
@@ -149,7 +149,7 @@ local loc="$DART/assimilation_code/location/$LOCATION \
           $DART/assimilation_code/location/utilities/ \
           $DART/models/model_mod_tools/test_interpolate_$LOCATION.f90"
 local modelsrc=$(find $DART/models/$MODEL -type d -name $EXCLUDE -prune -o -type f -name "*.f90" -print)
-local misc="$DART/models/utilities/ \
+local misc="$DART/models/utilities/default_model_mod.f90 \
             $DART/models/model_mod_tools/model_check_utilities_mod.f90 \
             $DART/observations/forward_operators/obs_def_mod.f90 \
             $DART//observations/forward_operators/obs_def_utilities_mod.f90 \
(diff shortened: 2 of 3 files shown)


hkershaw-brown:
build everything action: https://github.com/NCAR/DART/actions/runs/15426138267 caveat GEOS (Helen are you ever going to fix this? no).

jlaucar (commented):
This looks correct to me. It will definitely not bitwise reproduce. It remains unclear how often the result could be out of the range of the corner values due to roundoff. It we run into applications where this occurs, we may have to remove the error failure so that the currently unused bounds enforcement code comes into play.

jlaucar (changes requested):
This code still looks good to me. However, the comment stating that it is 'unclear' whether round-off could lead to the interpolated value being outside of the range of the vertex values is now clear. The modified code I'm using in aether_cube_sphere occasionally has an interpolated value outside or range when all of the vertex values are identical to machine precision. This comment should probably be changed to
"Round-off can lead to result being outside of range of grid points"

The comment further down should also be changed since we are issuing a MSG rather than an ERR. The part that says "this will not happen with current error check" should be deleted.

hkershaw-brown:
also check pole for IDW in addition to #978

hkershaw-brown:
Note Moha working on improved quad algorithms.

hkershaw-brown:
closing this pull request as it is now, pull #1089 has a more stable method than Cramer's rule method. 
If idw goes in to quad_utils_interp, go in as an optional interpolation method.

PR #892, opened 2025-06-03
Fix ITP & improve performance

iangrooms:
## Description:
This PR does two main things. First, the ITP algorithm in the rootfinding mod is fixed. The original version was based on Eq (15) from their paper, which was a typo. The buggy version still converged but was slow.

Second, several changes have been made to improve performance.
- In the rootfinding mod, for bounded cases we don't immediately go to ITP, which was inefficient.
- In the kde distribution mod the quadrature has been switched from ninth-order to fifth order. This could degrade accuracy for small ensembles, but significantly improves performance (nearly 40% speedup, since most of the time is spent in this function).
- In the kde distribution mod the logic associated with the boundary correction has been updated to reduce unnecessary computation. This doesn't change answers but significantly improves performance.

### Types of changes
- [ ] Documentation update
Performance improvement isn't on this list, but that's a big part of this PR.

(No doc changes needed)
### Tests
I ran the kde distribution test, which passes with both the rootfinding mod version of inv_cdf and the normal distribution mod version of inv_cdf. I tested performance by just looping that test 4000 times.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed

### DIFF
--- a/assimilation_code/modules/assimilation/kde_distribution_mod.f90
+++ b/assimilation_code/modules/assimilation/kde_distribution_mod.f90
@@ -256,12 +256,12 @@ function kde_pdf(x, p)
       u_lower = 0._r8; lx_lower = 1._r8 ; mx_lower = 0._r8
       u_upper = 0._r8; lx_upper = 1._r8 ; mx_upper = 0._r8
       if (p%bounded_below) then ! Bounded below
-         u_lower = min( 1._r8, max( 0._r8, (x - p%lower_bound) / p%more_params(i) ) ) ! p%more_params(i) holds kernel width for ensemble member i
-         call boundary_correction(u_lower, lx_lower, mx_lower)
+         u_lower = (x - p%lower_bound) / p%more_params(i) ! p%more_params(i) holds kernel width for ensemble member i
+         if ((u_lower >= 0._r8) .and. (u_lower <= 1._r8)) call boundary_correction(u_lower, lx_lower, mx_lower)
       end if
       if (p%bounded_above) then ! Bounded above
-         u_upper = min( 1._r8, max( 0._r8, (p%upper_bound - x) / p%more_params(i) ) )
-         call boundary_correction(u_upper, lx_upper, mx_upper)
+         u_upper = (p%upper_bound - x) / p%more_params(i)
+         if ((u_upper >= 0._r8) .and. (u_upper <= 1._r8)) call boundary_correction(u_upper, lx_upper, mx_upper)
       end if
       u_lower = (x - p%ens(i)) / p%more_params(i) ! Not u_lower any more, just (x-x_i)/h_i
       kde_pdf = kde_pdf + (1._r8 / p%more_params(i)) * &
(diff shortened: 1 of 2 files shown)


iangrooms:
Right now I've switched the quadrature from fifth to third order, but left the fifth order code in there, inactive. It would be nice to control this using a namelist parameter, but I'm not sure how to do that for a module. Is there a way to read a namelist variable directly into this module? If so, then I can set up the logic to use third or fifth order based on that variable.

hkershaw-brown:
For a namelist in a module, you'll need to add a check for the namelist being read for any routine that needs the namlist option. DART follows this pattern, where there is an initializing routine (e.g.  "rootfinding_init") called to read the namelist on the first call.   Pseudo code:

```
! module storage to indicate module has been initialized or not
logical :: module_initialized = .false.

namelist / rootfinding_nml / your_option_name

---
subroutine rootfinding_init

module_initialized = .true.

! Read the namelist entry
call find_namelist_in_file("input.nml", "rootfinding_nml", iunit) ! <- this will kill the program if rootfinding_nml is not in the input.nml
read(iunit, nml = probit_transform_nml, iostat = io)
call check_namelist_read(iunit, io, "rootfinding_nml")

! These log the namelist values
if (do_nml_file()) write(nmlfileunit,nml=rootfiniding_nml)
if (do_nml_term()) write(     *     ,nml=rootfinding_nml)


end subroutine  rootfinding_init
---

---
function integrate_pdf(x, p) result(q)

! If not initialized, read in the namelist - for routines that depend on the namelist
if(.not. module_initialized) call initialize_rootfiniding

! use "your_option_name" however you need in the routine
...
---


```

You can take a look at https://github.com/NCAR/DART/blob/d9082222ce142a81d30e0cf3f0f014af1c08db14/assimilation_code/modules/assimilation/probit_transform_mod.f90#L57 or your favorite module for an example.

iangrooms:
I'm working on using a namelist to set the quadrature order, and I think it's working. Is there a way to set it up so that if `rootfinding_nml` is not present in `input.nml` it will just use default values, rather than killing the program? Or maybe there's a reason to prefer for it to kill the program?

hkershaw-brown:
> I'm working on using a namelist to set the quadrature order, and I think it's working. Is there a way to set it up so that if `rootfinding_nml` is not present in `input.nml` it will just use default values, rather than killing the program? Or maybe there's a reason to prefer for it to kill the program?

My opinion is that optional namelists are a good idea, particularly for "advanced" user features or even experimental features, because otherwise all input.nmls have to have every namelist.

The change would be here, e.g. having an optional argument 
find_namelist_in_file(.... missing_fine)
if missing_is_fine skip this error
https://github.com/NCAR/DART/blob/d9082222ce142a81d30e0cf3f0f014af1c08db14/assimilation_code/modules/utilities/utilities_mod.f90#L425-L430

Do you want to be the first optional namelist in dart?

iangrooms:
I tried skipping the error message, but then it just hangs. The subroutine returns `iunit` and if I just skip the error then I'm guessing it returns a value for `iunit` that is not useful to the calling code. Maybe I should just leave it as a required namelist. I guess one benefit of making it required is that users will know about these options

hkershaw-brown:
@iangrooms can you update this pull request to NCAR main when you get a chance, I can not update it from git since your main branch is protected:
Screenshot 2025-08-28 at 12 42 25 PM
thanks!

PR #906, opened 2025-06-26
feat: MOM6 option to use pseudo depth rather than thickness for vertical location

### LINKED ISSUE #903: Feature request: MOM6 z* for depth
### Use case  
MOM6 perfect model where only z* (Layer pseudo-depth) is available

Layer thickness not available in some runs

Option to use z* rather than thickness for vertical location. 

Originally MOM6 model_mod was written with layer pseudo-depth for the vertical, but this was switched out to use layer thickness which varies by ensemble member.

https://github.com/NCAR/DART/blob/d93a56f5a464253e7b268df8fd99a66e788a42a6/models/MOM6/model_mod.f90#L807-L816

Note only layer pseudo depth is available (not interface) for this case.

hkershaw-brown:
## Description:

MOM6 option to use pseudo depth rather than thickness for vertical location
Name of layer pseduo depth variable as a namelist option Users may have set 'Layer' to something else, e.g. 'zl'

Use case is existing long (10 years ish) runs of MOM6 that have only pseudo depth available. 

### Fixes issue

fixes #903

### Types of changes

- [ ] Documentation update

### Tests
Not enough

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/models/MOM6/model_mod.f90
+++ b/models/MOM6/model_mod.f90
@@ -103,6 +103,9 @@ type(quad_interp_handle) :: interp_t_grid,  &
                             interp_u_grid,  &
                             interp_v_grid
 
+! Ocean vertical
+real(r8), allocatable :: zstar(:) ! pseudo depth for each layer
+
 ! Ocean vs land
 real(r8), allocatable :: wet(:,:), basin_depth(:,:)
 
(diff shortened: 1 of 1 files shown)


iangrooms (commented):
Looks good to me. One potential "gotcha" is that the pseudo-depth goes all the way to the deepest value even in places where the ocean depth is much shallower. For example, in a part of the domain where the total ocean depth is 10 m, the pseudo-depth will still go all the way down to 5750m (or whatever the deepest value is). So when using pseudo-depth, you'll only see the `OBS_TOO_DEEP` status when an observation is below the deepest pseudo-depth level, not when it's below the bottom of the ocean. I don't think this will break the code though. Total ocean depth (from 0, not from the actual sea surface, which doesn't have to be at 0) is defined in the ocean geometry file, if you want to avoid this gotcha.

hkershaw-brown:
> Looks good to me. One potential "gotcha" is that the pseudo-depth goes all the way to the deepest value even in places where the ocean depth is much shallower. For example, in a part of the domain where the total ocean depth is 10 m, the pseudo-depth will still go all the way down to 5750m (or whatever the deepest value is). So when using pseudo-depth, you'll only see the `OBS_TOO_DEEP` status when an observation is below the deepest pseudo-depth level, not when it's below the bottom of the ocean. I don't think this will break the code though. Total ocean depth (from 0, not from the actual sea surface, which doesn't have to be at 0) is defined in the ocean geometry file, if you want to avoid this gotcha.

Thanks Ian, the `on_basin_edge` check will stop the interpolation going further in the code as is, but you're right the depth should be checked with ocean_geometry before calling find_level_bounds, and get OBS_TOO_DEEP as a status.

mjs2369 (approved):

PR #911, opened 2025-07-08
Python packages - qceff table pretty printer

hkershaw-brown:
## Description:

The qceff table has a lot of info, which is hard to read.
This python package provides `display-qceff-table` which pretty prints the qceff table for qceff.csv files.

It is in a python package so people can install with:
pip install -e DART/pytools/pyqceff

and use `display-qceff-table`  without the full path.

Also allows packages to be put individually on [pypi](https://pypi.org/)

There is also pyDART.txt  which is list (one package in it currently) to install DART/pytools
pip install -r pyDART.txt

Example of pretty-print:
Screenshot 2025-07-07 at 8 59 57 PM

Usage:
```
(py-env) [hkershaw:pytools](python-packages) > display-qceff-table --help                                    
usage: display-qceff-table [-h] [--details [QTY]] csv_file

Display QCEFF table summary and details.

positional arguments:
  csv_file         Path to QCEFF table CSV file

options:
  -h, --help       show this help message and exit
  --details [QTY]  Print detailed information for all quantities, or for a specific quantity QTY
```

### Fixes issue

No issue

### Types of changes

- [ ] Documentation update

No docs, draft pull request to see what people think.

### Tests
There is a test directory, but this test is only checking the QTYs are in the printed output.

cd DART/pytools
pytest

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,19 @@ individual files.
 
 The changes are now listed with the most recent at the top.
 
+**July 24 2025 :: MOM6 pseudo depth, MODIS ice surface temp. Tag v11.12.0**
+
+New features:
+
+- MODIS ice surface temperature observation converter
+  *contributed by Molly Wieringa*
+- Option to use pseudo depth for vertical location in MOM6-DART
+- QCEFF table pretty printer: display-qceff-table
+
+Bug fix:
+
+  - perfect_model_obs progress print info 
+
 **July 16 2025 :: GitHub actions and MOM6 PMO defaults. Tag v11.11.3**
 
 - non-root user for containers
--- a/conf.py
+++ b/conf.py
@@ -21,7 +21,7 @@ copyright = '2023, University Corporation for Atmospheric Research'
 author = 'Data Assimilation Research Section'
 
 # The full version, including alpha/beta/rc tags
-release = '11.11.3'
+release = '11.12.0'
 root_doc = 'index'
 
 # -- General configuration ---------------------------------------------------


hkershaw-brown:
Note I am not sure on the gamma & beta  closed vs open [ vs ( have to check the .f90 code. Currently printing [ for closed.

pytools/pyqceff/src/pyqceff/display_qceff_table.py
```diff
@@ -0,0 +1,336 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""
+Script to read and pretty print QCEFF table information from a CSV file.
+"""
+
+import csv
+import sys
+from typing import Dict, List, Any
```
hkershaw-brown, line 9: python 3.8, could switch to 3.9+ but is python 3.8 still knocking about on some clusters.

mjs2369 (commented):
[https://github.com/NCAR/DART/blob/31d675a7ff24e23dce69874a1da17580696b6e95/pytools/pyqceff/src/pyqceff/display_qceff_table.py#L208](https://github.com/NCAR/DART/blob/31d675a7ff24e23dce69874a1da17580696b6e95/pytools/pyqceff/src/pyqceff/display_qceff_table.py#L208)

I'm a little unsure if we want to do this. I definitely get the idea here, but technically the names of the QTY are, for example, 'QTY_STATE_VARIABLE' instead of 'STATE_VARIABLE'

I noticed that when specifying a QTY to print detailed information for, you can do either the full name, i.e `/Users/masmith/Library/Python/3.13/bin/display-qceff-table tests/test.csv --details QTY_ONE` or just `/Users/masmith/Library/Python/3.13/bin/display-qceff-table tests/test.csv --details ONE`, which I think is a really nice touch

pytools/pyqceff/README.md
```diff
@@ -0,0 +1,32 @@
...
+where DART is the path to DART.
+
+> **Note:** If you update DART with `git pull`, using the editable install (`pip install -e DART/pytools/pyqceff`) ensures your installed package always matches the latest source code.
+
+
+## Usage
+
+After installation, you can use the `display_qceff_table.py` script to read and display QCEFF tables:
+
+```sh
```
mjs2369, line 22: **If we keep the install setup as is**, it would be good to add a note to the usage instructions to include the path to the display_qceff_table executable in the command below ( in my case, the executable was built at /Users/masmith/Library/Python/3.13/bin/display-qceff-table )

hkershaw-brown: Typically where python installs binaries should be in the path.  For the docs I think recommending people use a virtual environment (which takes care of PATH) is the way to go.  We don't really want people pip installing into their base python.

mjs2369 (approved):
Great start to this Python tools suite

PR #916, opened 2025-07-10
bug-fix: perfect_model_obs prints

### LINKED ISSUE #905: bug: perfect_model_obs prints out the local obs number for task 0 for num obs processed

### Describe the bug

1. run perfect_mode_obs with print_every_nth_obs = 1 with mpirun -n 128
2. print out 1 to n obs
3. What actually happened?  

    From the printout, it looks like only 391 of the obs have been processed.
    ```
    PE 0: move_ahead Next assimilation window contains up to    49953 observations
    ...
     PE 0: perfect_model_obs: Processing observation       390  of    49953
     PE 0: perfect_model_obs: Processing observation       391  of    49953
     PE 0:
     PE 0: perfect_model_obs: Main evaluation loop, starting iteration    1
     PE 0: p_m_o trace: Before move_ahead checks time of data and next obs
     PE 0: perfect_model_obs: No more obs to evaluate, exiting main loop
    ```

### Error Message  
No error, just missleading observation numbers

MOM6

### Version of DART      
v11.11.1-1-gca200df36 (mom6-z branch) but the problem is also on main

No, occurs on main.

### Build information
Please describe:  
 1. Derecho
 2. intel

### LINKED ISSUE #922: pmo trace printing problems.
> /glade/derecho/scratch/hkershaw/DART/MOM6/DART/models/MOM6/work(mom6-z)
prints trace message for first and last from for every task. 
> 
> ```
> PE 0: perfect_model_obs: Model does not need to run; data already at required t
>  ime
>  PE 0: p_m_o trace: Before setup for next group of observations
>  PE 0: p_m_o trace: Number of observations to be evaluated  49953
>  p_m_o trace: Time of first observation in window day=149748, sec=43275
>  p_m_o trace: Time of last  observation in window day=149749, sec=43200
>  p_m_o trace: Time of first observation in window day=149748, sec=43275
>  p_m_o trace: Time of last  observation in window day=149749, sec=43200
>  p_m_o trace: Time of first observation in window day=149748, sec=43275
>  p_m_o trace: Time of last  observation in window day=149749, sec=43200
>  p_m_o trace: Time of first observation in window day=149748, sec=43275
>  p_m_o trace: Time of last  observation in window day=149749, sec=43200
>  p_m_o trace: Time of first observation in window day=149748, sec=43275
>  p_m_o trace: Time of last  observation in window day=149749, sec=43200
>  p_m_o trace: Time of first observation in window day=149748, sec=43275
>  p_m_o trace: Time of last  observation in window day=149749, sec=43200
>  p_m_o trace: Time of first observation in window day=149748, sec=43275
>  p_m_o trace: Time of last  observation in window day=149749, sec=43200
>  p_m_o trace: Time of first observation in window day=149748, sec=43275
>  p_m_o trace: Time of last  observation in window day=149749, sec=43200
> ``` 

 _Originally posted by @hkershaw-brown in [#905](https://github.com/NCAR/DART/issues/905#issuecomment-3070064569)_

mjs2369:
## Description:

1) Adds a conditional to the print statements in pmo "Processing observation         X  of        Y" so that the error_handler is called with E_ALLMSG if pmo is being run on more than 1 mpi proc

This fixes the issue that was occurring where it was only printing the the obs on task 0, which made it appear that only those obs were being processed

2) Also adds a conditional to only call print_obs_time if on task 0, fixing the issue of it printing the same thing on every task

NOTE that now that we are not using trace_message, these prints will NO LONGER be executed if the namelist option silence is set to true

### Fixes issue
fixes #905 
fixes #922  

### Types of changes

- [ ] Documentation update

### Tests

Ran pmo for lorenz models and MOM6 with various numbers of MPI procs and values for print_every_nth_obs, ensured prints were correct

## Checklist for merging

- [ ] Updated changelog entry
- [x] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed

### DIFF
--- a/assimilation_code/programs/perfect_model_obs/perfect_model_obs.f90
+++ b/assimilation_code/programs/perfect_model_obs/perfect_model_obs.f90
@@ -449,8 +449,10 @@ AdvanceTime: do
    write(msgstring, '(A,I7)') 'Number of observations to be evaluated', &
       num_obs_in_set
    call trace_message(msgstring)
-   call print_obs_time(seq, key_bounds(1), 'Time of first observation in window')
-   call print_obs_time(seq, key_bounds(2), 'Time of last  observation in window')
+   if(my_task_id() == 0) then
+      call print_obs_time(seq, key_bounds(1), 'Time of first observation in window')
+      call print_obs_time(seq, key_bounds(2), 'Time of last  observation in window')
+   endif
 
    ! for multi-core runs, each core needs to store the forward operator and the qc value
    call init_ensemble_manager(fwd_op_ens_handle, ens_size, int(num_obs_in_set,i8), 1, transpose_type_in = 2)
--- a/assimilation_code/programs/perfect_model_obs/perfect_model_obs.f90
+++ b/assimilation_code/programs/perfect_model_obs/perfect_model_obs.f90
@@ -9,7 +9,7 @@ program perfect_model_obs
 use        types_mod,     only : r8, i8, metadatalength, MAX_NUM_DOMS
 use    utilities_mod,     only : error_handler, &
                                  find_namelist_in_file, check_namelist_read, &
-                                 E_ERR, E_MSG, E_DBG, nmlfileunit, timestamp, &
+                                 E_ERR, E_MSG, E_ALLMSG, E_DBG, nmlfileunit, timestamp, &
                                  do_nml_file, do_nml_term, logfileunit, &
                                  open_file, close_file
 use time_manager_mod,     only : time_type, get_time, set_time, operator(/=), print_time,   &
@@ -485,6 +485,8 @@ AdvanceTime: do
    ! Compute the forward observation operator for each observation in set
    do j = 1, fwd_op_ens_handle%my_num_vars
 
+      global_obs_num = fwd_op_ens_handle%my_vars(j)
+
       ! Some compilers do not like mod by 0, so test first.
       if (print_every_nth_obs > 0) nth_obs = mod(j, print_every_nth_obs)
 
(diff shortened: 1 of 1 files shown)


mjs2369:
A few extra notes:

1) The observations are out of order while being printed (especially when using lots of mpi procs) - currently the prints look like this with MOM6 on 256 procs

image

I could force the pmo prints to go in order, but I would have to sync across tasks, so I don't think we want to go for that...

2) The print statements are slightly mismatched in terms of how many spaces there are on task 0 vs all the other tasks. Here's an example with lorenz_63

image

We could fix this by adding the same number of spaces to line 780 as there are in 782 here in the error_handler subroutine

[https://github.com/NCAR/DART/blob/d9082222ce142a81d30e0cf3f0f014af1c08db14/assimilation_code/modules/utilities/utilities_mod.f90#L779-L782](https://github.com/NCAR/DART/blob/d9082222ce142a81d30e0cf3f0f014af1c08db14/assimilation_code/modules/utilities/utilities_mod.f90#L779-L782)

i.e.
```
   if (task_number == 0) then
      write(taskstr, '(a)' ) "PE     0: "
```

assimilation_code/programs/perfect_model_obs/perfect_model_obs.f90
```diff
@@ -485,22 +485,29 @@ subroutine perfect_main()
...
-         call trace_message(msgstring, 'perfect_model_obs:', -1)
+         if(task_count() == 1) then
+            write(msgstring, '(A,1x,I8,1x,A,I8)') 'Processing observation ', j, &
+                                         ' of ', num_obs_in_set
+            call trace_message(msgstring, 'perfect_model_obs:', -1)
+         else
+            write(msgstring, '(A,1x,I8,1x,A,I8)') 'Processing observation ', global_obs_num, &
+                                         ' of ', num_obs_in_set
+            call error_handler(E_ALLMSG, 'perfect_model_obs:' ,trim(msgstring))
+         endif
```
hkershaw-brown, line 505: This is neat.  

I was wondering if you could not have the if statement and use E_ALLMSG for both cases since j==global_obs_num for task_count()==1

But I have not looked through why trace_message is used rather than error_handler(E_MSG
Is the -1 to stop 'silence=.true.' from not printing the message? 

Side note, there is bunch of replicated code "trace_message", "set_trace" between pmo and filter and other modules.

hkershaw-brown:
The window printing is broken too:

https://github.com/NCAR/DART/issues/905#issuecomment-3070064569

assimilation_code/programs/perfect_model_obs/perfect_model_obs.f90
mjs2369: > I was wondering if you could not have the if statement and use E_ALLMSG for both cases since j==global_obs_num for task_count()==1
> 
> But I have not looked through why trace_message is used rather than error_handler(E_MSG Is the -1 to stop 'silence=.true.' from not printing the message?

That's a good point. I think it would be better to use E_ALLMSG in both cases. I'll make this change. 

I was also wondering about the general need for trace_message since we have the error_handler. It seems that trace_message and error_handler are basically doing the same thing in a slightly different way. I think we could consider switching out the places that use trace_message to use the error_handler alltogether.

But like you said, trace_message is replicated in pmo and filter_mod.f90 (and models/coamps_nest/coamps_util_mod.f90 but I don't think this is seeing a lot of use). So at the very least I think we should move trace_message to utilities_mod. Should I go ahead and make an issue about this and its overlap with the error_handler?

Finally, the -1 in `call trace_message(msgstring, 'perfect_model_obs:', -1)` is simply to force the message to print even if silence is set to true in the namelist

mjs2369:
Also fixes #922 by adding  a conditional to only call print_obs_time if on task 0

PR #989, opened 2025-10-29
ROMS_rutgers model_mod ens mean feature 

mgharamti:
## Description:

I added a feature to the ROMS-rutgers model-mod that allows easy and quick access to the ensemble mean through the input template file. There is a new logical namelist parameter that controls this. When the mean of the input ensemble is provided, the model-mod can use its SSH for conducting more accurate localization. This removes the need to implement vertical conversion routines. The PR also includes some cleanup of the model-mod and some of its unused code. 

### Types of changes

- [ ] Documentation update

### Tests
Ran ROMS-rutgers with an 80 member ensemble. 

## Checklist for merging

- [ ] Updated changelog entry
- [x] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets
The testing dataset can be accessed here: `/glade/derecho/scratch/gharamti/inacawo/DART/models/ROMS_rutgers/work`


mgharamti:
Hi Helen, 
Thanks for looking at this. To clarify a couple of things, there is a `compute_physical_depth` interface towards the top of the `model_mod` that switches between a scalar or array computation depending on the call. I could delete that interface and just use a scalar one, and loop over it when I need depths for the entire water column. I just thought having an interface is a cleaner way (like many other tools we have in DART). Note: the interface is not new to this PR. 

For `zeta`, there are 3 distinct cases:  
- it will come as an input argument if the interpolation routine is requesting depths,
- gsmd (or localization) is requesting depths, then it will check whether we have that information from the ensemble mean,
- not available, it will simply be set to 0 (which may not be that bad of an assumption). 

Happy to change the name of the new namelist parameter if it doesn't sound right. Can you clarify what you mean by terminology/docs?

hkershaw-brown (approved):
Nice stuff Moha!

PR #1009, opened 2025-11-25
ARVOR Floats and SVP Drifters Ocean Converters

mgharamti:
## Description:
This PR adds two new in-situ ocean converters (ARVOR profiling floats and SVP surface drifters). It also introduces a reusable CSV parsing utility in `parse_args_mod`. Both converters make use of this CSV interface, which simplifies code. In addition, documentation has been added to the converters. 

The CSV parsing utilities build on already existing parsing infrastructure (like a wrapper). The functionality mimics our netcdf handling in the sense that a file is opened, and data is accessed with a single call before closing the file. A few helper functions have been also added. These can be used to access the header, to inquire if a field exists, to find the dimensions, etc. 

### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

### Documentation changes needed?
- [x] My change requires a change to the documentation.
  - [x] I have updated the documentation accordingly.

### Tests
Tested both converters using actual raw ASCII data files. 

## Checklist for merging
- [ ] Updated changelog entry
- [x] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Create release from the main branch with appropriate tag
- [ ] Delete feature-branch

## Testing Datasets
ARVOR: `/glade/derecho/scratch/gharamti/inacawo/DART/observations/obs_converters/ARVOR/work/obs_files.txt`
SVP: `/glade/derecho/scratch/gharamti/inacawo/DART/observations/obs_converters/SVP/work/obs_files.txt`

PR #1023, opened 2025-12-12
change parse_args_test to be more automated.

nancycollins:
renamed the original test to interactive_parse_args_test.  it read standard input and prints out the values it gets.  there is still a parse_args_test.in text file that can be used as input, or typing from the terminal.

the new parse_args_test uses the fortran-testanything module which prints 'ok' for each test that passes.  this version of the test needs no input files.

## Description:
made the parse_args_test use the fortran-testanything routines and hardcoded the input inside the test so it can be more easily automated.

Edit Dec 29th 2025: 
This pull request also changes parse_args_mod to have ACHAR(92) rather than `'\'`

### Fixes issue
related to moha's https://github.com/NCAR/DART/pull/1009 pull request for the ARVOR obs converter.  if this test is considered better than the original, then i'll make the csv read test use the same structure.

### Types of changes

- [ ] Documentation update

### Tests
I ran the test program and it failed when I had bad values in the test cases and ran fine when I fixed them.

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed

### DIFF
--- a/developer_tests/utilities/parse_args_test.f90
+++ b/developer_tests/utilities/parse_args_test.f90
@@ -29,8 +29,11 @@ character(len=1), parameter :: BACKSLASH = ACHAR(92)
 ! main code here
  
 ! initialize the dart libs
-call initialize_utilities('parse_args_test')
+call initialize_utilities('parse_args_test', standalone_program=.true.)
 
+! this count should match the number of tests expected to be 
+! executed if the test program runs to completion.
+call plan(57)
 
 ! parse tests - divide the blank-separated words on a line
 ! into individual words.  note no interpretation of the
@@ -112,8 +115,8 @@ call get_args_from_string(nextline, nargs, args)
 
 call ok((nargs == 2),           trim(testnum)//', nargs')
 call ok((args(1) == 'quoted'),  trim(testnum)//', arg 1')
-call ok((args(2) == 'string'),  trim(testnum)//', arg 1')
-call ok((args(3) == ''),        trim(testnum)//', arg 2')
+call ok((args(2) == 'string'),  trim(testnum)//', arg 2')
+call ok((args(3) == ''),        trim(testnum)//', arg 3')
 
 
 testnum = 'test 8'
@@ -152,12 +155,12 @@ call ok((names(3) == 'c'),      trim(testnum)//', name 3')
 call ok((vals(3)  == '3'),      trim(testnum)//', val 3')
 call ok((names(4) == ''),       trim(testnum)//', name 4')
 call ok((vals(4)  == ''),       trim(testnum)//', val 4')
-call ok((.not. continue_line),  trim(testnum)//', continue_line')
+call ok((.not. continue_line),  trim(testnum)//', no continue_line')
 
 
-testnum = 'test 10'
+testnum = 'test 11'
 call resetvals()
-nextline = 'a = 1.0 b   =  2.0  c=3.0'
+nextline = 'a = 1.0 b   =  2.0  c=3.0  &'
 call get_name_val_pairs_from_string(nextline, nargs, names, vals, continue_line)
 
 call ok((nargs == 3),           trim(testnum)//', nargs')
@@ -169,7 +172,7 @@ call ok((names(3) == 'c'),      trim(testnum)//', name 3')
 call ok((vals(3)  == '3.0'),    trim(testnum)//', val 3')
 call ok((names(4) == ''),       trim(testnum)//', name 4')
 call ok((vals(4)  == ''),       trim(testnum)//', val 4')
-call ok((.not. continue_line),  trim(testnum)//', continue_line')
+call ok(continue_line,          trim(testnum)//', has continue_line')
 
 
 ! finalize parse_args_test
--- a/assimilation_code/modules/utilities/parse_args_mod.f90
+++ b/assimilation_code/modules/utilities/parse_args_mod.f90
@@ -57,6 +57,9 @@ public :: get_args_from_string, &
 
 character(len=*), parameter :: source = 'parse_args_mod.f90'
 
+! the ascii code for the backslash character is 92.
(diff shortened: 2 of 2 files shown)


developer_tests/utilities/interactive_parse_args_test.f90
```diff
@@ -0,0 +1,90 @@
...
+!
+
+! this test reads stdin.  to automate it, run it with:
+!  cat bob | ./parse_args_test
+! or
+!  ./parse_args_test < bob
+!
+! where the file bob contains one line per test with arguments to be parsed.
+
+program parse_args_test
```
hkershaw-brown: This program is not getting built, can it be removed?

developer_tests/utilities/parse_args_test.f90
```diff
@@ -1,90 +1,198 @@
...
 ! initialize the dart libs
 call initialize_utilities('parse_args_test')
 
-print *, 'enter lines with words separated by blanks.'
-print *, '(enter control-D to end)'
 
+! parse tests - divide the blank-separated words on a line
+! into individual words.  note no interpretation of the
+! contents is expected - the returns are always strings.
+
```
hkershaw-brown:
```suggestion
call plan(57)
```
Good to add the number of expected tests, that way if tests are skipped unintentionally this is caught, e.g. setting the plan(101) just to show the catch: 

```
./parse_args_test | ./tapview
.........................................................
Expected 101 tests but only 57 ran.
57 tests, 0 failures.
```

developer_tests/utilities/parse_args_test.f90
```diff
@@ -1,90 +1,198 @@
...
+! the ascii code for the backslash character is 92.
+character(len=1), parameter :: BACKSLASH = ACHAR(92)
+
 !----------------------------------------------------------------------
 
 ! main code here
  
 ! initialize the dart libs
 call initialize_utilities('parse_args_test')
```
hkershaw-brown:
```suggestion
call initialize_utilities('parse_args_test', standalone_program=.true.)
```

I don't think this test program needs to write namelists, logs so if possible it would be good to only output the test output to std out that way you can pipe it to something like tapview:

```
./parse_args_test | ./tapview
.........................................................
57 tests, 0 failures.
```

developer_tests/utilities/parse_args_test.f90
```diff
@@ -1,90 +1,198 @@
...
+call ok((names(2) == 'b'),      trim(testnum)//', name 2')
+call ok((vals(2)  == '2'),      trim(testnum)//', val 2')
+call ok((names(3) == 'c'),      trim(testnum)//', name 3')
+call ok((vals(3)  == '3'),      trim(testnum)//', val 3')
+call ok((names(4) == ''),       trim(testnum)//', name 4')
+call ok((vals(4)  == ''),       trim(testnum)//', val 4')
+call ok((.not. continue_line),  trim(testnum)//', continue_line')
+
+
+testnum = 'test 10'
```
hkershaw-brown: Should this be test 11 since it is a new nextline?  Or is it 10 because it is testing name/value pairs?
```suggestion
testnum = 'test 11'
```

developer_tests/utilities/parse_args_test.f90
```diff
@@ -1,90 +1,198 @@
...
+
+testnum = 'test 7'
+call resetvals()
+nextline = '"quoted" "string"'
+call get_args_from_string(nextline, nargs, args)
+
+call ok((nargs == 2),           trim(testnum)//', nargs')
+call ok((args(1) == 'quoted'),  trim(testnum)//', arg 1')
+call ok((args(2) == 'string'),  trim(testnum)//', arg 1')
+call ok((args(3) == ''),        trim(testnum)//', arg 2')
```
hkershaw-brown: It looks like the message does not match the test here, should this be 2,3?
```suggestion
call ok((args(2) == 'string'),  trim(testnum)//', arg 2')
call ok((args(3) == ''),        trim(testnum)//', arg 3')
```

developer_tests/utilities/parse_args_test.f90
```diff
@@ -1,90 +1,198 @@
...
+call ok((nargs == 3),           trim(testnum)//', nargs')
+call ok((names(1) == 'a'),      trim(testnum)//', name 1')
+call ok((vals(1)  == '1.0'),    trim(testnum)//', val 1')
+call ok((names(2) == 'b'),      trim(testnum)//', name 2')
+call ok((vals(2)  == '2.0'),    trim(testnum)//', val 2')
+call ok((names(3) == 'c'),      trim(testnum)//', name 3')
+call ok((vals(3)  == '3.0'),    trim(testnum)//', val 3')
+call ok((names(4) == ''),       trim(testnum)//', name 4')
+call ok((vals(4)  == ''),       trim(testnum)//', val 4')
+call ok((.not. continue_line),  trim(testnum)//', continue_line')
```
hkershaw-brown: You don't have a check for continue_line = .true. ( & ).
Maybe add & to this test

nextline = 'a = 1.0 b   =  2.0  c=3.0 &'

developer_tests/utilities/parse_args_test.f90
nancycollins: fixed

developer_tests/utilities/parse_args_test.f90
nancycollins: fixed

developer_tests/utilities/parse_args_test.f90
nancycollins: fixed - cut n paste error

nancycollins:
i think i've addressed all the helpful review comments.

Issue #1081, opened 2026-03-17
Documentation for quad_utils is missing


mgharamti:
### What's the issue? 
:books: Couldn't find any documentation for the quad utilities

### Where did you find the issue?
Provide the location(s) of the issue:  `models/utilities/quad_utils_mod.f90`

### What needs to be fixed?
The quad utils are very important and extremely useful for doing interpolation in large models. The documentation of these utilities is missing. It should clearly distinguish between the different types of supported interpolation (e.g., regular, regular but nonuniform, irregular, dipole vs tripole, ...) and provide instructions on how to use them. Perhaps some example cases and developer tests would be nice.

hkershaw-brown:
fixed in #1089

PR #1089, opened 2026-03-28
Quad Utilities: Enhanced Stability and New Documentation 

mgharamti:
## Description:

This PR improves the numerical robustness of quadrilateral interpolation in `quad_utils_mod` and introduces new module documentation. 

For the interpolation code, Cramer's rule (for solving the 3×3 system in`quad_bilinear_interp`) is now replaced with Gaussian Elimination with partial Pivoting (GEP). The new solver improves numerical stability, particularly for highly distorted or nearly degenerate quadrilateral cells. The computational cost remains constant (3×3 system) and negligible relative to the overall interpolation workflow.

I also added a new rst documentation page for `quad_utils_mod`. The documentation provides an overview of the interpolation procedure with detailed description of the supported grids, interpolation methods, and workflow. 

### Fixes issue
Fixes issue #1081

### Types of changes

- [x] Documentation update

### Tests
I ran the developer tests for the `quad_utils` and also tested the new changes on small grids using different distorted quads. I got identical results between previous and refactored implementations for well-conditioned grids. For pathological quad configurations, the new solver demonstrates improved accuracy (orders of magnitude reduction in error). 

## Checklist for merging
- [ ] Updated changelog entry
- [x] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets
- [x] No dataset needed


models/utilities/quad_utils_mod.f90
```diff
@@ -2246,15 +2245,22 @@ subroutine quad_bilinear_interp(lon_in, lat_in, x_corners_in, y_corners_in, cycl
...
-
+call gauss_elim_pivot(m, v, r, gep_stat)
+
+if (gep_stat /= 0) then 
+   if (debug > 0) then 
+      write(string1, '(2A)') 'The Gaussian Elimination with Partial Pivoting Scheme ', &
+                             'failed. It could be a singularity or a very tiny pivot'
+      write(string2, '(2(A, F10.4))') 'Obs location -> lon: ', lon, ', lat: ', lat  
+      write(string3, '(2(A, 4F10.4))') 'Quad corners -> x: ', x_corners, ', y: ', y_corners
+      call error_handler(E_MSG, 'quad_bilinear_interp', string1, &
```
hkershaw-brown: This message will only print on task 0, so if other tasks have failures you will not see the message. To have the message print on any task: 
E_ALLMSG

mgharamti: nice catch! I'll update this

models/utilities/quad_utils_mod.f90
```diff
@@ -2246,15 +2245,22 @@ subroutine quad_bilinear_interp(lon_in, lat_in, x_corners_in, y_corners_in, cycl
 if (debug > 10) write(*,'(A,I3,7F12.3)') 'i, m(3), p(2), v: ', i, m(i,:), p(i), p(i+1), v(i)
 enddo
 
-! look for degenerate matrix and rotate if needed
-! compute deter of m
-!d = deter3(m)
-
 ! Solve the matrix for b, c and d
```
hkershaw-brown, line 2248: @mgharamti Does this method need the do_rotate namelist option?

I ask because the code I think is doing a clockwise rotation (vs counterclockwise as the comment stays)
https://github.com/NCAR/DART/issues/833#issuecomment-2730460445
so if the rotation is not needed for stability with this new method, I think we should remove the rotation code.

mgharamti: Helen, in my tests rotations didn't add anything. I find the new solver to be stable without the need for rotations. Would that apply to the CICE examples is the issue you shared? probably.

hkershaw-brown: The do_rotate true/false did not improve the out-of-bounds for the CICE

mgharamti: hmm, ok. I don't know if users actively use this rotation option. For legacy reasons, we could keep it. It's defaulted to .false. anyways.

hkershaw-brown: ok I will put it on the todo list for the future to remove, and we leave it in for this pull request.

models/utilities/quad_utils_mod.rst
```diff
@@ -0,0 +1,200 @@
+.. _quad_utils_mod:
```
hkershaw-brown, line 1: Nice docs, I think it would be good to link to the quad_utils_mod from the "required model mod routines" doc

Definitely here:
https://docs.dart.ucar.edu/en/latest/guide/required-model-mod-routines.html#required-model-mod-routines

Maybe here?:
https://docs.dart.ucar.edu/en/latest/guide/advice-for-new-collaborators.html#interpolation

Something like:
"DART provides tools in the quad_utils_mod module for both grid search and interpolation on logically rectangular grids, that is, grids indexed like a regular (i,j) array but potentially curvilinear in physical space.     These quad_utils_mod routines can be used within model_interpolate() to avoid having to implement interpolation routines from scratch"

models/utilities/quad_utils_mod.rst
```diff
@@ -0,0 +1,200 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Dipole and tripole grids are treated as fully irregular grids. No special
+interpolation algorithm is used for these grids; instead, they are handled
+through the general quadrilateral interpolation framework.
+
+Mask Handling
+^^^^^^^^^^^^^
+
+An optional mask can be provided for fully irregular grids. If any of the four
```
hkershaw-brown, line 178: This is just a curiosity question, why do only irregular grids except masks?

mgharamti: Helen, good question. To me, I think of the regular grid's interpolation as index-based. We don't need mask to define geometry; the grid itself is valid everywhere. So, masking here is handled outside (e.g., point_on_land, quad_on_land, ...). 

On the other hand, interpolation in the irregular grids is geometry-based (physical space) in which we need the 4 corners to fit a surface. If one corner is land (or invalid), this will corrupt the interpolation. In essense, if any of the corners is masked we do not interpolate.  

We could add masks for regular grid, but it adds overhead. It's essential in the irregular case for correctness.

models/utilities/quad_utils_mod.rst
mgharamti: Good idea, will do.

hkershaw-brown:
Question on the tests:

> also tested the new changes on small grids using different distorted quads. I got identical results between previous and refactored implementations for well-conditioned grids. 

Are these test refering to `test_quad_reg_interp`  (well-conditioned grids) and `test_quad_irreg_interp` ( different distorted quads) or do you have other tests? 

> I ran the developer tests for the quad_utils

I'm not sure what the test is in the developer tests for quad_utils are testing, there's not a pass fail. So I just compared the .txt output from main and mgharamti:improve_quad_utils, (and plotted the results #1097)

The results are different on the last digit between main and this branch for `test_quad_irreg_interp`. I do not expect they would be bitwise between the 2 algorithms, because this is the point of the pull request. 
Both this branch and main give 'large interp residual' messages. 

[hkershaw:work] (improve_quad_utils) > ./test_quad_irreg_interp | grep 'large interp residual' | wc -l
   17049

[hkershaw:work] (main) > ./test_quad_irreg_interp | grep 'large interp residual' | wc -l
   15953

gfortran --version
GNU Fortran (MacPorts gcc13 13.3.0_2+stdlib_flag) 13.3.0
Copyright (C) 2023 Free Software Foundation, Inc.

mgharamti:
Helen, I built my own distorted quad tests. I also ran the the ones in the developer tests but I didn't find them very informative.

hkershaw-brown:
> Helen, I built my own distorted quad tests. I also ran the the ones in the developer tests but I didn't find them very informative.

cool, I figured you had something cooler from the plots you showed at standup.

hkershaw-brown (approved):
Approved!

Issue #978, opened 2025-10-07
pole for non-u component?


hkershaw-brown:
Just on note on possible bug, need to take a better look:

In pop this line is for u_current
https://github.com/NCAR/DART/blob/1ddc21f418175fab0768b572866f54a2983dbb66/models/POP/model_mod.f90#L457C15-L458

In quad utils this is in the general init_irreg_interp, which may or may not be u -> .  
https://github.com/NCAR/DART/blob/1ddc21f418175fab0768b572866f54a2983dbb66/models/utilities/quad_utils_mod.f90#L841-L842

also t_pole_y is set but not used.

PR #1102, opened 2026-04-27
quad_utils_mod search: replace 3D scratch arrays with compressed sparse row format

### LINKED ISSUE #979: quad_utils max_reg_list_num
quick note, will fill out later.

hard coded, not big enough

```
ERROR FROM:
  source : $URL$
  routine: update_reg_list
  message:  max_reg_list_num (         800 ) is too small ... increase
  message: ...  adding 1 to bin          132         142
  message: ...  bins:          131         162         139         142
```

hkershaw-brown:
## Description:

Quad_utils_mod box search for fully irregular grids:
* Determine  the search storage requirement at run time from the grid.  Currently there are 3 hardcoded cases, leading to people adding "big" numbers to the module to get their code running vs erroring out.  The hardcoded cases I believe are based on the POP workhorse and high res grids ~2015 (looking at POP/model_mod.f90 blame)
* Replaces the quad_utils_mod search storage (3D arrays) for fully irregular grids with a compressed sparse row format.

The runtime calculation of storage requirements is needed in particular for regional CESM where people are creating their own regional grids, but is applicable to any model using quad_utils.
The per-code memory is quite high for the box storage (see #979 & https://github.com/hkershaw-brown/quad_search_perf/blob/main/README.md). This pull request does not address distributing memory for the search storage (may be needed later), but does reduce the per core memory. 

Compressed Sparse Row format:
The GRID_QUAD_FULLY_IRREGULAR coarse-index search-info array setup previously allocated reg_list_lon/lat(nrx, nry, max_reg_list_num) as a 3D temporary, reaching ~5.5 GB at nrx=nry=900. It also errored out if any coarse box accumulated more than max_reg_list_num=800 candidate quads. Noted in issue #979

Replace with a two-pass compressed sparse row build:
- Pass 1: count overlaps into grid_num; peak memory ~11 MB
- Prefix sum: derive grid_start and exact total entry count
- Pass 2: fill flat lists using a cursor array

Replace hardcoded num_reg with calculation based on grid
  num_reg = max(10, min(nx_or_ny, int(sqrt(nx*ny / TARGET_CANDIDATES))))
where TARGET_CANDIDATES=8, giving ~8 candidate quads per coarse box regardless of grid resolution.

Remove no longer used max_reg_list_num from quad_irreg_grid_coords

To do:
- [ ] regional
- [ ] wrap long regional
- [ ] pole regional
- [x] global - bitwise

### Fixes issue

Fixes https://github.com/NCAR/DART/issues/979

### Types of changes

- [ ] Documentation update

### Tests
Looking at brute force vs. current method vs new method: https://github.com/hkershaw-brown/quad_search_perf
~~I've only created global test grids.~~ regional grids now, 
Quad_util_developer tests bitwise, but this are fairly limited. 

Currently running ROMS_rutgers case to compare memory (still in queue)

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/models/utilities/quad_utils_mod.f90
+++ b/models/utilities/quad_utils_mod.f90
@@ -99,6 +99,10 @@ logical  :: do_rotate = .false.     ! rotate edge from pts 1,2 to horizontal bef
 
 namelist /quad_interpolate_nml/ do_rotate, debug
 
+! Two-pass CSR build: target this many candidate quads per coarse box on average.
+! num_reg ≈ sqrt(nx*ny / TARGET_CANDIDATES), clamped to [10, min(nx,ny)].
+integer, parameter :: TARGET_CANDIDATES = 8
+
 !> @todo FIXME internal routines could use h for the handle; externally callable
 !> routines should use interp_handle for clarity in the interface.
 
@@ -206,7 +210,6 @@ type quad_irreg_grid_coords
    ! the sizes of these depend on the grid size.  these are good defaults for ?
    integer  :: num_reg_x = 180
    integer  :: num_reg_y = 180
-   integer  :: max_reg_list_num = 800
    real(r8) :: min_lon =     0.0_r8
    real(r8) :: max_lon =   360.0_r8
    real(r8) :: lon_width = 360.0_r8
@@ -404,42 +407,17 @@ select case (grid_type)
       interp_handle%ii%lats_2D(num_lons, num_lats) = MISSING_R8
       interp_handle%ii%lons_2D(num_lons, num_lats) = MISSING_R8
 
-      ! tx0.1v2 num_reg_x = num_reg_y = 900
-      ! tx0.5v1 num_reg_x = num_reg_y = 180
-      ! gx1v6   num_reg_x = num_reg_y = 90
-      ! max_reg_list_num = 800
+      ! Dynamic coarse grid sizing: target ~TARGET_CANDIDATES quads per coarse box.
+      ! num_reg ≈ sqrt(nx*ny / TARGET_CANDIDATES), clamped to [10, min(nx,ny)].
+      interp_handle%ii%num_reg_x = max(10, min(num_lons, &
+            int(sqrt(real(num_lons * num_lats, r8) / real(TARGET_CANDIDATES, r8)))))
+      interp_handle%ii%num_reg_y = max(10, min(num_lats, &
+            int(sqrt(real(num_lons * num_lats, r8) / real(TARGET_CANDIDATES, r8)))))
 
-      ! adjust num_regs here based on numlons, numlats
-      if (num_lats * num_lons > 6 * 1000 * 1000) then  ! ~1/10th degree
-         interp_handle%ii%num_reg_x = 900
-         interp_handle%ii%num_reg_y = 900
-         interp_handle%ii%max_reg_list_num = 800   !todo  what is good val?
-         if(debug > 10) then
-            write(string1, *) 'case 1: ', interp_handle%ii%num_reg_x, interp_handle%ii%num_reg_y, &
-                               interp_handle%ii%max_reg_list_num
-            call log_it(string1)
-         endif
-
-      else if (num_lats * num_lons > 250 * 1000) then  ! ~1/2th degree
-         interp_handle%ii%num_reg_x = 180
-         interp_handle%ii%num_reg_y = 180
-         interp_handle%ii%max_reg_list_num = 800
-         if(debug > 10) then
-            write(string1, *) 'case 2: ', interp_handle%ii%num_reg_x, interp_handle%ii%num_reg_y, &
-                               interp_handle%ii%max_reg_list_num
-            call log_it(string1)
-         endif
-
-      else
-         interp_handle%ii%num_reg_x = 90
-         interp_handle%ii%num_reg_y = 90
-         interp_handle%ii%max_reg_list_num = 800
-         if(debug > 10) then
-            write(string1, *) 'case 3: ', interp_handle%ii%num_reg_x, interp_handle%ii%num_reg_y, &
-                               interp_handle%ii%max_reg_list_num
-            call log_it(string1)
-         endif
-
+      if (debug > 0) then
+         write(string1, *) 'two-pass num_reg_x, num_reg_y: ', &
+               interp_handle%ii%num_reg_x, interp_handle%ii%num_reg_y
+         call log_it(string1)
       endif
 
       allocate(interp_handle%ii%grid_start(interp_handle%ii%num_reg_x, &
@@ -727,7 +705,9 @@ endif
 end subroutine shapecheck
 
 !------------------------------------------------------------
-!> Build the data structure for interpolation for an irregular quad grid
+!> Build the data structure for interpolation for an irregular quad grid.
+!> Two-pass CSR build: pass 1 counts, prefix sum allocates exact flat lists,
+!> pass 2 fills.  No 3-D temporary arrays, no max_reg_list_num cap.
 
 subroutine init_irreg_interp(h)
 
@@ -735,20 +715,13 @@ type(quad_interp_handle), intent(inout) :: h
 
 character(len=*), parameter :: routine = 'init_irreg_interp'
 
-! Need a temporary data structure to build this.
-! These arrays keep a list of the x and y indices of dipole quads
-! that potentially overlap the regular boxes.
-integer, allocatable :: reg_list_lon(:,:,:)
-integer, allocatable :: reg_list_lat(:,:,:)
+integer, allocatable :: fill_pos(:,:)
 
 real(r8) :: u_c_lons(4), u_c_lats(4), pole_row_lon
-integer  :: i, j, k, pindex, nx, ny, nrx, nry, istatus
-integer  :: reg_lon_ind(2), reg_lat_ind(2), u_total, u_index
+integer  :: i, j, pindex, nx, ny, nrx, nry, istatus
+integer  :: reg_lon_ind(2), reg_lat_ind(2), u_total
 logical  :: cyclic, pole
 integer  :: xlim
-
-allocate(reg_list_lon(h%ii%num_reg_x, h%ii%num_reg_y, h%ii%max_reg_list_num))
-allocate(reg_list_lat(h%ii%num_reg_x, h%ii%num_reg_y, h%ii%max_reg_list_num))
 
 ! poles?  span?
 cyclic = h%opt%spans_lon_zero
@@ -757,9 +730,6 @@ nx  = h%nlon
 ny  = h%nlat
 nrx = h%ii%num_reg_x
 nry = h%ii%num_reg_y
-
-reg_list_lon(:, :, :) = 0
-reg_list_lat(:, :, :) = 0
 
 ! for a global grid, the initial values have already been set in
 ! the derived type.  otherwise, find the min/max of lons and lats.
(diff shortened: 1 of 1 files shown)


hkershaw-brown:
Q. global vs. regional. 
if you initialize the interpolation as global, but the grid is regional how does this affect things?

wrap where there is no wrap. (mad wrapping of region around globe)
global, sparse -> slow search
global, not sparse, too many points in a box. -> error out or slow search

How bad is it to have mad-wrap? <-- -->  

https://github.com/NCAR/DART/blob/acf0a14d4cf371c11aafbac7d7cf9cf356fd56a7/models/utilities/quad_utils_mod.f90#L764-L769

hkershaw-brown:
or regional refinements in global grids: 

Screenshot 2026-04-28 at 9 59 45 AM
Screenshot 2026-04-28 at 9 59 41 AM

hkershaw-brown:
@mgharamti this is the quad search I am playing with. I think the main concerns are calculating the course grid from the actual grid rather than the hardcoded 3 cases, and the search structure size. But open to any input. 

https://github.com/NCAR/DART/blob/acf0a14d4cf371c11aafbac7d7cf9cf356fd56a7/models/utilities/quad_utils_mod.f90#L730-L751

models/utilities/quad_utils_mod.f90
```diff
@@ -818,73 +788,69 @@ subroutine init_irreg_interp(h)
...
 enddo
 
-! Confirm that the indices come out okay as debug
-if(u_index /= u_total + 1) then
-   string1 = 'Storage indices did not balance for U grid: : contact DART developers'
-   call error_handler(E_ERR, routine, string1, source, revision, revdate)
-endif
-
-deallocate(reg_list_lon, reg_list_lat)
+deallocate(fill_pos)
```
mgharamti, line 877: The new new approach is based on the assumption (fact?) that what we counted in the first pass is equal to what we filled in the second pass. I'd be more comfortable if we added a sanity check after filling (similar to the check in the old code). You could do something like this: 

```
do i = 1, nrx
   do j = 1, nry
      if (fill_pos(i,j) /= h%ii%grid_start(i,j) + h%ii%grid_num(i,j)) then
         call error_handler(E_ERR, 'init_irreg_interp', &
              'count/fill mismatch in irregular grid interpolation lists')
      endif
   enddo
enddo
```

models/utilities/quad_utils_mod.f90
```diff
@@ -818,73 +788,69 @@ subroutine init_irreg_interp(h)
...
-
- !>@todo is istatus /= 0 a failure condition
-
-         ! Set up array of lons and lats for the corners of these u quads
+      if( all_corners_valid(h%opt, i, j, nx) ) then
          call get_quad_corners(h%ii%lons_2d, i, j, cyclic, pole, nx, ny, u_c_lons, istatus)
          if (istatus /= 0) print *, 'get_quad_corners for lons returns failure'
-
          call get_quad_corners(h%ii%lats_2d, i, j, cyclic, pole, nx, ny, u_c_lats, istatus)
          if (istatus /= 0) print *, 'get_quad_corners for lats returns failure'
```
mgharamti, line 804: Do we need to cycle here rather than just print an error? If a failure produces bad corners, both passes may count/fill bad bins or one failure path could behave differently depending on the state (right?)

mgharamti (changes requested):
@hkershaw-brown, thanks for the work on this. I was always on the edge with the previous style of 3 hard-coded cases but didn't know how to go around that. I think I understand your design and strategy. CSR clearly avoids working with huge 3D arrays. 

My first lazy question is why the 8 targets on average? I guess how did you choose it? It seems to me that search cost per obs is ~O(num_grids) and number of bins/boxes is ~O(N/TARGET_CANDIDATES) and so it's a performance tuning parameter. At least we need to explain it in the docs if you decide to go with it. 

Second, could we add a post-fill consistency check that verifies fill_pos(i,j) == grid_start(i,j) + grid_num(i,j) for every coarse bin? This would catch any future divergence between the counting pass and filling pass, especially around edge cases such as longitude wrapping, masked/invalid quads, or boundary-touching cells.

models/utilities/quad_utils_mod.f90
hkershaw-brown: this is a great idea,  & the check is in init so not a problem runtime wise to add.

hkershaw-brown:
> My first lazy question is why the 8 targets on average? I guess how did you choose it? It seems to me that search cost per obs is ~O(num_grids) and number of bins/boxes is ~O(N/TARGET_CANDIDATES) and so it's a performance tuning parameter. At least we need to explain it in the docs if you decide to go with it.

My lazy choice: Dr. Google for what was a good choice for this kind of algorithm. 
I was looking at brute force (best memory, but the worst case timing ~50,000 slower) vs. current  hard coded method (max memory, target is to go as fast as that). 
https://github.com/NCAR/DART/issues/979#issuecomment-4271196214.   

I think the equivalent average target (`nx*ny/num_reg*num_reg`)is ~10ish for the hard coded versions for the POP grids:


|grid | nx	        |         ny	 |num_reg |average quads per coarse box|
|---------|----------|---------|------| --|
|gx1  (3deg?)|     320|	384	|  90	       |  15.17037037|
| tx0.5 0.5deg | 720|	480	 | 180|	10.66666667|
| tx0.1 0.1 deg | 3600|	 2400|	900|	10.66666667|

The grids are not uniform so 'average' is probably a disgusting assumption.  

I think it is well worth doing some runs with the test code and different (and variable) grid resolutions to count the actual number of quads per box with the new code (and counting the per course box with the 3 hardcoded POP cases). Maybe we have 799 quads in one box, 3 in another.  Currently the 8 is just an ok choice where I was finding comparable speeds to the current version, but it is definitely a magic number that needs justifying.

models/utilities/quad_utils_mod.f90
```diff
@@ -818,73 +788,69 @@ subroutine init_irreg_interp(h)
...
 endif
 
+! ---------------------------------------------------------------
+! PASS 1: count how many quads overlap each coarse box.
+! No storage of quad indices — just increment grid_num.
+! ---------------------------------------------------------------
+h%ii%grid_num = 0
+
 do i = 1, xlim
-   ! There's no wraparound in y, one box less than grid boundaries
```
hkershaw-brown, line 822: this is a good comment, put back in @hkershaw-brown

hkershaw-brown: actually can you use quad_utils with a periodic y domain? e.g. wrf idealized. Not at the moment I guess.

models/utilities/quad_utils_mod.f90
hkershaw-brown: I will take a closer look - I think I am missing the logic in the original code. 

The loop limits are the x,y bounds of the lon and lat arrays.
Is the print just checking if the loop limits are wrong?

If you can go off the end of the arrays,
it seems like both should fail, or both should pass. 

If one fails and one passes I think this should be a catastrophic error rather than a print, since lon,lat are for the same point.

models/utilities/quad_utils_mod.f90
hkershaw-brown: Added assertion that pass 1 count matches pass 2 fill in 117ba47d5acd709f553b3d2c610a00a977e2b72e

Also logs the box info:  max #quads, min# quads, mean #quads, #boxes that have 0, how boxes have  >2*mean quads, so you can see how the boxes are being filled.

models/utilities/quad_utils_mod.f90
```diff
@@ -1287,68 +1277,72 @@ end subroutine quad_index_neighbors
...
+   enddo
+enddo
 
-      if ((index_x < 1 .or. index_x > nrx) .or. (index_y < 1 .or. index_y > nry)) then
-         string1 = 'unable to find right box'
-         write(string2,*) 'index_x may be out-of-range: ', 1, index_x, nrx
-         write(string3,*) 'index_y may be out-of-range: ', 1, index_y, nry
-         call error_handler(E_ERR,'update_reg_list',string1, &
-                 source, revision, revdate, text2=string2, text3=string3)
-      endif
```
mjs2369, line 1328: Shouldn't we keep this check? Even though this is now just the first pass, I think this check would still work at the end of count_reg_overlaps

Assuming I'm understanding the initial motivation for this check correctly, I believe it is still possible for index_x or index_y to return invalid box numbers due to wraparound shenanigans

models/utilities/quad_utils_mod.f90
```diff
@@ -818,73 +788,93 @@ subroutine init_irreg_interp(h)
...
 
-write(string1,*)'to determine (minimum) max_reg_list_num values for new grids ...'
-write(string2,*)'interp_handle%ii%grid_num is ',maxval(h%ii%grid_num)
-call error_handler(E_MSG, routine, string1, text2=string2)
+write(string1,*)'two-pass: max candidates per coarse box = ', maxval(h%ii%grid_num)
+call error_handler(E_MSG, routine, string1)
+write(string1,*)'two-pass: min candidates per coarse box = ', minval(h%ii%grid_num)
+call error_handler(E_MSG, routine, string1)
+write(string1,*)'two-pass: empty coarse boxes = ', count(h%ii%grid_num == 0), ' of ', nrx*nry
+call error_handler(E_MSG, routine, string1)
```
mjs2369, line 816: Do we want to always print these statements / put them in the dart_log? Seems like they could be debug statements.

if (debug > 0) then

mjs2369: same comment for 830-7

mjs2369:
I think the updates here look really solid. Two-pass CSR looks clean and is an awesome way to reduce the memory overhead of the quad search. The computation time of get_quad_corners is tiny enough that doing it twice doesn't really make a difference in the runtime. Makes you wonder what possibilities could be out there for similar-ish updates elsewhere in the code

Just checking though that this was tested on sufficiently large grids - I see that the timing / memory comparison on quad_search_perf was done with size Nx = 5000, Ny = 4500
Did that ROMS_Rutgers test you mentioned ever finish? How big can we expect grid size to reasonably get, and would having a huge / fine grid change how well the two-pass CSR build performs time wise?

Dynamic calculation of the dimensions for the coarse grid is also clearly a huge win, but I am also curious about the value for TARGET_CANDIDATES being 8.

Earlier you mentioned you were planning to do more tests: "I think it is well worth doing some runs with the test code and different (and variable) grid resolutions to count the actual number of quads per box with the new code (and counting the per course box with the 3 hardcoded POP cases). "
Any findings from these?

models/utilities/quad_utils_mod.f90
hkershaw-brown: Yes in the dart log because they are telling the user how the interpolation search is being setup. 
Its not a crazy amount of extra lines in the log, but is important info on the search:

```
PE 0: init_irreg_interp  two-pass: max candidates per coarse box =          231
 PE 0: init_irreg_interp  two-pass: min candidates per coarse box =            0
 PE 0: init_irreg_interp  two-pass: empty coarse boxes =        23384  of        24964
 PE 0: init_irreg_interp  two-pass: total coarse-index entries =       281844
 PE 0: init_irreg_interp two-pass: mean candidates per coarse box =   11.3
 PE 0: init_irreg_interp  two-pass: boxes with >2x mean =         1580  of        24964
```

hkershaw-brown:
> I think the updates here look really solid. Two-pass CSR looks clean and is an awesome way to reduce the memory overhead of the quad search. The computation time of get_quad_corners is tiny enough that doing it twice doesn't really make a difference in the runtime. Makes you wonder what possibilities could be out there for similar-ish updates elsewhere in the code
> 

get_close uses a two pass, I am not sure why quad utils did not originally use a two pass.  But for sure there are all sorts of places in the code that could be improved. 
The non-two pass is in POP and CICE. 

> Just checking though that this was tested on sufficiently large grids - I see that the timing / memory comparison on quad_search_perf was done with size Nx = 5000, Ny = 4500 Did that ROMS_Rutgers test you mentioned ever finish? How big can we expect grid size to reasonably get, and would having a huge / fine grid change how well the two-pass CSR build performs time wise?
> 
> Dynamic calculation of the dimensions for the coarse grid is also clearly a huge win, but I am also curious about the value for TARGET_CANDIDATES being 8.

Its just a heuristic, here is the comment to Moha on the same question. It was ~10 on main for the hardcoded choice. 

https://github.com/NCAR/DART/pull/1102#issuecomment-4360515336

> 
> Earlier you mentioned you were planning to do more tests: "I think it is well worth doing some runs with the test code and different (and variable) grid resolutions to count the actual number of quads per box with the new code (and counting the per course box with the 3 hardcoded POP cases). " Any findings from these?

This is one of the reasons for keeping the diagnostics prints in the log file about how many candidates are in each box. 
Note there is also a  bug(s) https://github.com/NCAR/DART/issues/1124 (in main and this pull request) with longitudes close to 0 and https://github.com/NCAR/DART/issues/978
This pull request is mostly to allow bigger grids that the hardcoded values in main, and use the same search algorithm as main but use less memory per core.  I think the quad_utils_mod could be improved greatly https://github.com/NCAR/DART/issues?q=is%3Aissue%20state%3Aopen%20quad_utils however I have not made the attempt to do this in this pull request, just made issues as a found them.

models/utilities/quad_utils_mod.f90
mjs2369: @hkershaw-brown last outstanding comment before approval

hkershaw-brown: Put check back in: b0c90c194bbbbe97c07e03085f65875de8cadcb5

mjs2369 (approved):
Nice work Helen

models/utilities/quad_utils_mod.f90
hkershaw-brown: moved to separate issue #1160

hkershaw-brown:
@mgharamti any remaining issues with this pull request?  I believe I addressed the requested changes. 

Cheers,
Helen

mgharamti (approved):
Looks good and ready to ship!

Issue #595, opened 2023-12-11
Feature request: ifx compiler usage for DART


hkershaw-brown:
### Use case  
intel ifx compiler 

### Is your feature request related to a problem?  
Deprecation of intel classic compiler ifort:
https://community.intel.com/t5/Blogs/Tech-Innovation/Tools/Deprecation-of-The-Intel-Fortran-Compiler-Classic-ifort/post/1541699

### Describe your preferred solution  

*  Determine approriate flags for DART, production, debugging, bitwise, (particularly [floating point behaviour](https://www.intel.com/content/www/us/en/developer/articles/guide/porting-guide-for-ifort-to-ifx.html)).
*  Create a mkmf.template.ifx
* replace fixsystem with f2008 standards, EXECUTE_COMMAND_LINE

### Describe any alternatives you have considered  
same as above but update fixsystem to have ifx option.

mjs2369:
@hkershaw-brown it seems that this has been addressed and can be closed?

- mkmf.template.ifx.linux now available with appropriate flags
- fixsystem updated to have ifx option
- ifx added to new build_everything 

The bullet point to replace fixsystem with f2008 standards is being addressed with the general update to the code to be compliant with the newer fortran standards

hkershaw-brown:
Mechanically the ifx is available in dart (you can build with it no problem, and run the code.). 

There is a part of the porting guide about the floating point behaviour:

>Floating Point Behavior
To get ifort behavior in floating point comparisons, ifx users should specify assume [no]ieee_compares and/or assume nan_compares.
Tells the compiler to generate IEEE compareSignaling operations for floating point comparisons, as required by the Fortran 2018 standard. By default ifx assume NaNs are not going to be operands in floating point comparisons and does not generate that check for NaNs.  
Linux
-assume [no]ieee_compares
Windows
/assume:[no]ieee_compares
fp-model | fp and fast
The behavior with the fp-model | fp fast=1 or 2 option is different between ifort and ifx with respect to floating point compares.  To get ifort behavior when using the fp-model or fp fast option with ifx, assume nan_compares should also be specified on the ifx command line.
Linux
-fp-model fast=1 or -fp-model fast=2 -assume nan_compares
Windows
/fp:fast 1 or /fp:fast 2 /assume:nan_compares

So in particular for us, how it deals with nans and how the fp model options affect the results
-fp model (and fp fast=1, 2)

I think this needs a bit of exploration before marking this issue as done.

[...]

Issue #634, opened 2024-02-08
bug: NCEP/prep_bufr code outdated, fails to compile with gfortran


mjs2369:
### Describe the bug
The code in this directory is very outdated and difficult to use and produces some compile time errors with more recent versions of gfortran. This is not DART code, and has been fragile in the past (similar to the JPL module code). In order to change this code, we need to have extensive tests. 

This code is not built with the dart build system (mkmf, quickbuild). Instead, there's two scripts for building, one being bash and the other cshell: [DART/observations/obs_converters/NCEP/prep_bufr/install.sh](https://github.com/NCAR/DART/blob/main/observations/obs_converters/NCEP/prep_bufr/install.sh) and [DART/observations/obs_converters/NCEP/prep_bufr/convert_bufr/convert_bufr.csh](https://github.com/NCAR/DART/blob/main/observations/obs_converters/NCEP/prep_bufr/convert_bufr/convert_bufr.csh). 

The following files also include the use of outdated subroutines argc and getarg, which were replaced by the fortran intrinsics everywhere else in the DART code:
observations/obs_converters/NCEP/prep_bufr/convert_bufr/arg_test.f and observations/obs_converters/NCEP/prep_bufr/convert_bufr/stat_test.f:
observations/obs_converters/NCEP/prep_bufr/convert_bufr/arg_test.f

Finally, grabbufr.f fails to compile with gfortran 12.2.0 due to type mismatch errors (see screenshot). It does compile with intel.

1. List the steps someone needs to take to reproduce the bug.  
Run "./install.sh" and then "./convert_bufr/convert_bufr.csh" using gfortran. To do so, you must edit convert_bufr.csh to set gfortran as the Fortran compiler.
2. What was the expected outcome?
Three executables are produced: stat_test.x, arg_test.x, and grabbufr.x
3. What actually happened?  
Grabbufr.f fails to compile with multiple errors

### Error Message  
Please provide any error messages.

### Which model(s) are you working with?
NCEP observation converter

### Screenshots   
![image](https://github.com/NCAR/DART/assets/44214771/2128f395-79ec-4d1d-8f0f-b6c1d957468b)

### Version of DART      
v11.0.2

### Have you modified the DART code?   
No

### Build information
Please describe:  
 1. The machine you are running on (e.g. windows laptop, NCAR supercomputer Cheyenne).   
 3. The compiler you are using (e.g. gnu, intel).  

All machines, gfortran 12.2.0

mjs2369:
[...]

PR #1121, opened 2026-06-02
Remove fixsystem from the build system

### LINKED ISSUE #870: bug: mkmf + fixsystem relies on the FC variable in the makefile being the name of the compiler

### Describe the bug

   Build dart on a machine such as Perlmutter, where the recommended compiler wrapper is ftn.
  FC = ftn
  Or in EMSFx containers where the compiler is gfortran-version
   You can build DART
3. What actually happened?  
  DART fails to build
### Error Message  

```
ftn types_mod.o null_mpi_utilities_mod.o parse_args_mod.o preprocess.o utilities_mod.o time_manager_mod.o -o /global/u2/h/hkershaw/DART/DART/assimilation_code/programs/preprocess/preprocess  -O2 -ffree-line-length-none -I/opt/cray/pe/netcdf/4.9.0.13/gnu/12.3/include -L/opt/cray/pe/netcdf/4.9.0.13/gnu/12.3/lib -lnetcdff -lnetcdf
/usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: null_mpi_utilities_mod.o: in function `__mpi_utilities_mod_MOD_shell_execute':
null_mpi_utilities_mod.f90:(.text+0x391): undefined reference to `system_'
collect2: error: ld returned 1 exit status
make: *** [Makefile:45: /global/u2/h/hkershaw/DART/DART/assimilation_code/programs/preprocess/preprocess] Error 1
```

You can set FC = gfortran

```
MPIFC = ftn
MPILD = ftn
FC = gfortran
LD = ftn
```

but then you loose the compiler wrapper functionality (e.g. linking to netcdf, whatever else they have set up on the machine). 

ESM3, lorenz_96, all models.

### Version of DART      
v11.10.9

No 

### Build information
Please describe:  
 1. Perlmutter
 2. The compiler you are using (e.g. gnu, intel).   GNU Fortran (SUSE Linux) 13.2.1 20240206

mjs2369:
## Description:

In Fortran 2008 and later standards, EXECUTE_COMMAND_LINE is the preferred intrinsic for executing system commands.

With the use of the Fortran intrinsic execute_command_line, fixsystem is no longer needed for fixing compiler dependencies in the code

This PR replaces calls to the C system function and Fortran instrinsic sleep() with execute_command_line and replaces 'call exit(error_code)' with 'stop error_code'; in the past, these functions either required or rejected interfaces/use statements depending on the compiler (NAG, gfortran). Both execute_command_line and stop work with all compilers and therefore removes the need for these blocks of compiler dependent source code.

This means that the DART code no longer needs to be changed with fixsystem (by commenting in/out these specific blocks of code). mkmf + fixsystem are also no longer reliant on the FC variable in the makefile being the name of the compiler

An additional benefit as well to no longer having fixsystem is that the mpi utilities mods are not automatically edited upon running quickbuild with certain compilers, therefore messing up your git working tree.

This does not change the way in which the user builds DART or interacts with DART.

Link to the spec for more info - [https://docs.google.com/document/d/1xOb25pg8vJhHJy7xu6CtFFH_AklQbxpVDB7o8qNfGKs/edit?tab=t.0](https://docs.google.com/document/d/1xOb25pg8vJhHJy7xu6CtFFH_AklQbxpVDB7o8qNfGKs/edit?tab=t.0)

I also updated various bits and bobs of the code to be consistent with these changes.

- In the developer_tests/mpi_utilities/tests , removed the use of fixsystem. These dev tests do not build with quickbuild though ; see #1123 
- Replaced 'call exit(error_code)' with stop error_code in WRF utility programs, developer_tests/forward_operators/rttov_unit_tests.f90

Note for @hkershawbrown - if we just call execute_command_line directly in exit_all, we don’t strictly NEED to move the exit_all subroutine from the mpi_utilities mods to its own module. BUT this is maybe still something we want to do. Maybe this would be best in a separate PR as well

This PR also removes "ifort" from build_everything as it is no longer supported on Derecho.
 
### Fixes issue
Fixes #870 
Also mentioned in #595

### Types of changes

- [ ] Documentation update

Update to the build system 

### Tests
Built and ran lorenz_96 with all mpi options and NAG (Izumi), gfort, ifx compilers

Repeated this after I added some code to force DART to call sleep and exit; works great and is sleeping and exiting appropriately; also repeated this with async4_verbose = .true. causing execute_command_line to be called  

Ran quickbuild for some larger models as well with NAG (WRF, cam-fv)
run_all_quickbuilds with all compilers on Derecho - identical failures to that when run on the main branch

NOTE here that the NAG build is failing with mpif08, but this is independent of these changes. This is an issue with the outdated version of the NAG compiler and the mpif08 module:
```
NAG Fortran Compiler Release 7.0(Yurakucho) Build 7005
Option warning: The -mismatch option disables the -C=calls option
Fatal Error: /home/masmith/DART/assimilation_code/modules/utilities/mpif08_utilities_mod.f90, line 48: Cannot find module MPI_F08
             detected at MPI_F08@<end-of-statement>
```

Supposedly the later versions of NAG (7.1) fix this issue, but 7.0 is the latest version available on Izumi. Here is the separate issue for this #1122 

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed


hkershaw-brown:
very exciting Marlee

models/wrf/WRF_DART_utilities/add_pert_where_high_refl.f90
```diff
@@ -112,105 +112,105 @@ PROGRAM add_pert_where_high_refl
 
 if( COMMAND_ARGUMENT_COUNT() .ne. 13 ) then
   print*, 'INCORRECT # OF ARGUMENTS ON COMMAND LINE:  ', COMMAND_ARGUMENT_COUNT()
-  call exit(1)
+  stop 1
```
hkershaw-brown, line 115: are these changes because of fixstystem being removed, or are they due to exit() no being fortran standard and causing problems with compilers (nvhpc?)

hkershaw-brown: By these changes I mean the exit(X) in the wrf_dart_utiilies, prepbufr code

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -51,66 +51,11 @@ module mpi_utilities_mod
...
-! !! end block
-!
-!  !#endif
-! !!NAG_BLOCK_EDIT END COMMENTED_OUT
+! to decide what needs preprocessing
 
 implicit none
 private
 
-!include "mpif.h"
```
hkershaw-brown, line 90: This 
! include "mpif.h" is part of the build tip 1 above.  Where to put include mpif.h instead of use mpi

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -1829,19 +1772,21 @@ end function shell_execute
...
-! !!OTHER_BLOCK_EDIT START COMMENTED_IN
-    rc = system(trim(shell_name)//' '//trim(execute)//' '//char(0))
-! !!OTHER_BLOCK_EDIT END COMMENTED_IN
+integer :: exitstat
+integer :: cmdstat
+character(len=256) :: cmdmsg
 
-end subroutine do_system
+    call execute_command_line(trim(execute), exitstat=exitstat, &
+       cmdstat=cmdstat, cmdmsg=cmdmsg)
```
hkershaw-brown: looks like a few optional arguments here but only exitstat is used.

assimilation_code/modules/utilities/mpi_utilities_mod.f90
hkershaw-brown, line 1800: This routine is not called in DART, but it is public.  Wonder if there are people out there making use of this routine.

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -2005,8 +1953,7 @@ end module mpi_utilities_mod
 !-----------------------------------------------------------------------------
 !-----------------------------------------------------------------------------
 
-!> In case of error, call this instead of the fortran intrinsic exit().
-!> It will signal the other MPI tasks that something bad happened and they
+!> Signals to the other MPI tasks that something bad happened and they
 !> should also exit.
 
 subroutine exit_all(exit_code)
```
hkershaw-brown, line 1961: At some point the exit logic from utilities should be revisited. 
This pull request is skirting around the issue to only deal with fixsystem. 


https://github.com/NCAR/DART/blob/890a3beec7cde2099218ea30efa7c2b268be968d/assimilation_code/modules/utilities/utilities_mod.f90#L89-L99

assimilation_code/modules/utilities/mpif08_utilities_mod.f90
```diff
@@ -51,66 +51,11 @@ module mpi_utilities_mod
 ! We build on case-insensitive systems so we cannot reliably
 ! count on having the build system run the fortran preprocessor
 ! since the usual distinction is between bob.F90 and bob.f90
-! to decide what needs preprocessing.  instead we utilize a
-! script we provide called 'fixsystem' which looks for the
```
hkershaw-brown, line 55: I think the whole "we build on case-insenstivie ..." can go, since null|mpi|mpif08 are not preprocessed. 

An advantage of not having fixsystem is the git status (you don't instantly have local modifications when you build with certain compilers).

assimilation_code/modules/utilities/mpif08_utilities_mod.f90
```diff
@@ -1829,19 +1775,21 @@ end function shell_execute
...
-! !!OTHER_BLOCK_EDIT START COMMENTED_IN
-    rc = system(trim(shell_name)//' '//trim(execute)//' '//char(0))
-! !!OTHER_BLOCK_EDIT END COMMENTED_IN
+integer :: exitstat
+integer :: cmdstat
+character(len=256) :: cmdmsg
 
-end subroutine do_system
+    call execute_command_line(trim(execute), exitstat=exitstat, &
+       cmdstat=cmdstat, cmdmsg=cmdmsg)
```
hkershaw-brown: Same comment, optional arguments which are then not used.

assimilation_code/modules/utilities/null_mpi_utilities_mod.f90
```diff
@@ -504,19 +454,21 @@ end function shell_execute
...
-! !!OTHER_BLOCK_EDIT END COMMENTED_IN
+integer :: exitstat
+integer :: cmdstat
+character(len=256) :: cmdmsg
 
-end subroutine do_system
+    call execute_command_line(trim(execute), exitstat=exitstat, &
+       cmdstat=cmdstat, cmdmsg=cmdmsg)
+
+    rc = exitstat
```
hkershaw-brown: same comment on optional arguments not used.

assimilation_code/modules/utilities/mpif08_utilities_mod.f90
```diff
@@ -262,7 +207,7 @@ subroutine initialize_mpi_utilities(progname, alternatename, communicator)
 call MPI_Comm_rank(my_local_comm, myrank, errcode)
 if (errcode /= MPI_SUCCESS) then
    write(*, *) 'MPI_Comm_rank returned error code ', errcode
-   call exit(-99)
+   error stop errcode
```
hkershaw-brown: explain to me 
```
stop exit_code
```
vs/
```
error stop errcode
```

build_templates/mkmf
hkershaw-brown: May as well take out the svn junk at the top if you're taking the bottom svn junk out
```diff
- DART $Id$
```

developer_tests/build_everything/run_all_quickbuilds.sh
```diff
@@ -65,17 +65,6 @@ esac
 
 cp build_templates/$mkmf_template build_templates/mkmf.template
 
-# Run fixsystem to avoid all make commands from altering mpi_*_utilities_mod.f90 simultaneously
```
hkershaw-brown, line 68: nice not to have this, no race condition now since no need to run fixsystem

developer_tests/mpi_utilities/tests/ftest_mpi.f90
```diff
@@ -16,57 +16,8 @@ program ftest_mpi
...
-! !  SUBROUTINE EXIT(STATUS)
-! !    INTEGER,OPTIONAL :: STATUS
-! !! end block
-!
-! !#endif
-! !!NAG_BLOCK_EDIT END COMMENTED_OUT
-
 implicit none
 
-!include "mpif.h"
```
hkershaw-brown, line 47: !include "mpif.h"
should stay in.

developer_tests/mpi_utilities/tests/ftest_mpi.f90
```diff
@@ -120,28 +69,22 @@ program ftest_mpi
 contains
 
 !> wrapper so you only have to make this work in a single place
```
hkershaw-brown, line 72: Have you tried this with csh?

assimilation_code/modules/utilities/mpi_utilities_mod.f90
mjs2369: good note to keep on this PR

assimilation_code/modules/utilities/mpi_utilities_mod.f90
mjs2369: Another good note!

models/wrf/WRF_DART_utilities/add_pert_where_high_refl.f90
mjs2369: These changes are actually for NAG compatibility - NAG can't use exit(X) without an interface being added to the code

nvhpc can't use 'error stop X' but can use 'stop X'

assimilation_code/modules/utilities/mpif08_utilities_mod.f90
mjs2369: @hkershaw-brown 

stop exit_code - normal termination
error stop exit_code - error termination 

These two termination processes are made to work with Coarray, which is a Fortran mechanism for parallelism.  Coarray uses this concept of "images", which is conceptually the same thing as an MPI task. So the two termination processes differ in the way they handle synchronizing multiple images upon termination. Obviously, we use MPI, so they are effectively the same to DART. 

The only other difference is the error code they return if one isn't specified: "If stop-code is a character expression, STOP returns a status of zero and ERROR STOP returns a status of non-zero. If stop-code is an integer, a status equal to stop-code is returned for both STOP and ERROR_STOP." But since we are always passing an integer code to stop, again these two statements behave the same in the DART code. 

And having run both versions, the behavior was exactly the same.

We only call stop from either the null_mpi_utilities_mod or before initialize_mpi_utilities, so we also don't have to be concerned with dealing with MPI at all. Once we initialize MPI, we use MPI_abort instead. 

So we actually never needed to use 'error stop' over 'stop'; so it works out that the regular stop is supported by all the compilers, while 'error stop' is incompatible with nvhpc

I forgot to replace these in the non-null versions of the mpi utils mods as shown above, but I've done so now.

Eventually, I think we could reorg a bit and then use the error handler instead

developer_tests/mpi_utilities/tests/ftest_mpi.f90
mjs2369: @hkershaw-brown I did. I did so by doing `call do_execute_command_line("csh -c 'echo hello world'", rc)` on line 56 instead of just using bash by default.

I also wrote a tiny bit of code to verify it's actually running in csh by using setenv. This is csh-specific syntax that would fail in bash:

```
! Test execute_command_line using csh explicitly
! setenv is csh-specific syntax; if rc==0, csh ran it successfully
call do_execute_command_line("csh -c 'setenv TESTVAR hello && echo $TESTVAR'", rc)
if (rc /= 0) then
   print *, 'csh execute_command_line test FAILED - rc = ', rc
else
   print *, 'csh execute_command_line test passed'
endif
```

If rc comes back non-zero, then the command failed. But it returned 0 so success:

```
masmith@derecho4:/glade/derecho/scratch/masmith/DART11.24.0/DART/developer_tests/mpi_utilities/tests> ./ftest_mpi 
 program start
 MPI initialized successfully
 My MPI rank is:            0
 Total MPI tasks:            1
hello world
 All MPI calls succeeded, test passed.
 program end
hello
 csh execute_command_line test passed
```

build_templates/mkmf
mjs2369: @hkershaw-brown see this commit - [https://github.com/NCAR/DART/pull/1121/changes/f13e3352588ba59890a8250fb952a52fe91561e3](https://github.com/NCAR/DART/pull/1121/changes/f13e3352588ba59890a8250fb952a52fe91561e3)

In order to remove $Id$, I had to also remove the $version stuff. Which was junk anyways cause $Id$ wasn't doing anything for us.

Basically the only real change this causes is instead of putting "# Makefile created by mkmf $Id$ " at the top of the Makefile, it puts "# Makefile created by mkmf" instead. Which is good cause "$Id$" tells us nothing anyways

This is kind of a separate issue though, I can move it to a different PR if needed. Lmk

hkershaw-brown:
awesome thanks Marlee, it might be Friday before I look at this. Cheers, Helen

hkershaw-brown:
@hkershaw-brown build everything modules for Derecho need updating for ifort.  I have this change (in the queue waiting to check it runs)

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -240,13 +183,13 @@ subroutine initialize_mpi_utilities(progname, alternatename, communicator)
 call MPI_Initialized(already, errcode)
 if (errcode /= MPI_SUCCESS) then
    write(*, *) 'MPI_Initialized returned error code ', errcode
-   call exit(-99)
+   stop errcode
```
hkershaw-brown: This is f2018

mpif90 -O2 -std=f2008 -ffree-line-length-none -I/opt/local/include  -c	/Users/hkershaw/DART/pull_requests/fixsystem/pull_1121/assimilation_code/modules/utilities/mpi_utilities_mod.f90
/Users/hkershaw/DART/pull_requests/fixsystem/pull_1121/assimilation_code/modules/utilities/mpi_utilities_mod.f90:186:8:

  186 |    stop errcode
      |        1
Error: STOP code at (1) must be a scalar default CHARACTER or INTEGER constant expression
/Users/hkershaw/DART/pull_requests/fixsystem/pull_1121/assimilation_code/modules/utilities/mpi_utilities_mod.f90:192:11:

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -240,13 +183,13 @@ subroutine initialize_mpi_utilities(progname, alternatename, communicator)
 call MPI_Initialized(already, errcode)
 if (errcode /= MPI_SUCCESS) then
    write(*, *) 'MPI_Initialized returned error code ', errcode
-   call exit(-99)
```
hkershaw-brown, line 243: POSIX -99 wraps around to 157, -999 to 25.  Not sure what the goal of the -99 ( or -999) is. 

```
[hkershaw:test_exit]() > cat test_exit.f90 
program test_exit
implicit none

print*, "hello Helen"
call exit(-99)

end program test_exit
[hkershaw:test_exit]() > gfortran test_exit.f90 
[hkershaw:test_exit]() > ./a.out 
 hello Helen
[hkershaw:test_exit]() > echo $?               
157
[hkershaw:test_exit]() > 
```

observations/obs_converters/NCEP/prep_bufr/convert_bufr/arg_test.f
```diff
@@ -32,7 +32,7 @@ program arg_test
 C       System   - getarg
 C       W3LIB    - errexit
 C
-C   EXIT STATES:
+C   STOP STATES:
```
hkershaw-brown: I'm concerned about making changes to the NCEP/prep_bufr code since this code fails to compile with gfortran
#634

So I think lets not change the prep bufr code in this pull request.

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -1825,23 +1766,19 @@ end function shell_execute
...
 integer,          intent(out) :: rc
 
-! !!NAG_BLOCK_EDIT START COMMENTED_OUT
-!  call system(trim(shell_name)//' '//trim(execute)//' '//char(0), errno=rc)
-! !!NAG_BLOCK_EDIT END COMMENTED_OUT
-! !!OTHER_BLOCK_EDIT START COMMENTED_IN
-    rc = system(trim(shell_name)//' '//trim(execute)//' '//char(0))
-! !!OTHER_BLOCK_EDIT END COMMENTED_IN
+integer :: exitstat
+
```
hkershaw-brown, line 1777: shell_name is now a namelist option that does nothing, but also the mpi_utilites_nml does nothing

https://github.com/NCAR/DART/blob/80502a8e030f5308f557a326819c18cb5079c494/assimilation_code/modules/utilities/mpi_utilities_mod.rst?plain=1#L43-L45

Seems a bit silly to have &mpi_utilities_nml in so many input.nmls

assimilation_code/modules/utilities/mpi_utilities_mod.f90
hkershaw-brown: put in issue #1142

build_templates/mkmf
hkershaw-brown: fine to have in this PR

developer_tests/mpi_utilities/tests/Makefile
```diff
@@ -36,7 +36,6 @@ MPICC = mpicc  # mpi wrapper for cc
 # to comment in or out blocks of code.
```
hkershaw-brown, line 36: The README file for developer_tests/mpi_utilities is out of date.

hkershaw-brown: updated in 9a45ce96a9e6f22d6a963642b732a4252133df3b

assimilation_code/modules/utilities/mpi_utilities_mod.f90
hkershaw-brown: Also shell_name in docs:

https://github.com/NCAR/DART/blob/80502a8e030f5308f557a326819c18cb5079c494/assimilation_code/modules/utilities/mpi_utilities_mod.rst?plain=1#L102-L108

assimilation_code/modules/utilities/mpi_utilities_mod.f90
```diff
@@ -1825,23 +1766,19 @@ end function shell_execute
...
+    call execute_command_line(trim(execute), exitstat=exitstat)
```
hkershaw-brown: Documentation still says shell_execute is a wrapper around  'system'

https://github.com/NCAR/DART/blob/80502a8e030f5308f557a326819c18cb5079c494/assimilation_code/modules/utilities/mpi_utilities_mod.rst?plain=1#L594-L605

hkershaw-brown: Not using the cmdstat to check if the shell process was started.  I think if you don't have cmdstat and it does fail, the program will die without a message.

hkershaw-brown:
> @hkershaw-brown build everything modules for Derecho need updating for ifort. I have this change (in the queue waiting to check it runs)

I put these changes in https://github.com/NCAR/DART/pull/1144. But really I think remove ifort from the build everything developer_test.

assimilation_code/modules/utilities/mpi_utilities_mod.f90
hkershaw-brown: To be fair, exectute_command_line is mostly likely calling system(). 
Not sure if it is better to say wrapper around the fortran intrinsic execute_command_line for a user reading the docs.

hkershaw-brown: added cmdstat in 857739845c4e0a4ecc5dba0f43cfe29da18838ee

assimilation_code/modules/utilities/mpi_utilities_mod.f90
hkershaw-brown: stop -99 is f2008

assimilation_code/modules/utilities/mpi_utilities_mod.f90
mjs2369: @hkershaw-brown should I go ahead and strip out shell_name? In this PR, I removed all instances where it was used, but Ieft the namelist option in for the time being because I didn't want to break people's stuff with an error reading the namelist

assimilation_code/modules/utilities/mpi_utilities_mod.f90
mjs2369: Yeah the -99 seems to be chosen completely arbitrarily. As long as the code is not 0, signifying success, it seems it doesn't really matter what the former exit code, now stop code is. The only thing it changes is how the 
stop prints after terminating the program:
```
masmith@CISL-DUNCAN build_everything % ./test_stop 
 hello Marlee
STOP -99
masmith@CISL-DUNCAN build_everything % vim test_stop.f90             -> changed stop code to -999
masmith@CISL-DUNCAN build_everything % cat test_stop.f90                    
program test_stop
implicit none

print*, "hello Marlee"
stop -999
print*, "hello Marlee"

end program test_stop
masmith@CISL-DUNCAN build_everything % gfortran test_stop.f90 -o test_stop_999                          
masmith@CISL-DUNCAN build_everything % ./test_stop_999 
 hello Marlee
STOP -999        
```

And double-checking this is the behavior in filter:
 ```
--------------------------------------
 Starting ... at YYYY MM DD HH MM SS = 
                 2026  7 14 17  3 52
 Program Filter v11.24.0-20-g4a20d373a-dirty
 --------------------------------------

  set_nml_output Echo NML values to log file only
  initialize_mpi_utilities: Running single process
 
 Assimilate_these_obs_types:
    RAW_STATE_VARIABLE
    RAW_STATE_1D_INTEGRAL
    RAW_STATE_VAR_POWER
 Evaluate_these_obs_types:
    none
 Use the precomputed Prior Forward Operators for these obs types:
    none
 
  quality_control_mod: Will reject obs with Data QC larger than    3
  quality_control_mod: No observation outlier threshold rejection will be done
 
  log_qceff_table_data: Logging the data in the QCEFF Table
  log_qceff_table_data: QCEFF table version: 1
  log_qceff_table_data: QTY_NAME: bounded_below bounded_above lower_bound upper_bound dist_type bounded_below bounded_above lower_bound upper_bound dist_type bounded_below bounded_above lower_bound upper_bound dist_type bounded_below bounded_above lower_bound upper_bound filter_kind bounded_below bounded_above lower_bound upper_bound
  log_qceff_table_data:  QTY_STATE_VARIABLE F F  -888888.00000000000       -888888.00000000000      NORMAL_DISTRIBUTION F F  -888888.00000000000       -888888.00000000000      NORMAL_DISTRIBUTION F F  -888888.00000000000       -888888.00000000000      NORMAL_DISTRIBUTION F F  -888888.00000000000       -888888.00000000000      EAKF F F  -888888.00000000000       -888888.00000000000
 
STOP -99
```

File: assimilation_code/modules/utilities/mpi_utilities_mod.f90 (hunk 12 above)
Comment (mjs2369) [reply assimilation_code/modules/utilities/mpi_utilities_mod.f90]: changed this anywhere it used "stop errcode" or "stop exitcode" to just be "stop -999"

only thing with this is it makes the print it does when it ends the execution is always "STOP -999" instead of the actual error code that is returned from the MPI calls. But it still does print the correct MPI error code in all these instances right before it calls stop.

```
    if (errcode /= MPI_SUCCESS) then
       write(*, *) 'MPI_Init returned error code ', errcode
       stop -999
```

see commit [https://github.com/NCAR/DART/pull/1121/changes/1e3808185d3a618db5620bd4f4c76d444d951252](https://github.com/NCAR/DART/pull/1121/changes/1e3808185d3a618db5620bd4f4c76d444d951252)

assimilation_code/modules/utilities/mpi_utilities_mod.f90
hkershaw-brown: So what is mad is, shell_name is a namelist item, but you have to hand edit the code to read the namelist
in mpi_utilities_mod.f90 or  mpif08_utilties_mod.f90

https://github.com/NCAR/DART/blob/80502a8e030f5308f557a326819c18cb5079c494/assimilation_code/modules/utilities/mpi_utilities_mod.f90#L190-L195

null_mpi_utilities_mod.f90 has the "shell_name is a namelist item" comment 
https://github.com/NCAR/DART/blob/80502a8e030f5308f557a326819c18cb5079c494/assimilation_code/modules/utilities/null_mpi_utilities_mod.f90#L503

**but doesn't actually have a namelist in the module:**
https://github.com/NCAR/DART/blob/80502a8e030f5308f557a326819c18cb5079c494/assimilation_code/modules/utilities/null_mpi_utilities_mod.f90#L124-L125

Then https://github.com/NCAR/DART/issues/1143 there are a bunch of input.nmls that have &mpi_utilities_nml but these are not getting read unless you hand edit the code. I think the input.nml are just copy and paste, copy and paste, copy and paste...generate with AI (copy and paste with more $$$). 

So I would say, leave shell_name in the (non-read) namelist for now, and we'll deal with this mpi/null mpi namelist in another issue (#1143)

mjs2369: @hkershaw-brown cool, this looks like it's ready for approvals then if you agree! I have actually already removed the comment from null_mpi_utilities_mod.f90 in a previous commit:

[https://github.com/NCAR/DART/blob/919d3542166b45f88181c124f91e106e135373c8/assimilation_code/modules/utilities/null_mpi_utilities_mod.f90#L445-L449](https://github.com/NCAR/DART/blob/919d3542166b45f88181c124f91e106e135373c8/assimilation_code/modules/utilities/null_mpi_utilities_mod.f90#L445-L449)

hkershaw-brown: I'm just waiting for my build_everything job to run!

hkershaw-brown (approved):
Approved! Great work on this Marlee!

PR #1144, opened 2026-07-08
edits to compile with ifort on Derecho as of July 2026

hkershaw-brown:
## Description:

edits to compile with ifort on Derecho as of July 2026 for build_everything

The older modules on Derecho appear to be janked up,
I had to set the LD_LIBRARY_PATH and netcdf c and netcdf f libraries in lib64 and lib respectively

I'm not sure whether it is worth having the ifort build in the build_everything since ifort is not supported on Derecho (or by intel). 
But here is the changes need to compile with ifort.  I'm leaning to removing ifort from the list on compilers to build with on Derecho, since I'm not sure how long these old modules will be available.

### DIFF
--- a/developer_tests/build_everything/run_all_quickbuilds.sh
+++ b/developer_tests/build_everything/run_all_quickbuilds.sh
@@ -58,12 +58,21 @@ fi
 
 # mkmf for chosen compiler
 case $compiler in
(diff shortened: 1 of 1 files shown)


mjs2369:
Closing this; removed ifort from build_everything in #1121

PR #705, opened 2024-07-23
Automation of observation converter creation

c-merchant:
## Description:
This PR provides a script and associated infrastructure to automatically generate an observation converter based on the input data format and gets the user started with development, documentation, and building. The previously generalized "text" observation converter was absorbed into this feature with the associated infrastructure kept largely the same. The readme provides condensed documentation for observation converters specifically and from references across other pages in the guide as well as a new step-by-step guide on how to go about creating an observation converter. The existing documentation has been edited for clarity and flow. 

To do: the toctree for the observation converter documentation displays in the webpage, and there may be a .gitignore floating around somewhere. 
### Fixes issue
fixes: #684 

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

### Documentation changes needed?

- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### Tests
I ran the quickbuild files for each data type for the observation converter, worked through a brief implementation based on the available files, and ran the standard ./quickbuild.sh, ./perfect_model_obs, and ./filter for Lorenz-63 to ensure no bugs were introduced.

## Checklist for merging
- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Create release from the main branch with appropriate tag
- [ ] Delete feature-branch

## Testing Datasets
- [ ] Dataset needed for testing available upon request
- [ ] Dataset download instructions included
- [x] No dataset needed

Issue #684, opened 2024-05-29
Feature request: Automate the initial creation of new observation converters


hkershaw-brown:
### Use case  

A user wants to add a new observation converter, where do they start from?

### Is your feature request related to a problem?  

When a user wants to add a new observation converter ([obs_converters](https://github.com/NCAR/DART/tree/main/observations/obs_converters)) to DART they typically copy an existing obs_converter directory. This has a few unwanted side effects:

* Copied code from an existing observation converter is not necessary relevant to the new observation converter (code is messier than it needs to be)
* Users are not necessarily getting the latest code/best practices (creates more work for DART developers during pull requests)
* Copied documentation can be misleading (it looks up to date, but it is for another observation converter)

### Describe your preferred solution  

Create a script that automatically sets up an observation converter for users to start developing and writing their documentation.  I believe with good template files for X_to_obs.f90 we can have a observation converter that compiles from the get-go ( note: this may not be true).

We have a new_model automation, implemented by Benjamin Gunn #376 #349. This is a good reference. 

e.g. for a new converter for the [PACE](https://en.wikipedia.org/wiki/Plankton,_Aerosol,_Cloud,_ocean_Ecosystem) satellite
which uses the threed_sphere location mod and the PACE data is in netcdf:

```
./new_converter PACE threed_sphere netcdf
```

Would give you the starting point:

```
DART/observations/obs_converters/PACE/
                                     pace_to_obs.f90
                                     pace_to_obs.rst
                                     work/
                                          quickdbuild.sh
                                          input.nml
```

### Describe any alternatives you have considered  
I think we just need code that writes code.
There is probably a lot of boiler plate code (e.g. read lat, lon, variable from netcdf) that could be tidied/automated or at least given sensible defaults.

PR #1149, opened 2026-07-13
doc: external forward operator page

### LINKED ISSUE #724: Documentation: Precomputed (external) Forward operator documentation is missing from docs.dart.ucar.edu
### What's the issue? 
:books: Please describe the problem you’ve noticed in the documentation. 

Precomputed Forward operator documentation is missing from docs.dart.ucar.edu

There is some information on this wiki page: 
https://github.com/NCAR/DART/wiki/Precomputed-Forward-Observations-(FOs)

obs_kind_mod.html has the namelist option use_precomputed_FOs_these_obs_types:
https://docs.dart.ucar.edu/en/latest/assimilation_code/modules/observations/obs_kind_mod.html#namelist
but hard to find unless you already know the namelist option. 

the forward operator page of guide, mentions precomputed fields, but not really external forward operators. 
https://docs.dart.ucar.edu/en/latest/guide/forward_operator.html#optional-precomputed-fields

### Where did you find the issue?
Provide the location(s) of the issue:  

user emailed, doc is missing.

### What needs to be fixed?
How to use external (precomputed) FOs.
How to create the obs_seq,
Which namelist options to use in when running filter.

hkershaw-brown:
## Description:

Adds a documentation page on using external forward operators with DART.  
Connects to obs_kind_mod.rst, forward_operator.rst. 
Mentions pyDARTdiags, so added note on pyDARTdiags in pytools. 

### Fixes issue

Fixes #724 

### Types of changes

- [x] Documentation update

### Tests
Just documentation build locally. 

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [x] No dataset needed


hkershaw-brown:
External forward operators are calculated from state before it is inflated.  Not mentioning it in these docs at the moment, but it seems important for users to be aware of this.

mjs2369:
With regards to this warning -

Screenshot 2026-07-29 at 12 25 51 PM

If it will mess up people's experiments, maybe we should make it more explicit that you cannot use inflation when using external forward ops.

Side note, I can't actually find this warning in the code, just on the doc website.

Screenshot 2026-07-29 at 12 30 05 PM

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,257 @@
...
+#. List the observation types that should use the precomputed values in the
+   ``use_precomputed_FOs_these_obs_types`` item of the ``&obs_kind_nml``
+   namelist.
+#. Run ``filter``.
+
+.. warning:: 
+
+   External forward operators are calculated from the ensemble of model states
+   *before* inflation is applied in filter. If you are using inflation, the precomputed 
+   values will not be consistent with the inflated ensemble.
```
hkershaw-brown: @mjs2369 here is the warning in the .rst

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,249 @@
...
+values and can be used as examples:
+
+- :ref:`GSI2DART <gsi2dart>` converts GSI diagnostic files, which contain the
+  expected values computed by GSI for each ensemble member, into a DART
+  observation sequence file.
+- :ref:`convert_roms_obs <roms>` converts ROMS verification observation
+  files, one per ensemble member, into a DART observation sequence file.
+
+If you are writing your own converter, attach the ensemble of expected values
+to each observation with the ``set_obs_def_external_FO`` routine from
```
mjs2369, line 92: It would be great someday to include this info in a doc page on writing your own observation converters for DART. I don't think such a page exists that goes through the necessary routines and such for an obs converter like we have for model_mods

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,249 @@
...
+description of ``&obs_kind_nml``.
+
+Requirements
+~~~~~~~~~~~~
+
+- Every observation of a type listed in
+  ``use_precomputed_FOs_these_obs_types`` must contain precomputed values in
+  the observation sequence file. It is a fatal error if ``filter`` needs a
+  precomputed value that is not present.
+- The number of precomputed values stored with an observation must be at
```
mjs2369, line 213: Will it cause a fatal error if there are more precomputed values than number of ens members? Or will it just ignore them like it does for precomputed values that are in the obs seq but its type is not listed in ``use_precomputed_FOs_these_obs_types``

guide/external-forward-operators.rst
hkershaw-brown: Yup, Charlotte started work on a 'automated' obs converter generator ( same idea as new_model_mod) 
https://github.com/NCAR/DART/issues/684
https://github.com/NCAR/DART/pull/705

However, hot(ish) take I think all the observation converters are essentially the same code and I don't think they need to be Fortran.

guide/external-forward-operators.rst
hkershaw-brown: write a quick test with Lorenz_96 and see if it matches your guess for what happens. 

You are hitting on (one of) the massive design flaw(s) in how external operators are implemented: they deal with an ensemble, but are completely divorced from the ensemble manager.

mjs2369: I did a test using 81 external FOs for an exsemble size of 80, and it didn't error out. This is what I was expecting to happen, despite it not being what is intuitive haha

I also used pyDARTdiags to add external FOs to the obs_seq.out for lorenz_96 for the first time, which worked great btw

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,257 @@
...
+External (precomputed) forward operators
+========================================
+
+Overview
+--------
+
+In data assimilation the process of computing an observation's expected value
+from a model state is called the :ref:`forward operator <FO>` (FO).
+Normally ``filter`` computes the expected values by calling the forward
+operator code for each observation, which interpolates fields from the model
```
braczka: Normally, filter computes the expected values by calling the forward operator code, **internal to DART,** for each observation, which interpolates...

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,257 @@
...
+state.
```
braczka: Also maybe be more specific of what an FO inclues....calling the forward operator code for each observation, **which applies appropriate conversions from fundamental model state variables, followed by a spatial interpolation**.

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,257 @@
...
+
+- Another data assimilation system, such as NOAA's Gridpoint Statistical
+  Interpolation (GSI) system, has already computed expected values for
+  radiances or other observations that require expensive forward operator
+  calculations,
+- The model itself applies its own forward operator as it advances, such as
+  ROMS, which produces its own '*verification*' observations,
+- The forward operator requires model-specific fields or capabilities which
+  are not available in the DART state vector.
+
```
braczka, line 25: For the list of examples, perhaps you should provide the example (or a special case) of radiance models like RTTOV for WRF, where the complexity of the radiance calculation is performed externally within the model RTTOV then passed back into DART, where the spatial interpolation forward operator is applied.   In this case much of the FO is calculated externally to DART, but does not fall under the category of a true external forward operator as described here, as this does not require the expected observation to be inserted directly into the obs_seq file.   Some distinction should be made here of these special examples, maybe put it in the “Note” later on in the documentation.  The advantage of this approach seems to be you can still use inflation…. I think.

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,257 @@
...
+.. warning:: 
+
+   External forward operators are calculated from the ensemble of model states
+   *before* inflation is applied in filter. If you are using inflation, the precomputed 
+   values will not be consistent with the inflated ensemble.
+
+
+Creating an obs_seq file with precomputed forward operators
+-----------------------------------------------------------
+
```
braczka, line 70: Somewhere in this section, it would be helpful to give a workflow of where this calculation of the precomputed forward operator occurs within the typical cycling workflow (forecast 1 --> FO --> filter update --> forecast 2) of a complex model.  For example, because external forward operators take advantage of the dynamic prior ensemble, the obs_seq file with external FO must be created ‘on the fly’.  This is quite different than the traditional obs_seq file which can be prepared well beforehand, and set aside, prior to the assimilation run.   External FO’s require the newly generated forecast ensemble, thus this external FO generation step must be woven into the cycling steps.   Perhaps a simple box/arrow workflow example, or just more specific language of this would be appropriate.

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,257 @@
...
+    OBS            1
+      0.135100000000000
+      0.000000000000000
+             -1           2          -1
+   obdef
+   loc3d
+        4.712388980384690        0.6283185307179586        -500.0000000000000     3
+   kind
+             60
+   external_FO       3       1
```
braczka, line 160: The ‘key’ is this example is ‘1’, which is identical to the observation number ‘1’.   Is this a requirement, or just a coincidence in this case?

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,257 @@
...
+  listed in ``use_precomputed_FOs_these_obs_types``, the values are ignored
+  and the forward operator code is called as usual.
+
+Posterior forward operators and DART quality control
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The precomputed values are *prior* forward operators - they were computed
+from the ensemble of model states before assimilation. There is no way for
+``filter`` to compute a consistent *posterior* forward operator for these
+observations, since the real forward operator code is not being used. If
```
braczka: Technically you could pass the posterior ensemble from DART to your external FO and calculate it directly, and pass it back into DART, right?  You are merely saying the intrinsic DART posterior FO will not function, correct?

guide/external-forward-operators.rst
```diff
@@ -0,0 +1,257 @@
...
+:doc:`../assimilation_code/modules/assimilation/quality_control_mod` for the
+complete list of DART QC values.
+
+Note that the ``obs_seq.final`` file written by ``filter`` does *not* contain
+the ``external_FO`` metadata records. If you request that ensembles of prior
+expected values are output (``num_output_obs_members`` in ``&filter_nml``),
+the precomputed values appear as the ``prior ensemble member`` copies of the
+observations, just as they would for forward operators computed by
+``filter``.
+
```
braczka, line 255: But there is no way to detect whether the prior ensemble member values were calculated internally through DART or through an external FO from the obs_seq.final alone correct?

braczka (changes requested):
I think this looks really good.  I ask for some clarification on the WRF-RTTOV example, which has many characteristics of an external FO, but doesn't use the obs_seq file functionality, so doesn't fall under the category of true FO by this definition.   I also think a bit more emphasis that the added external FO operates dynamically on the forecast ensemble, so it needs to be incorporated within any cycling workflow.  The obs-seq file needs to be updated in real time for this to work.

hkershaw-brown:
> If it will mess up people's experiments, maybe we should make it more explicit that you cannot use inflation when using external forward ops.

People do use inflation when using external FOs. Briefly chatting at the standup about this, my understanding is that it is not clear how 'bad' this is. I will make the docs clearer on 'inflation happens in filter'.

guide/external-forward-operators.rst
hkershaw-brown: added in 472b1d71f

guide/external-forward-operators.rst
hkershaw-brown: added in 472b1d71f

guide/external-forward-operators.rst
hkershaw-brown: added in 472b1d71f

guide/external-forward-operators.rst
hkershaw-brown: added info on this in e93d621ab

guide/external-forward-operators.rst
hkershaw-brown: You could calculate your external FO.  You could pass it back to DART as the prior FO for the next cycle, but filter has no way to ingest posterior forward operators. Added text to describe this in more detail in d882891ee

guide/external-forward-operators.rst
hkershaw-brown: nope, look the same as internal FO.

guide/external-forward-operators.rst
hkershaw-brown: added in a16b1a987

braczka (approved):

PR #1156, opened 2026-07-21
Mom6 read_model_time 

### LINKED ISSUE #959: Feature request: MOM6 Time Handling
### Use case  
Run MOM6-DART experiments using correctly specified model time

Yes. Currently, DART assigns the reference/base time for MOM6 to Year 1. In Kate's experiments for instance MOM6 starts on Jan 1st, 1958 based on the available forcing. So, the time variable that is being read from the restart file should be `days since 01-01-1958`. Note that MOM6 doesn't use an actual calendar. Simulations start from simply 1. 

I discussed this with Ian. After checking, it seems like it won't be possible to add in additional time attributes (because of limitations within the CESM framework). However, Ian and Kate know the reference date of their problem, i.e., 01-01-1958. So, my proposed solution is to add two namelist parameters to `&model_nml`: [1] `reference_date` and [2] `calendar_type`. These can be defined by the user depending on their specific applications. In Kate's case, these can be set to: `reference_date = 01-01-1958` and `calendar_type = NOLEAP`. Internally, DART can parse the calendar's type in `static_init_model` and the reference date in `read_model_time`. 

We tried enforcing the starting time using `init_time_days` and `init_time_seconds` but that did't seem to solve the problem. The filter wanted to advance the model, probably because the observation sequence files had observations beyond the assimilation window (but not sure).

hkershaw-brown:
## Description:

MOM6 read_model_time

* Allow fractions of days in the time, (variable in the file is real not int)
* Allow time dimension to be called 'Time' or 'time' (restart vs history)
* Allow user to set a reference date different to 0001-01-01:00:00:00

Note the reference data has to be after darts own reference date.
This is for ease of setting the date really.

### Fixes issue

fixes https://github.com/NCAR/DART/issues/959
fixes https://github.com/CROCODILE-CESM/dartobsgen/issues/6

### Types of changes

- [ ] Documentation update

### Tests
Read model time from netcdf files

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/models/MOM6/model_mod.f90
+++ b/models/MOM6/model_mod.f90
@@ -12,7 +12,8 @@ module model_mod
 
 use        types_mod, only : r8, i8, MISSING_R8, vtablenamelength
 
-use time_manager_mod, only : time_type, set_time, set_date, set_calendar_type
+use time_manager_mod, only : time_type, set_time, set_date, set_calendar_type, &
+                             get_time
 
 use     location_mod, only : location_type, get_close_type, &
                              loc_get_close_obs => get_close_obs, &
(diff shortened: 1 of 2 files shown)


hkershaw-brown:
Not sure why your comment is not showing up on github.
Yup mom6 has time as a real (double) number:

```
 ncdump -v time ./Observations/TestTutorials/dartobsgen/example_mom6/
EEP_MITgcm185Lvgrid_Whitt2026hgrid.mom6.h.z.2015-09-004.nc
...
        double time(time) ;
                time:units = "days since 0001-01-01 00:00:00" ;
                time:long_name = "time" ;
                time:axis = "T" ;
                time:calendar = "gregorian" ;
                time:bounds = "time_bounds" ;

...
data:

 time = 735844.5 ;
}
```

mjs2369:
> Not sure why your comment is not showing up on github. Yup mom6 has time as a real (double) number: 

I had removed the comment because I knew the answer, just thought it was strange. But then I thought better to not crowd the PR with semi-random comments

hkershaw-brown:
> > Not sure why your comment is not showing up on github. Yup mom6 has time as a real (double) number:
> 
> I had removed the comment because I knew the answer, just thought it was strange. But then I thought better to not crowd the PR with semi-random comments

It is a really good comment.  Always worth bring up things that seems strange.

mjs2369 (approved):
Hi @hkershaw-brown this looks great to me. I went through the logic you added to work with the user defined reference_date. I used an example ref date of 1801-01-01 and a 'Time' value corresponding to 2001-01-01, and I got the right answer for number of days since the dart base (1601-01-01) - 146097

Approved

Issue #277, opened 2021-08-27
note on inflation, allow_missining_in_clm, and groups


hkershaw-brown:
This is note about missing values.
allow_missing_in_clm is in the inflation handle. 

filter:
```
1649 do j = 1, ens_handle%my_num_vars
1650      call inflate_ens(inflate, ens_handle%copies(grp_bot:grp_top, j), &
1651      ens_handle%copies(ENS_MEAN_COPY, j), ens_handle%copies(inflate_copy, j), 0.0_r8, &
1652       ens_handle%copies(SPARE_PRIOR_SPREAD, j), ens_handle%copies(ENS_SD_COPY, j)) 
1653 end do 
```

inflate_ens:
```
532 ! it's possible to have MISSING_R8s in the state vector now.  
533 ! so we need to be able to avoid changing MISSING_R8 values by inflation here.
534 if (inflate_handle%allow_missing_in_clm) then
535   if (any(ens == MISSING_R8)) return
endif
```
So if some groups have missing values, but some don't, you will apply inflation to some groups but not others.

nancycollins:
in looking at the inflation code i believe for adaptive inflation kinds 2 and 5 that the ```update_inflation()``` routine is called for each group, so the adaptive inflation values are updated N times if you have N groups, which may be the intended algorithm.  i haven't figured out yet if the RTPS implementation took into account num_groups > 1.

hkershaw-brown:
note on missing_8 footprint in quad_utils_mod #249

hkershaw-brown:
models calling set_missing_ok_status as of Jan 6th 2022:

../../../models/noah/model_mod.f90:252:call set_missing_ok_status(.true.)
../../../models/clm/model_mod.f90:484:call set_missing_ok_status(.true.)

`call set_missing_ok_status` is called in `static_init_model` and then overwritten by the call in filter. 

```
filter_mod.f90:
367 call filter_initialize_modules_used() ! static_init_model called in here

369 ! Read the namelist entry
     call find_namelist_in_file("input.nml", "filter_nml", iunit)
     read(iunit, nml = filter_nml, iostat = io)
     call check_namelist_read(iunit, io, "filter_nml")

...

399 call set_missing_ok_status(allow_missing_clm)  ! this call to set_missing_ok_status overwrites whatever was set in static_init_model
400 allow_missing = get_missing_ok_status()
```

This is not so good.

nancycollins:
[...]

PR #1157, opened 2026-07-27
Localization test perturbations

### LINKED ISSUE #967: Feature request: alternative perturbs & increments
From standup discussion 4th Sept 2025 - please edit/add as needed ☠️ 

### Use case  
What are you trying to accomplish?  Sometimes there may already be a way to do what you need.  

Testing where you want different ensemble members but not with Gaussian noise.
e.g. constant value added *ensemble member so so you can check model_interpolate for each ensemble member. 

Also localization - increment with a constant so you can test get_close. 

Ease of testing. 

Please provide enough detail here to help people have a discussion about your proposal
Not yet. 

None yet.

jlaucar:
Modified filter_mod.f90 so that it can generate uniform ensemble perturbations to every state variable. This means that the covariance of an observation is the same with every state variable. The only think that leads to non-uniform increments is the localization, so it can be easily displayed and studied. Changes include a new logical nameless entry for filter called perturb_for_localization_test. If this and perturb_from_single_instance are both true, the new perturbation method is applied. The perturbations are computed in the new subroutine perturb_copies_uniform.

This is a new feature which adds functionality and has no impact if the new nameless parameter is false (default). The filter.rst documentation has been updated.

The new method was tested with single observations for CAM, CLM, and Lorenz-96, and in strongly coupled tests where a single cam observation impacts CLM and a single CLM observation impacts CAM. Visual inspection of the results shows the localization patterns clearly.

fixes #967

### DIFF
--- a/assimilation_code/modules/assimilation/filter_mod.f90
+++ b/assimilation_code/modules/assimilation/filter_mod.f90
@@ -64,7 +64,7 @@ use adaptive_inflate_mod,  only : do_ss_inflate, mean_from_restart, sd_from_rest
 use mpi_utilities_mod,     only : my_task_id, task_sync, broadcast_send, broadcast_recv,      &
                                   task_count
 
-use perturb_mod,           only : perturb_ensemble, PERT_MODEL_MOD, PERT_UNIFORM
+use perturb_mod,           only : perturb_ensemble
 
 use state_vector_io_mod,   only : state_vector_io_init, read_state, write_state, &
                                   set_stage_to_write, get_stage_to_write
@@ -205,9 +205,12 @@ character(len=256) :: output_state_file_list(MAX_NUM_DOMS) = ''
 ! Read in a single file and perturb this to create an ensemble
 logical  :: perturb_from_single_instance = .false.
 real(r8) :: perturbation_amplitude       = 0.2_r8
-! Perturb all variables uniformly for clean test of observation localization
-! Must have perturb_from_single_instance true also to do this
-logical  :: perturb_for_localization_test = .false.
+! How to perturb: 
+!    'model' lets the model_mod do it, falling back to a
+! perturbation that is bitwise across any number of tasks.
+!    'uniform' perturbs every state variable identically, used when testing
+! observation localization.  
+character(len=32) :: perturbation_method = 'model' !  Ignored unless perturb_from_single_instance.
 
 ! File options.  Single vs. Multiple.  really 'unified' or 'combination' vs 'individual'
 logical  :: single_file_in  = .false. ! all copies read  from 1 file
@@ -300,7 +303,7 @@ namelist /filter_nml/ async,     &
    single_file_out,              &
    perturb_from_single_instance, &
    perturbation_amplitude,       &
-   perturb_for_localization_test,&
+   perturbation_method,          &
    compute_posterior,            &
    stages_to_write,              &
    input_state_files,            &
@@ -346,7 +349,6 @@ integer                 :: OBS_VAR_START, OBS_VAR_END, TOTAL_OBS_COPIES
 integer                 :: input_qc_index, DART_qc_index
 integer                 :: num_state_ens_copies
 logical                 :: read_time_from_file
-integer                 :: pert_method ! how to perturb a single instance
 
 integer :: num_extras ! the extra ensemble copies
 
@@ -386,19 +388,6 @@ call timestamp_message('Filter start')
 if(ens_size < 2) then
    write(msgstring, *) 'ens_size in namelist is ', ens_size, ': Must be > 1'
    call error_handler(E_ERR,'filter_main', msgstring, source)
-endif
-
-! perturb_for_localization_test can only be true if perturb_from_single_instance is true
-if(perturb_for_localization_test .and. .not. perturb_from_single_instance) then
-   write(msgstring, *) 'perturb_for_localization_test cannot be true unless  &
-      perturb_from_single_instance is also true in filter_nml'
-   call error_handler(E_ERR,'filter_main', msgstring, source)
-endif
-
-if(perturb_for_localization_test) then
-   pert_method = PERT_UNIFORM
-else
-   pert_method = PERT_MODEL_MOD
 endif
 
 ! informational message to log
@@ -601,7 +590,7 @@ if (perturb_from_single_instance) then
 
    ! Only zero has the time, so broadcast the time to all other copy owners
    call broadcast_time_across_copy_owners(state_ens_handle, time1)
-   call perturb_ensemble(state_ens_handle, ens_size, perturbation_amplitude, pert_method)
+   call perturb_ensemble(state_ens_handle, ens_size, perturbation_amplitude, perturbation_method)
 else
    call error_handler(E_MSG,'filter_main:', &
       'Reading in initial condition/restart data for all ensemble members from file(s)')
--- a/assimilation_code/modules/assimilation/filter_mod.f90
+++ b/assimilation_code/modules/assimilation/filter_mod.f90
@@ -205,6 +205,9 @@ character(len=256) :: output_state_file_list(MAX_NUM_DOMS) = ''
 ! Read in a single file and perturb this to create an ensemble
 logical  :: perturb_from_single_instance = .false.
 real(r8) :: perturbation_amplitude       = 0.2_r8
+! Perturb all variables uniformly for clean test of observation localization
+! Must have perturb_from_single_instance true also to do this
+logical  :: perturb_for_localization_test = .false.
 
 ! File options.  Single vs. Multiple.  really 'unified' or 'combination' vs 'individual'
 logical  :: single_file_in  = .false. ! all copies read  from 1 file
@@ -297,6 +300,7 @@ namelist /filter_nml/ async,     &
    single_file_out,              &
    perturb_from_single_instance, &
    perturbation_amplitude,       &
+   perturb_for_localization_test,&
    compute_posterior,            &
    stages_to_write,              &
    input_state_files,            &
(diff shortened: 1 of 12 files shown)


assimilation_code/modules/assimilation/filter_mod.nml
```diff
@@ -6,6 +6,7 @@
    init_time_seconds            = 0,
    perturb_from_single_instance = .false.,
    perturbation_amplitude       = 0.2,
+   perturbation_locatlization_test = .false.,
```
hkershaw-brown: the namelist variable is called perturb_for_localization_test in the code

assimilation_code/modules/assimilation/filter_mod.rst
```diff
@@ -33,6 +33,7 @@ namelist.
       init_time_seconds            = 0,
       perturb_from_single_instance = .false.,
       perturbation_amplitude       = 0.2,
+      perturbation_localization_test = .false.,
```
hkershaw-brown: the namelist variable is called perturb_for_localization_test in the code

assimilation_code/modules/assimilation/filter_mod.rst
```diff
@@ -145,7 +146,24 @@ prior inflation and the second controls the posterior inflation.
 |                              |                     | ``pert_model_copies``. For more,          |
 |                              |                     | see pert_model_copies_  below.            |
 |                              |                     | Ignored if                                |
-|                              |                     | ``perturb_from_single_instance = .false.``| 
+|                              |                     | ``perturb_from_single_instance= .false.`` | 
++------------------------------+---------------------+-------------------------------------------+
+| perturbation_amplitude       | logical             | Can only be true if                       |
```
hkershaw-brown: these docs have perturbation_amplitude listed twice (once a logical)

assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -392,6 +396,13 @@ subroutine filter_main()
 
 call trace_message('Before initializing inflation')
 
+! perturb_for_localization_test can only be true if perturb_from_single_instance is true
+if(perturb_for_localization_test .and. .not. perturb_from_single_instance) then
+   write(msgstring, *) 'perturb_for_localization_test cannot be true unless  &
+      perturb_from_single_instance is also true in filter_nml'
+   call error_handler(E_ERR,'filter_main', msgstring, source)
+endif  
+
```
hkershaw-brown: I'm not quite following the need to error out here. 
If I turn off perturb_from_single_instance why do I also need to turn off dependent namelists?, should I set the perturbation_amplitude to zero also?

hkershaw-brown: This check is in the 
"Before Initializing inflation" section so the error looks like an inflation initialization error.: 

```
 PE 0:  filter trace: Before initializing inflation
 ERROR FROM:
  source : filter_mod.f90
  routine: filter_main
  message:  perturb_for_localization_test cannot be true unless  perturb_from_single_instance is also true in filter_nml
 
```

assimilation_code/modules/assimilation/filter_mod.f90
hkershaw-brown: There is no message here that you are perturbing using the localization test perturb.

assimilation_code/modules/assimilation/filter_mod.f90
```diff
@@ -2172,6 +2189,27 @@ subroutine perturb_copies_task_bitwise(ens_handle)
 
 end subroutine perturb_copies_task_bitwise
 
+!------------------------------------------------------------------
+! Perturb the copies array uniformly for localization tests
+! The covariance between an observation and any state variable should be
+! the same.
+
+subroutine perturb_copies_uniform(ens_handle)
```
hkershaw-brown: Question on why this is filter_mod rather than its own perturb module? Is the reason the other perturb routine was already in filter?

Some perturbation logic is duplicated in assimilation_code/programs/perturb_single_instance/perturb_single_instance.f90 - this standalone does not respect missing_r8 in the state, so we already have codes drifting away from each other. 

I think this is a good time to make a perturb module rather than adding more routines to filter_mod.  Particularly with something like perturbs that would be used for testing without a full filter run: 
Potentially other 'perturbs' that you might want for tests outside filter e.g. make 3 ensemble members the same or other weird math stuff.   Add some missing_r8s in some ensemble members only, see if your code catches it.  Switch out NaNs for missing_r8s.  Perturb but respect qceff bounds.

hkershaw-brown (changes requested):
Hi Jeff,

This is nice, 
You're documentation is a bit messed up, I put comments in where the docs don't match the code. 
I think perturb should go in its own module. (see comment on perturb_single_instance code drift + other perturbs)
Small comments/question on printing a message in filter that the perturb is 'localization test mode' & where the message prints.

hkershaw-brown:
Also this is more a note for myself than anything, with lorenz_96 out-of-the box the first observation time is 3600 0 (which I grabbed for my one-obs test), and the input.nml sets the 
 init_time_days               = 0,
 init_time_seconds            = 0,

so there is a uniform pertrub, then model advance, so the uniform is not so uniform. 
I think I will not be the only dum-dum to chase my tail on this - maybe worth putting in the docs for testing localization?

hkershaw-brown:
Re standup yesterday, @jlaucar does not have time to put perturb in a module, so I will go ahead and get that done today.

assimilation_code/modules/assimilation/perturb_mod.f90
```diff
@@ -0,0 +1,219 @@
...
+
+subroutine perturb_bitwise(copies, ens_size, my_vars, num_vars, amplitude)
+
+real(r8),    intent(inout) :: copies(:,:) ! (num_copies, my_num_vars)
+integer,     intent(in)    :: ens_size
+integer(i8), intent(in)    :: my_vars(:)  ! global index of each local element
+integer(i8), intent(in)    :: num_vars    ! global number of state elements
+real(r8),    intent(in)    :: amplitude
+
+integer(i8)           :: i ! global state index
```
hkershaw-brown, line 177: note i8, this was i4 on main,  so would have problems for large state vectors:

https://github.com/NCAR/DART/blob/ab4a62cfdcff107005f6c9b508035c7dfb6d6ca5/assimilation_code/modules/assimilation/filter_mod.f90#L2144-L2157

assimilation_code/modules/assimilation/filter_mod.rst
hkershaw-brown: fixed in a79ed8dbfdb39399c31d318bac7d00a470f48eca

assimilation_code/modules/assimilation/filter_mod.rst
hkershaw-brown: fixed in a79ed8dbfdb39399c31d318bac7d00a470f48eca

assimilation_code/modules/assimilation/filter_mod.f90
hkershaw-brown: Put perturbaiton method in message printed when perturbing a single instance

```
Reading in a single member and perturbing data for the other ensemble members'
  ... perturbation_method =  blah
```

assimilation_code/modules/assimilation/filter_mod.f90
hkershaw-brown: moved to module in 8ee1138dabdb176a3e5c9dafac26850d266ff8b6

assimilation_code/modules/assimilation/filter_mod.f90
hkershaw-brown: Put perturbaiton method in message printed when perturbing a single instance

```
Reading in a single member and perturbing data for the other ensemble members'
  ... perturbation_method =  blah
```

in ae77d2f7a14689834306d190983998764d879763

hkershaw-brown:
> Also this is more a note for myself than anything, with lorenz_96 out-of-the box the first observation time is 3600 0 (which I grabbed for my one-obs test), and the input.nml sets the init_time_days = 0, init_time_seconds = 0,
> 
> so there is a uniform pertrub, then model advance, so the uniform is not so uniform. I think I will not be the only dum-dum to chase my tail on this - maybe worth putting in the docs for testing localization?

Maybe this info goes in this section:
https://docs.dart.ucar.edu/en/latest/assimilation_code/modules/assimilation/filter_mod.html#create-an-initial-ensemble-from-a-single-file

https://github.com/NCAR/DART/blob/ab4a62cfdcff107005f6c9b508035c7dfb6d6ca5/assimilation_code/modules/assimilation/filter_mod.rst?plain=1#L413

assimilation_code/modules/assimilation/perturb_mod.f90
```diff
@@ -0,0 +1,219 @@
...
+!>               pert_model_copies.  If it does not provide that interface,
+!>               perturb_bitwise is used instead.  Note: not enforcing a
+!>               model_mod to produce a pert_model_copies that is bitwise
+!>               across any number of tasks, although there is enough
+!>               information in the ens_handle to do this.
+!>   'uniform' - perturb_uniform, for localization tests.
+!>
+!> Some models allow missing_r8 in the state vector.  If missing_r8 is
+!> allowed the locations of missing_r8s are stored before the perturb,
+!> then the missing_r8s are put back in after the perturb.
```
hkershaw-brown, line 66: note on missing_r8

`get_missing_ok_status()` `allow_missing_in_clm` are typically used for models where only some of the ensemble members are missing_r8 vs. models (e.g. ocean) models where the state is missing_r8 across the ensemble (land).

Pics:
https://github.com/NCAR/DART/issues/277#issuecomment-1010142449

jlaucar (commented):
I've reviewed Helen's modifications of the original pull request. Tested on L96 and CAM-FV. Tested the test routines Everything looks good. I think this is ready to go.

guide/instructions-for-porting-a-new-model-to-dart.rst
```diff
@@ -293,3 +293,34 @@ item will contain the changed values.  If your model is reading NetCDF format
 it can ingest these directly.  If not, an additional step is needed to copy
 over the updated values for the next model run.
 
+
+Testing Localization for a New Model or New Forward Operator
```
hkershaw-brown: This makes sense for testing localization for a new model.

I think it is a bit misleading to have  'or new forward operator', as if "localization looks good, forward operator is correct". 

```suggestion
Testing Localization Using a Single Observation and An Idealized Ensemble
```

Also for lazy people who only read headings, they will get the point of what to do to test. 

Otherwise looks good.

hkershaw-brown (approved):

PR #1165, opened 2026-08-13
use obs_ens_handle for get_var_owner_index and broadcast in seq_obs do loop

### LINKED ISSUE #1163: bug: Use of state ens_handle instead of obs_ens_handle in SEQUENTIAL_OBS loop in assume_tools_mod.f90
The following code block occurs (approximately line 616 in current main) in assim_tools_mod.f90.
   ! Find out who has this observation and where it is
   call get_var_owner_index(ens_handle, int(i,i8), owner, owners_index)

   ! Following block is done only by the owner of this observation
   !-----------------------------------------------------------------------
   if(ens_handle%my_pe == owner) then
      ! each task has its own subset of all obs.  if they were converted in the

I believe that the two instances of 'ens_handle' should be 'obs_ens_handle'. Since we almost always use the default layout in ensemble_managers, this is not normally an issue. If someone used different layouts for the obs ensemble, this would cause incorrect behavior.

This code is from DART main retrieved on 11 August 2026, but it appears to have existed pretty much forever in assim_tools_mod

hkershaw-brown:
## Description:

Use obs_ens_handle for get_var_owner_index and broadcast rather than ens_handle
Note the get_var_owner_index does not use ens_handle passed to it. 

map_pe_to_task _does_ use the ensemble handle, which is used in the broadcast.  You could have set a different layout (mapping of physical to logical tasks) when you initialize the ensemble handles. There is a note in ensemble manager that you should no do this.

https://github.com/NCAR/DART/blob/b270ca69bee4915141f0948043c64d5c9b263e5a/assimilation_code/modules/utilities/ensemble_manager_mod.f90#L165-L172

Q. Why even have layout as an option, many parts of the code assume the layout is the same for ens, obs_ens, and qc_obs_ens?   "Do not shoot your self in the foot with this foot gun we have provided"

### Fixes issue

fixes #1163

### Types of changes

- [ ] Documentation update

I do not think this bug has caused problems, however if someone did pass different layouts to the various ensemble initializations this would cause problems. 

### Tests
Bitwise lorenz_96 with mpi (4,5, procs my mac) & serial

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/assimilation_code/modules/assimilation/assim_tools_mod.f90
+++ b/assimilation_code/modules/assimilation/assim_tools_mod.f90
@@ -614,11 +614,11 @@ SEQUENTIAL_OBS: do i = 1, obs_ens_handle%num_vars
    call get_obs_values(observation, obs, obs_val_index)
 
    ! Find out who has this observation and where it is
-   call get_var_owner_index(ens_handle, int(i,i8), owner, owners_index)
+   call get_var_owner_index(obs_ens_handle, int(i,i8), owner, owners_index)
 
    ! Following block is done only by the owner of this observation
    !-----------------------------------------------------------------------
-   if(ens_handle%my_pe == owner) then
+   if(obs_ens_handle%my_pe == owner) then
       ! each task has its own subset of all obs.  if they were converted in the
       ! vertical up above, then we need to broadcast the new values to all the other
       ! tasks so they're computing the right distances when applying the increments.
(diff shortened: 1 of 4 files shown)


assimilation_code/modules/assimilation/assim_tools_mod.f90
```diff
@@ -614,11 +614,11 @@ subroutine filter_assim(ens_handle, obs_ens_handle, obs_seq, keys,           &
    call get_obs_values(observation, obs, obs_val_index)
 
    ! Find out who has this observation and where it is
-   call get_var_owner_index(ens_handle, int(i,i8), owner, owners_index)
+   call get_var_owner_index(obs_ens_handle, int(i,i8), owner, owners_index)
```
mgharamti, line 617: I'm wondering whether we should remove the first argument (handle) all together. Clearly, the subroutine doesn't use it. Or is this related to the distribution type being set to 1?

mgharamti (commented):
Helen, this looks good. My only comment is in regard to the unused ensemble handle when returning the PE for the var number.

assimilation_code/modules/assimilation/assim_tools_mod.f90
hkershaw-brown: you're exactly right Moha, with the code as is it would make sense to remove this unused argument, however there is some open discussion on the distibution type (#1166).

hkershaw-brown:
> Helen, this looks good. My only comment is in regard to the unused ensemble handle when returning the PE for the var number.

Hi Moha, are you ok with approving this for the next release? The unused ensemble handle is still hanging out there (to be address in issue #1166)

mgharamti:
sounds good. Let's put it out and revisit the other ununsed argument stuff later.

mgharamti (approved):

PR #1181, opened 2026-08-28
MOM6 model_md fix: u,v grid may have different size to t grid

### LINKED ISSUE #1178: bug: Regional MOM6 assuming geolon|u|v are the same size
:bug:  🐳 
### Describe the bug

    Create a regional case (Hawaii) and run filter 
    run ok  
3. What actually happened?  
    wacky longitude during assimilation
    PE 0: filter: Ready to assimilate up to 3956 observations
    ERROR FROM:
    ERROR FROM:
    ERROR FROM:
    source : threed_sphere/location_mod.f90
    routine: set_location
    message: longitude ( 1.926455913427073E+185 ) is not within range [0,360]
 
### Error Message  
Please provide any error messages.

```
PE 0: filter: Ready to assimilate up to 3956 observations
    ERROR FROM:
    ERROR FROM:
    ERROR FROM:
    source : threed_sphere/location_mod.f90
    routine: set_location
    message: longitude ( 1.926455913427073E+185 ) is not within range [0,360]
 ```

or if you have debugging flags on: 
```
forrtl: severe (408): fort: (2): Subscript #1 of the array GEOLON_U has value 101 which is greater than the upper bound of 100
```

Regional MOM6 via Crocodile CESM

### Version of DART      
 v11.25.1-dirty
Yes  I have but only mkmf for parallel builds (hence -dirty)

### Build information
Please describe:  
 1. Derecho
 2. ifx (IFX) 2025.2.1 20250806

This code is a bad assumption in the regional (and maybe also in the global sometimes? though for the CESM global cases they are the same size)

https://github.com/NCAR/DART/blob/b270ca69bee4915141f0948043c64d5c9b263e5a/models/MOM6/model_mod.f90#L715-L723

Not hitting this in PMO (which is mostly what people have been doing with regional) but you would miss (fail) obs that are at the edge of the domain.  Note this is separate issue to #1124.

### PR DESCRIPTION
**Pull request by hkershaw-brown**
## Description:

MOM6 model_md fix: u,v grid may have different size to t grid for regional domains (extra grid point in x or y). The code was assuming these grids are the same size, which is true for the cesm global grids, but is not generally true. 

This pull request makes geolon t|u|v arrays different to allow for the different grid sizes.  

Since wet (ocean or land) and D (basin depth) are _only_ defined on the t grid, I have assumed that if the tlon,tlat point is not ocean then the u,v points are not with min(ilon, nx)

### Fixes issue

fixes #1178

Note I believe this issue will affect regional cice (#1167) since cice is on the same grid but I have not addressed cice in this pull request. 

### Types of changes

- [ ] Documentation update

### Tests
Running mom6 regional Hawaii case. 

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed

### DIFF
--- a/models/MOM6/model_mod.f90
+++ b/models/MOM6/model_mod.f90
@@ -97,7 +97,7 @@ integer :: dom_id ! used to access the state structure
 integer(i8) :: model_size
 integer :: nfields ! number of fields in the state vector
 ! Grid parameters, nz is number of layers
-integer :: nx=-1, ny=-1, nz=-1     ! grid counts for each field
+integer :: nx=-1, ny=-1, nx_u=-1, ny_u=-1, nx_v=-1, ny_v=-1, nz=-1     ! grid counts for each field
 real(r8), allocatable :: geolon(:,:), geolat(:,:),     & ! T
                          geolon_u(:,:), geolat_u(:,:), & ! U
                          geolon_v(:,:), geolat_v(:,:)    ! V
@@ -712,22 +712,28 @@ character(len=*), parameter :: routine = 'read_horizontal_grid'
 
 ncid = nc_open_file_readonly(static_file)
 
+! T
 call nc_get_variable_size(ncid, 'geolon', nxy)
-nx = nxy(1)
-ny = nxy(2)
+nx = nxy(1); ny = nxy(2)
 allocate(geolon(nx,ny), geolat(nx,ny))      ! T grid
-allocate(geolon_u(nx,ny), geolat_u(nx,ny))  ! U grid
-allocate(geolon_v(nx,ny), geolat_v(nx,ny))  ! V grid
 allocate(mask(nx,ny))  ! missing values
-allocate(mask_u(nx,ny))  ! missing values
-allocate(mask_v(nx,ny))  ! missing values
-
 call nc_get_variable(ncid, 'geolon', geolon, routine)
-call nc_get_variable(ncid, 'geolon_u', geolon_u, routine)
-call nc_get_variable(ncid, 'geolon_v', geolon_v, routine)
-
 call nc_get_variable(ncid, 'geolat', geolat, routine)
+
+! U 
+call nc_get_variable_size(ncid, 'geolon_u', nxy)
+nx_u = nxy(1); ny_u = nxy(2)
+allocate(geolon_u(nx_u,ny_u), geolat_u(nx_u,ny_u))  ! U grid
+allocate(mask_u(nx_u,ny_u))  ! missing values
+call nc_get_variable(ncid, 'geolon_u', geolon_u, routine)
 call nc_get_variable(ncid, 'geolat_u', geolat_u, routine)
+
+! V
+call nc_get_variable_size(ncid, 'geolon_v', nxy)
+nx_v = nxy(1); ny_v = nxy(2)
+allocate(geolon_v(nx_v,ny_v), geolat_v(nx_v,ny_v))  ! V grid
+allocate(mask_v(nx_v,ny_v))  ! missing values
+call nc_get_variable(ncid, 'geolon_v', geolon_v, routine)
 call nc_get_variable(ncid, 'geolat_v', geolat_v, routine)
 
 ! mom6 has missing values in the grid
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: models/MOM6/model_mod.f90 [hunk 1]
```diff
@@ -712,22 +712,28 @@ subroutine read_horizontal_grid()
...
-allocate(mask_v(nx,ny))  ! missing values
-
 call nc_get_variable(ncid, 'geolon', geolon, routine)
-call nc_get_variable(ncid, 'geolon_u', geolon_u, routine)
-call nc_get_variable(ncid, 'geolon_v', geolon_v, routine)
-
 call nc_get_variable(ncid, 'geolat', geolat, routine)
+
+! U 
+call nc_get_variable_size(ncid, 'geolon_u', nxy)
```
masmith99, line 724: Double checking that nc_get_variable_size is definitely overwriting the values in nxy, correct? i.e this is equivalent to having a separate array for u and v (nxy_u, nxy_v)?

masmith99 (commented):
https://github.com/NCAR/DART/blob/02d74698da5d0a404629fb35690377b7d3c92bcf/models/MOM6/model_mod.f90#L823-L827

I'm not quite understanding this logic. I get the assumption that if nx is wet, nx_u/v are also wet

But I don't understand why this translates into the clamps on ilon_t and ilat_t:
`ilon_t = min(ilon, nx)`

models/MOM6/model_mod.f90
```diff
@@ -712,22 +712,28 @@ subroutine read_horizontal_grid()
...
+allocate(mask_u(nx_u,ny_u))  ! missing values
+call nc_get_variable(ncid, 'geolon_u', geolon_u, routine)
 call nc_get_variable(ncid, 'geolat_u', geolat_u, routine)
+
+! V
+call nc_get_variable_size(ncid, 'geolon_v', nxy)
+nx_v = nxy(1); ny_v = nxy(2)
+allocate(geolon_v(nx_v,ny_v), geolat_v(nx_v,ny_v))  ! V grid
+allocate(mask_v(nx_v,ny_v))  ! missing values
+call nc_get_variable(ncid, 'geolon_v', geolon_v, routine)
```
hkershaw-brown, line 736: @hkershaw-brown you have `call nc_get_variable(ncid, 'geolon_v', geolon_v, routine)` twice

models/MOM6/model_mod.f90
hkershaw-brown: Good double check.

nxy(2)  is s length 2 array to collect the x,y lengths of the variable
nxy(2) is local to this routine. 

The module variables 
nx, ny
nx_u, ny_u
nx_v, ny_v
are the x,y lengths  of T,U,V grid.

hkershaw-brown:
> https://github.com/NCAR/DART/blob/02d74698da5d0a404629fb35690377b7d3c92bcf/models/MOM6/model_mod.f90#L823-L827
> 
> I'm not quite understanding this logic. I get the assumption that if nx is wet, nx_u/v are also wet
> 
> But I don't understand why this translates into the clamps on ilon_t and ilat_t: `ilon_t = min(ilon, nx)`


ilon can be nx+1 (for non-t grids)
ilat can be ny+1  (for non-t grids)

but the wet and mask grid only extends 1-nx, 1-ny so if you try and access nx+1 or ny+1 you will be out of bounds.

T grid indices:

```
ilon_t = min(ilon, nx)  ! will be ilon unless ilon > nx, then it will be nx 
ilat_t = min(ilat, ny)    ! will be ilat unless ilat > y, then  it will ny
```

There is no periodic domains in MOM6 so I'm not worrying about any wrapping:  nx+1 is not 1

Does this make sense?

masmith99:
> ilon can be nx+1 (for non-t grids) ilat can be ny+1 (for non-t grids)
> 
> but the wet and mask grid only extends 1-nx, 1-ny so if you try and access nx+1 or ny+1 you will be out of bounds.
> 
> T grid indices:
> 
> ```
> ilon_t = min(ilon, nx)  ! will be ilon unless ilon > nx, then it will be nx 
> ilat_t = min(ilat, ny)    ! will be ilat unless ilat > y, then  it will ny
> ```
> 
> There is no periodic domains in MOM6 so I'm not worrying about any wrapping: nx+1 is not 1
> 
> Does this make sense?

Yes, perfectly. Thanks for the explanation.

models/MOM6/model_mod.f90
masmith99: Don't think you do? Second call is for geolat_v

hkershaw-brown: man I really struggle with reading

masmith99 (approved):
Ready for release!

PR #1187, opened 2026-09-02
fix: non-cppflags version of adding the git version string to dart

### LINKED ISSUE #1164: bug: Build failure with mpiifort: DART_VERSION macro loses quotes in version_mod.F90
### Describe the bug

When building DART with Intel Fortran/Intel MPI, `preprocess` builds successfully using `ifort`, but subsequent MPI programs fail to compile with `mpiifort`.

The failure occurs while compiling `build_templates/version_mod.F90`. The source contains:

```fortran
character(len=*), parameter :: dart_version_string = DART_VERSION
```

and the DART build script defines the version macro using:

```bash
version_def="-DDART_VERSION="'$git_version'""
```

For my checkout,

```bash
git describe --tags --dirty --always
```

returns:

```text
v11.1.0-alpha-1716-g69417e8
```

The serial build command generated for `preprocess` is:

```bash
ifort -DDART_VERSION="'v11.1.0-alpha-1716-g69417e8'" ...
```

and this builds successfully.

However, when the build moves to MPI executables, the corresponding command is:

```bash
mpiifort -DDART_VERSION="'v11.1.0-alpha-1716-g69417e8'" ...
```

and compilation of `version_mod.F90` fails.

The compiler appears to see the preprocessed line as:

```fortran
character(len=*), parameter :: dart_version_string = v11.1.0-alpha-1716-g69417e8
```

rather than a quoted Fortran character string.

#### Steps to reproduce

1. Configure DART with Intel Fortran and Intel MPI.
2. Run `quickbuild.sh` from the EAM-SE work directory.
3. `preprocess` builds and runs successfully with `ifort`.
4. The build then starts MPI programs using `mpiifort`.
5. Compilation fails on `build_templates/version_mod.F90`.

#### Expected outcome

`DART_VERSION` should expand to a valid Fortran character literal, for example:

```fortran
character(len=*), parameter :: dart_version_string = 'v11.1.0-alpha-1716-g69417e8'
```

and both serial and MPI executables should compile successfully.

#### Actual outcome

During the `mpiifort` build, the quotation marks around the version string appear to be lost, causing the version text to be interpreted as Fortran code.

### Error Message

The first error is:

```text
build_templates/version_mod.F90(20): error #5082: Syntax error, found REAL_CONSTANT '.1'

character(len=*), parameter :: dart_version_string = v11.1.0-alpha-1716-g69417e8
--------------------------------------------------------^
```

This is followed by errors such as:

```text
error #6592: This symbol must be a defined parameter ... [V11]
error #6592: This symbol must be a defined parameter ... [ALPHA]
error #6592: This symbol must be a defined parameter ... [G69417E8]
error #6404: This name does not have a type ... [V11]
```

The build then terminates with:

```text
compilation aborted for build_templates/version_mod.F90 (code 1)
make: *** [version_mod.o] Error 1
```

E3SM Atmosphere Model / EAM-SE (`models/eam-se`).

### System / compiler information

The build is being performed on PNNL Compy using Intel Fortran and Intel MPI.

Serial compiler:

```text
ifort
```

MPI compiler wrapper:

```text
mpiifort
```

Relevant library paths in the generated build include:

```text
NetCDF 4.6.3
PnetCDF 1.9.0
Intel MPI 2020
```

### Version of DART

```text
v11.1.0-alpha-1716-g69417e8
```

### Additional information

The interesting part is that the same apparent macro definition works during the serial `ifort` compilation of `preprocess`, but fails when used through `mpiifort`.

The current DART shell code generating the definition is:

```bash
# preprocessor definition for DART version
version_def="-DDART_VERSION="'$git_version'""
```

This may require more robust escaping/quoting when the compiler is invoked through an MPI compiler wrapper.

I am happy to provide the output of `mpiifort -show`, the generated Makefile, or test alternative quoting approaches if that would be useful.

hkershaw-brown:
## Description:

non-cppflags version of adding the git version string to dart
Buildfunction dartversion now writes the module with the dart version string. The cppdefs version string was not surviing certain mpiifort compiler wrappers. Rather than keep adding escapes \\\\, to get the string preprocessed, this pull request creates the version_mod.f90 file during the build. 

temp file created with temp.$$ to avoid race condition creating the file (when multiple quickbuilds are run simultaneously)
Sanitizing the string for funky characters and trimming just incase people get creating with tags. Note I am in two minds about whether this should be bash substitution since we already use bash substitution elsewhere in the buildfunctions. 
```
git_version="${git_version//[^A-Za-z0-9._+-]/_}" 
git_version="${git_version:0:64}" 
```

Currently checking whether the version has changed before overwriting version_mod.f90. Quickbuild.sh removes all .o and .mod so stuff gets rebuild anyway, but possibly at somepoint we may be more sensible about compiling. 
removes the -c from the mkmf calls in the build functions. 

### Fixes issue

fixes #1164

### Types of changes

- [ ] Documentation update

### Tests
I am testing the builds with CIRRUS on this pull request. 
Still to test:
- [x] outofcore builds
- [x] funky tags
- [x] run all quickbuild.sh on Derecho

## Checklist for merging

- [ ] Updated changelog entry
- [ ] Documentation updated
- [ ] Update conf.py

## Checklist for release
- [ ] Merge into main
- [ ] Delete feature-branch

## Testing Datasets

- [ ] No dataset needed


hkershaw-brown:
Note we cannot be more sensible about compiling without the other race condition (obs_def_mod, obs_kind_mod) when preprocess out-of-the-box in dart is writing the same file for all dart builds.

mjs2369 (approved):
Hi @hkershaw-brown I failed your challenge and could not break this code, so kudos

The value of git_version is now read as plain text with the printf and string format directive; this means it can't be read as a format string or interpreted as pure fortran code like in the issue share by our user. 

And I think we are protected from edge cases in the git_version as well, since you replace all characters that aren't a letter, digit, or `.`, `_`, `+`,`-` with underscores; this command is also performed in posix so the character classification is in plain ASCII.

I also went ahead and did run_all quickbuilds on Derecho and the results were identical with those on main (10 FAILED for each compiler, except ifx which also fails for GSI2DART and therefore is 11)

I think this is ready to go!

P.S. I learned some cool things reviewing this PR

